PR #18, opened 2016-05-30
Pluggable message class and pep8 fixes

Gr1N:
Hi!

I want to introduce some simple but useful functionality. In some cases we need to redefine message class, e.g. add some additional attributes. In current library this is impossible (yes I know about python.mock, but it is dirty).

Also I fixed some pep8 warning and unused imports.

What do you think?

### DIFF
--- a/nats/__init__.py
+++ b/nats/__init__.py
@@ -3,6 +3,7 @@
 import asyncio
 from .aio.client import Client as NATS
 
+
 @asyncio.coroutine
 def connect(**options):
     nc = NATS()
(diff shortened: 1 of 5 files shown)


wallyqs:
I think sounds good, thanks for the pep8 fixes too! 👍

Gr1N:
Nice!

@wallyqs, could you please create new package (something like v0.3.1) with this functionality and publish on PyPI?

wallyqs:
OK, pushed earlier a minor release with your changes with tag `v0.3.1`, thanks again for the contribution!

Gr1N:
Many thanks!

PR #19, opened 2016-06-03
Lazy initialization of loop in client

habibutsu:
In current implementation the following code it causes an exception - **RuntimeError**

```python
>>> import asyncio
>>> loop = asyncio.new_event_loop()
>>> asyncio.set_event_loop(None)
>>> import nats
...
RuntimeError: There is no current event loop in thread 'MainThread'.
```


wallyqs:
Thanks, merging

PR #34, opened 2017-05-04
Add TLS support

brianshannan-wf:
Adds TLS support
- Some general code cleanup like removing unused imports and code deduplication
- Added certs/keys from the nats tornado library for testing
- Updated the travis distro to what the nats tornado library was it appeared the default had an old ssl version


nats/aio/client.py
```diff
@@ -768,6 +772,28 @@ def _process_connect_init(self):
...
+            if not sock:
+                # This shouldn't happen
+                raise NatsError('unable to get socket')
+            yield from self._io_writer.drain()  # just in case something is left
+
+            self._io_reader, self._io_writer = \
+                yield from asyncio.open_connection(
+                    loop=self._loop,
+                    limit=DEFAULT_BUFFER_SIZE,
+                    sock=sock,
```
wallyqs: Nice... I missed that calling `open_connection`  this way does not create another connection and works ok to wrap it using the TLS context.

script/install_gnatsd.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-export DEFAULT_NATS_SERVER_VERSION=v0.8.0
+export DEFAULT_NATS_SERVER_VERSION=v0.9.6
```
wallyqs: 👍

tests/client_test.py
```diff
@@ -762,6 +763,52 @@ def worker_handler(msg):
     self.assertEqual(1, err_count)
 
 
+class TLSTest(TLSServerTestCase):
```
wallyqs: would be nice to add some reconnection tests with TLS context as well (can add in different PR)

readme.md
```diff
@@ -339,6 +339,19 @@ if __name__ == '__main__':
...
+TLS connections can be configured with an [ssl context](https://docs.python.org/3/library/ssl.html#context-creation)
+
+```python
+ssl_ctx = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
+ssl_ctx.protocol = ssl.PROTOCOL_TLSv1_2
+ssl_ctx.load_verify_locations('ca.pem')
+ssl_ctx.load_cert_chain(certfile='client-cert.pem',
+                        keyfile='client-key.pem')
+yield from nc.connect(io_loop=loop, tls=ssl_ctx)
+```
```
wallyqs: like the API here too 👍

PR #39, opened 2017-07-11
Add support for auth token

### LINKED ISSUE #37: Auth Code Authentication
Will auth code authentication be included? Authenticating with token in connection string (i.e. "nats://TOKEN@localhost:4222") passes in the token as username server side.  Changing to "nats://:TOKEN@localhost:4222" passes the token in as password, but still does not work.

wallyqs:
Enable passing an auth token to the server on `CONNECT` by setting it as part of the URI from a server.

```python
options = {
  'servers': [
      'nats://token@127.0.0.1:4223'
   ]
}
yield from nc.connect(**options)
```

Fixes #37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,6 @@ python:
   - 3.5.1
   - 3.6
 
-env:
-  - DEBUG_NATS_TEST=true
-
 before_install:
   - bash ./script/install_gnatsd.sh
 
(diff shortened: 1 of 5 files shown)


delfick:
Hi, is there a timeline on when a new version of this library will be released with this fix?

wallyqs:
Hi @delfick sorry for the delay. I will do another major release next week but in the meantime I have just pushed earlier this minor version which includes this fix: https://github.com/nats-io/asyncio-nats/releases/tag/v0.5.2

delfick:
excellent! much appreciated!

PR #47, opened 2017-11-16
fixup reconnect: raise `NatsError` on empty server info response

Gr1N:
Hi!

Here is a small fix for reconnect logic. @wallyqs what do you think about fix? Or maybe you know a better solution?

Best regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ var
 sdist
 develop-eggs
 .installed.cfg
+
+.vscode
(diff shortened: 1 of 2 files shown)


wallyqs:
Thanks for the patch, this is good for now. Tests coverage would be nice but can take a look at adding before making a release with the fix.

Gr1N:
Could you please upload new version of package?

wallyqs:
Thanks again for the contribution, have uploaded a new version of the package now which includes this fix: https://github.com/nats-io/asyncio-nats/releases/tag/v0.6.2

Gr1N:
Thanks!

PR #49, opened 2017-12-04
Prevent TLS from closing randomly

oblalex:
Hi,

This PR keeps references to bare transport objects to prevent GC from deleting them prematurely, as deletion of transport objects closes underlying sockets.

This is a critical issue, as such behaviour breaks write operations and totally ruins loop which listens for incoming read events from socket.

See also nats-io/asyncio-nats#43.

Thanks

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -10,7 +10,7 @@ from nats.aio.errors import *
 from nats.aio.utils import new_inbox
 from nats.protocol.parser import *
 
-__version__ = '0.6.2'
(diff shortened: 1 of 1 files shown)


wallyqs:
Just pushed a release including your changes: https://github.com/nats-io/asyncio-nats/releases/tag/v0.6.4
Thanks a lot for the root cause investigation of this tricky issue!

oblalex:
Thank you very much for rapid response!

PR #50, opened 2017-12-05
Cancel reconnection task on close

oblalex:
Hi,

If you try to close client when it is reconnecting, loop will throw a `Task was destroyed but it is pending!` error which cannot be caught.

This PR cancels reconnection task when client is closed, so that event loop can be closed gracefully.

Thanks

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -111,6 +111,7 @@ class Client(object):
         self._disconnected_cb = None
         self._closed_cb = None
         self._reconnected_cb = None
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks!

oblalex:
I can see tests have failed after merge, however they actually were ok before merge. Is this some known issue?

oblalex:
Hm, from build history I see this happens from time to time and the reason is somewhere else.

PR #98, opened 2019-02-24
Added async callbacks wrapped in functools.partial

aratz-lasa:
Possibility to add async functions wrapped in functools.partial, when creating a subscription.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -506,6 +506,8 @@ class Client(object):
         if cb is not None:
             if asyncio.iscoroutinefunction(cb):
                 sub.coro = cb
+            elif hasattr(cb, "func") and asyncio.iscoroutinefunction(cb.func):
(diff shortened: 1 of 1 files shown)


aratz-lasa:
Will this pull request be merged? Just curious 🤔

wallyqs:
Hi, thanks for the PR and sorry for the late feedback. I think change LGTM but we need to add a small test for it.  I've added one here https://github.com/nats-io/asyncio-nats/compare/functools-subs let me know if it looks good to you then I'll merge.  Also maybe should include some examples here: https://github.com/nats-io/asyncio-nats/tree/master/examples

aratz-lasa:
I see it very correct! Do you want me to write an example?

wallyqs:
OK! I will merge this then, and yes more examples would be great :) thanks!

Issue #113, opened 2019-08-23
Pending_size of subscription `_INBOX` is always growing


laishzh:
I am using nats in a long-time task. After a long period, an error of "nats: slow consumer, messages dropped" is raised. I find out that there is a problem when dequeue a message from Subscription `_INBOX`. Following is a demo to show this case. The pending_size of `_INBOX` is always growing. 

* NATS Version: 2.0.0-linux (docker)
* nats.py version: 0.9.2

Demo:
```
import asyncio
from nats.aio.client import Client as NATS
from nats.aio.errors import ErrConnectionClosed, ErrTimeout, ErrNoServers

async def run(loop):
    nc = NATS()

    await nc.connect("127.0.0.1:4222", loop=loop)

    async def help_request(msg):
        subject = msg.subject
        reply = msg.reply
        data = msg.data.decode()
        print("Received a message on '{subject} {reply}': {data}".format(
            subject=subject, reply=reply, data=data))
        await nc.publish(reply, b'I can help')

    # Use queue named 'workers' for distributing requests
    # among subscribers.
    sid = await nc.subscribe("help", "workers", help_request)

    while True:
        # Send a request and expect a single response
        # and trigger timeout if not faster than 1 second.
        try:
            response = await nc.request("help", b'help me', timeout=1)
            print("Received response: {message}".format(
                message=response.data.decode()))
        except ErrTimeout:
            print("Request timed out")

        # Output all subscriptions' pending_size
        print("NC subs: {}".format(len(nc._subs)))
        for k, v in nc._subs.items():
            print("Key: {}, SubName: {}, pending_size: {}".format(k, v.subject, v.pending_size))

    # Remove interest in subscription.
    await nc.unsubscribe(sid)

    # Terminate connection to NATS.
    await nc.close()

if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run(loop))
    loop.close()

```

Output: 
```
NC subs: 2
Key: 1, SubName: help, pending_size: 0
Key: 2, SubName: _INBOX.sXm35gMyTu5v7u0Zmpi9h2.*, pending_size: 28450
Received a message on 'help _INBOX.sXm35gMyTu5v7u0Zmpi9h2.sXm35gMyTu5vlpVcmpi9h2': help me
Received response: I can help
NC subs: 2
Key: 1, SubName: help, pending_size: 0
[...]

PR #114, opened 2019-08-23
Decrease pending_size when dequeue a message in request.wait_for_msgs

laishzh:
Decrease pending_size when dequeue a message in request.wait_for_msgs. #113

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -876,6 +876,7 @@ class Client(object):
                 while True:
                     try:
                         msg = yield from sub.pending_queue.get()
+                        sub.pending_size -= len(msg.data)
                         token = msg.subject[INBOX_PREFIX_LEN:]
 
                         try:


vshender:
Why don't you make a release for this fix? It's a critical problem, and these changes are not in any release version of the library yet.

wallyqs:
@vshender apologies... I just cut the release including this fix, thanks.
https://pypi.org/project/asyncio-nats-client/0.10.0/

vshender:
@wallyqs thank you very much!

abhaira:
I see the same bug for pull subscribers. Are they handled as part of this request?

wallyqs:
His @abhaira maybe try with latest version, it should be fixed there for pull subscribers too.

abhaira:
@wallyqs Thanks for the response. I have checked the code on the latest version. The code is there to subtract the queue byte size upon consumption of a message. I will test it and confirm. Thanks again.

PR #120, opened 2019-11-29
Provide an option to open tls connection right from the beginning

krrishnarraj:
This option can be used when nats server is behind a
load-balancer with tls endpoint

References:
https://github.com/nats-io/nats-server/issues/291#issuecomment-499315013
https://gist.github.com/saadtazi/6ff02f3b38a2bdebbcc28acbe7885145


wallyqs:
Added as part of #511

PR #121, opened 2019-12-02
Configurable concurrency level for messages processing

### LINKED ISSUE #119: Configurable concurrency level for messages processing
I have encountered some limitation of the client, I  would like to set how many concurrent tasks can be created simultaneously for the subscription.
Currently it can be either one or unlimited (`is_async=True`).

I found a comment in in the client:
```
# NOTE: Deprecate this usage in a next release,
# the handler implementation ought to decide
# the concurrency level at which the messages
# should be processed.
```
We can take it a bit further and add ability to configure an exact number of running tasks per subscription. 
It should relatively easy to implement it using `asyncio.Semaphore`.

What do you think about it? I would be happy to create a PR.

maruqu:
This PR adds `max_cb_concurrency` parameter to `subscribe` method.
Concurrent cb execution limit is achieved using `asyncio.Semaphore`.

TO DISCUSS:
- [x] What should be a default max concurrency parameter
- [x] Should we deprecate `is_async` parameter (seems redundant and a bit misleading now)

Resolves #119


nats/aio/client.py
```diff
@@ -707,6 +712,14 @@ def _drain_sub(self, sid):
                 nonlocal sub
                 nonlocal err_cb
 
+                semaphore = asyncio.Semaphore(sub.max_cb_concurrency)
```
wallyqs: I think creating the semaphore object should be optional for when `sub.max_cb_concurrency` is greater than 1, wdyt?

wallyqs:
Thanks for the contribution! About the discussion points you mention:

> What should be a default max concurrency parameter 

I think default max concurrency should be still one following the example of the Go client

> Should we deprecate is_async parameter (seems redundant and a bit misleading now)

Yes we should deprecate this, probably do the same with `subscribe_async`, document that this is controlled by the concurrency level field now.  We can do this in a separate PR.

nats/aio/client.py
maruqu: maybe the semaphore might be a `sub` property? I personally don't like accessing variables declared in conditional blocks from other places. 
`if sub.is_async:` would have to become `if sub.is_async and sub.max_cb_concurrency > 1:` 

TBH I would take this chance to refactor this part of code a bit more. Maybe we can get rid of `is_async` and demand that `sub.cb` is a coroutine (remove `coro` parameter). I am aware that it will need few more refactors (and probably a braking change), but it might be a right time for it. 
This way current  code:

```
                            # Invoke depending of type of handler.
                            if sub.coro is not None:
                                if sub.is_async:
                                    # NOTE: Deprecate this usage in a next release,
                                    # the handler implementation ought to decide
                                    # the concurrency level at which the messages
                                    # should be processed.
                                    self._loop.create_task(sub.coro(msg))
                                    await semaphore.acquire()
                                    self._loop.create_task(
                                        release_semaphore_after_cb(msg)
                                    )
                                else:
                                    await sub.coro(msg)
                            elif sub.cb is not None:
                                if sub.is_async:
                                    raise NatsError(
                                        "nats: must use coroutine for async subscriptions"
                                    )
                                else:
                                    # Schedule regular callbacks to be processed sequentially.
                                    self._loop.call_soon(sub.cb, msg)
```
Will become simpler:
```
                            if sub.max_cb_concurrency > 1:
                                await semaphore.acquire()
                                self._loop.create_task(
                                    release_semaphore_after_cb(msg)
                                )
                            else:
                                await sub.cb(msg)
```

wallyqs: I like those ideas a lot 👍  I'm ok as well with a breaking change if it makes things simpler as well.

maruqu: Cool, I'll see what I can do 🙂

maruqu:
@wallyqs can you fix `ClientTLSTest` tests and document the changes in next PR?

wallyqs:
Sorry the the TLS certs needed and update for the tests to pass (fixed via https://github.com/nats-io/nats.py/pull/123), if you rebase you should get the changes now.

maruqu:
@wallyqs What do you think? Did I go too far with the refactor?

tests/client_test.py
```diff
@@ -370,28 +375,46 @@ def test_connect_syntax_sugar(self):
...
+        for i in range(0, 10):
+            await nc.publish("tests".format(i), b'bar')
 
         # Wait a bit for messages to be received.
         await asyncio.sleep(1, loop=self.loop)
-        self.assertEqual(5, len(msgs))
-        self.assertEqual("tests.1", msgs[4].subject)
-        self.assertEqual("tests.3", msgs[3].subject)
+        self.assertEqual(10, len(msgs))
+        assert running_handlers.max == 3
```
wallyqs: do you recommend to use more `assert` instead of `assertEqual`?

tests/client_test.py
```diff
@@ -271,6 +273,9 @@ def test_connect_syntax_sugar(self):
         self.assertEqual(1, connz['connections'][0]['out_msgs'])
         self.assertEqual(11, connz['connections'][0]['out_bytes'])
 
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8), reason="https://bugs.python.org/issue33261"
+    )
```
wallyqs: Thanks for linking to the issue 👍

wallyqs:
@maruqu I think it looks great! Also good to have a few things cleaned up finally

tests/client_test.py
maruqu: Just muscle memory, fixed to match other tests.

maruqu:
@wallyqs  I am glad I could help. Can you release a new package version containing this change?  Maybe an alpha?

hvt:
Is this the `MaxInFlight` behavior of the Go client?

Positron1990:
Will anybody resolve this PR or it will be other concurrency level tools in the library?

PR #128, opened 2020-01-15
Raise error when client callback functions are not coroutines

### LINKED ISSUE #127: Client silently crashes if callbacks are not co-routines
Documentation does not make it clear that callback functions must be async coroutines. The fact that message handlers can be simple functions does not help with that either.

If, by mistake, someone writes...

```
def error_cb():
    print("an error occured")

client = Client(...., error_cb=error_cb)
```

... he will not get any error, and the client will silently crash the first time the error callback is awaited, without any indications.

neoaggelos:
Closes #127.

Raise a `nats.aio.errors.ErrInvalidCallbackType` error if any of the client callbacks are not coroutine functions (check with `asyncio.iscoroutinefunction`)

Also added relevant test cases.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -242,6 +242,11 @@ class Client(object):
             user_credentials=None,
             nkeys_seed=None,
     ):
+        for cb in [error_cb, disconnected_cb, closed_cb, reconnected_cb,
+                   discovered_server_cb]:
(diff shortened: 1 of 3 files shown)


neoaggelos:
Hmm, some existing tests are mocking the `disconnected_server_cb` and currently failing with these changes. Will look into it

PR #139, opened 2020-02-04
formating changes

matthiashanel:
ran yapf


matthiashanel:
The problem is a difference between make format and make ci.
The change has to happen in the make file first.

PR #146, opened 2020-03-23
Ability to access the connection identifier

arun11299:
There are cases where I would have liked to know about the connection identifier in the application logs. This helps in figuring out what happened to those connections from the nats server log.

This pull, exposes the "client_id" which I believe is also the connection identifier in nats server.


imsazzad:
how to get the cid from python client.
my client version is 0.10.0
I can see no client_id/cid here.
I need it for monitoring.

PR #147, opened 2020-03-30
Drain non-callback subscriptions correctly

chiaolun:
https://github.com/nats-io/nats.py/blob/2d5ba442d09397c5318ecb90512435e8e6a6d70f/nats/aio/client.py#L654
Subscriptions that use the `future` instead of the `cb` argument do not have `pending_queue` and throw an exception when drained

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -577,10 +577,11 @@ class Client(object):
                 # Roundtrip to ensure that the server has sent all messages.
                 await nc.flush()
 
-                # Wait until no more messages are left,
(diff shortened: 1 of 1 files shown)


wallyqs:
LGTM, thanks for the PR! One thing though is that if you are using `future` is most likely because you might be using the type of request/response where each one of the messages is creating a subscription and the response was represented with a future. If that is the case it might be better to use the `request` API instead.

chiaolun:
Indeed that's the case. Thank you for the comment, didn't realize what the other `request` method was for before this.

PR #150, opened 2020-04-05
Use loop.start_tls to upgrade to TLS

chiaolun:
Was running into problems using uvloop with TLS on, because the current TLS upgrade reuses the socket and throws
```Traceback (most recent call last):
  File "/home/chiao/anaconda3/lib/python3.7/site-packages/nats/aio/client.py", line 293, in connect
    await self._process_connect_init()
  File "/home/chiao/anaconda3/lib/python3.7/site-packages/nats/aio/client.py", line 1537, in _process_connect_init
    server_hostname=hostname,
  File "/home/chiao/anaconda3/lib/python3.7/asyncio/streams.py", line 77, in open_connection
    lambda: protocol, host, port, **kwds)
  File "uvloop/loop.pyx", line 2007, in create_connection
  File "uvloop/loop.pyx", line 1995, in uvloop.loop.Loop.create_connection
  File "uvloop/handles/tcp.pyx", line 175, in uvloop.loop.TCPTransport._open
  File "uvloop/handles/tcp.pyx", line 34, in uvloop.loop.__tcp_open
FileExistsError: [Errno 17] File exists
```

Was able to get things to work after using asyncio's provided function for TLS upgrades. Unfortunately mutating the private transport of self._io_writer directly and also reading the private protocol directly. Looking for ideas on how to implement this better but the hack in this PR is working for us in the meantime.


chiaolun:
Unfortunately only available in Python 3.7

rit:
I ran into this error as well. I'm punting on `uvloop` for now. :-(

PR #156, opened 2020-05-18
clean up subscriptions in the event of a timeout

chiaolun:
Also, aren't the `future.cancel()` redundant given that `wait_for` cancels in the event of a timeout?


wallyqs:
you are right, cancelling the future is redundant: https://github.com/python/cpython/blob/4482337decdbd0c6e2150346a68b3616bda664aa/Lib/asyncio/tasks.py#L472

Issue #134, opened 2020-02-01
Revise TLS usage for asyncio 3.9 (deprecation warning)


wallyqs:
```
DeprecationWarning: Using send() method on sockets returned from get_extra_info('socket') will be prohibited in asyncio 3.9
```

chiaolun:
Using asyncio.start_tls instead of reusing the socket should fix this?
https://github.com/nats-io/nats.py/pull/150

PR #164, opened 2020-09-08
Revise tls upgrade process in newer versions of python

brianshannan:
Resolves https://github.com/nats-io/nats.py/issues/134

The current process used for tls is deprecated and will be removed in python 3.9. In python 3.7, [start_tls](https://docs.python.org/3/library/asyncio-eventloop.html#tls-upgrade) was added, which is a cleaner way to do the tls upgrade process than what is currently implemented. This requires manually creating the stream reader/writer from a transport and protocol, but it isn't that hard to do. I left the previous method for python 3.5 and 3.6, as `start_tls` wasn't present for those versions.

Additionally, the `allow_failures` section of the ci config doesn't actually run those configurations, it just allows failures, so they must be run in the main section. Because of this, python 3.6 and 3.8 weren't being run in ci. I made 3.8 non-optional but left 3.6 as optional, as the tests use an api introduced in 3.7, and an exception class for tls hostname mismatches appears to be different in 3.6.

@wallyqs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ cache:
   - $HOME/nats-server
 
 python:
+  - "3.6"
   - "3.7"
+  - "3.8"
 
 before_install:
   - bash ./script/install_nats.sh
@@ -29,4 +31,3 @@ jobs:
   allow_failures:
     python:
       - "3.6"
-      - "3.8"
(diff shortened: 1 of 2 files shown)


wallyqs:
LGTM! 💯 Also confirmed this fixes the issues with `uvloop` and tls.

```
python3.7 examples/nats-pub -s tls://demo.nats.io:4443 hello -d world
python3.7 examples/nats-sub -s tls://demo.nats.io:4443 hello
```

(For some reason when using `uvloop` now I get this error `returning true from eof_received() has no effect when using ssl`). Also re: `allow_failures` 3.6 maybe we can add a skip for those tests for now.  Thanks!

Issue #100, opened 2019-04-23
NATS consumer breaks the last callback on drain()


alec-alpha:
NATS consumer breaks the last message callback while drain()-ing with non-empty message buffer.

Testing code:
```
#!/usr/bin/env python3

import asyncio
from nats.aio.client import Client as NATS

async def main(loop, uri='nats://localhost:4222', subj='tests.drain_bug', nmsgs=10):
        sent, got, ok = 0, 0, 0

        async def on_msg(msg):
                nonlocal got, ok
                got += 1
                await asyncio.sleep(0.1, loop=loop)
                ok += 1

        sub = NATS()
        await sub.connect(uri, loop=loop)
        await sub.subscribe(subj, cb=on_msg)

        pub = NATS()
        await pub.connect(uri, loop=loop)
        for i in range(nmsgs):
                body = 'message #%d' % i
                await pub.publish(subj, body.encode())
                sent += 1
        await pub.flush() 
        await pub.close() 

        await sub.drain()
        print('sent=%d got=%d ok=%d' % (sent, got, ok))

if __name__ == '__main__':
        loop = asyncio.get_event_loop()
        loop.run_until_complete(main(loop))
```

Testing output: sent=10 got=10 ok=9
Expected output: sent=10 got=10 ok=10

Tested on Debian/buster with Python v3.7.3rc1 and asyncio-nats-client==0.8.2 (pip version).

PR #165, opened 2020-09-08
wait for all messages to be processed while draining

brianshannan:
Resolves https://github.com/nats-io/nats.py/issues/100

Before the client would block until the last message was retrieved from the queue, and then cancel the task the message callback would run in. With this change, the task isn't cancelled until all the messages from the queue are successfully processed. Additionally, `drain` now won't return until all messages have been successfully processed, which I think is the proper thing to do.

@wallyqs


wallyqs (approved):
LGTM, agree this approach is much better. Thanks.

PR #167, opened 2020-09-09
remove explicit loops

brianshannan:
See commit message for details.

This is a breaking change, so there's some extra considerations for that, though a 1.0 release hasn't been cut yet. IMO the loop parameters should be removed before a 1.0 release. Python 3.7 introduced a convenience method `asyncio.create_task()` which would cut down on some of the loop overhead if support for 3.6 is ever dropped.

@wallyqs

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -145,7 +145,6 @@ class Client:
         return f"<nats client v{__version__}>"
 
     def __init__(self):
-        self._loop = None
         self._current_server = None
         self._server_info = {}
         self._server_pool = []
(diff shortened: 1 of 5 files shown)


wallyqs:
Thanks @brianshannan , ok so it's been awhile since there has been a release so think I will push v0.9.4 first to pypi with latest change, then we start including the breaking changes that would be part of v1.0.0.

charliestrawn:
It might not be a bad idea to have a pre-1.0 integration/preview branch to target said breaking changes. This could act as a sort of release candidate and let consumers weigh in on coming changes before the release gets cut?

wallyqs:
Just pushed the release v0.11.0, I'm +1 on merging this already and starting to work towards v1.0.0 on the main branch with the breaking changes but having something like a `v1-dev` branch that we could also merge later sounds good to me as well.

wallyqs (approved):
LGTM

brianshannan:
Closing as this has been added to v1-dev

PR #169, opened 2020-09-11
Add Python 3.8 to setuptools classifiers

gtaylor:
@wallyqs Adds on https://github.com/nats-io/nats.py/pull/168 by setting the 3.8 setuptools classifier.

I didn't add 3.9 yet since it's in dev.


wallyqs:
LGTM, thanks. I already uploaded to pypi so may do a patch release so that it displays properly in the metadata.

Issue #91, opened 2018-10-13
Why is the default behavior of the subscribe method to ignore errors when there is no error_cb defined?


jccardonar:
(This was probably discussed before, but I couldn't find where.)

The examples of the library do not force one to use the error cb from day-1. However, without it defined,  errors are silently discarded and there are no signs of them ever happening.

I would prefer for the error to actually be raised. Right, the task will stop, but that is a louder noise than just processing messages as if nothing had happened. It would help if error_cb is always defined in the examples.

We really need #76 for this not to matter, I think.

cgebe:
+1 just wasted an hour since the message_handler was silence on every error.

bvanelli:
Even though I agree (since I've run into similar problems with errors being supressed), it has to be noted that the `error_cb` catches everything, not only problems in the message handling, like connection lost with the server and reconnect atempts, things that the user might want to supress.

I believe errors in the callbacks are supressed to prevent the application from crashing if one subscription has an unexpected problem. Maybe stress on the documentation or replace in the examples:

```python
await nc.connect("demo.nats.io:4222", loop=loop)
```

by something like:

```python
await nc.connect("demo.nats.io:4222", loop=loop, error_cb=print)
```

So that errors actually get printed to the user.

Issue #92, opened 2018-11-01
Silently Hiding Exceptions in Callbacks


tolmanam:
An exception that occur in the call back function are not bubbling up.

Here's an example based on the **Basic Usage** code form the README.  The print statements within the callbacks have been sabotaged.

``` python
import asyncio
from nats.aio.client import Client as NATS
from nats.aio.errors import ErrConnectionClosed, ErrTimeout, ErrNoServers

async def run(loop):
    nc = NATS()

    await nc.connect("demo.nats.io:4222", loop=loop)

    async def message_handler(msg):
        subject = msg.subject
        reply = msg.reply
        data = msg.data.decode()
        # 'oops' is not provided so this print will raise an exception
        print("Received a message on '{subject} {reply}': {data} {oops}".format(
            subject=subject, reply=reply, data=data))

    # Simple publisher and async subscriber via coroutine.
    sid = await nc.subscribe("foo", cb=message_handler)

    # Stop receiving after 2 messages.
    await nc.auto_unsubscribe(sid, 2)
    await nc.publish("foo", b'Hello')
    await nc.publish("foo", b'World')
    await nc.publish("foo", b'!!!!!')

    async def help_request(msg):
        subject = msg.subject
        reply = msg.reply
        data = msg.data.decode()
        # 'oops' is not provided so this print will raise an exception
        print("Received a message on '{subject} {reply}': {data} {oops}".format(
            subject=subject, reply=reply, data=data))
        await nc.publish(reply, b'I can help')

    # Use queue named 'workers' for distributing requests
    # among subscribers.
    sid = await nc.subscribe("help", "workers", help_request)

    # Send a request and expect a single response
    # and trigger timeout if not faster than 200 ms.
    try:
        response = await nc.request("help", b'help me', 0.2)
        print("Received response: {message}".format(
            message=response.data.decode()))
    except ErrTimeout:
        print("Request timed out")

    # Remove interest in subscription.
    await nc.unsubscribe(sid)

    # Terminate connection to NATS.
    await nc.close()

if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run(loop))
    loop.close()
```

captain-sysadmin:
I too have bumped into this. 
I was trying to create a proof of concept for `nc.requests` , however I forgot to convert the reply to `bytes()` like @tolmanam this resulted in silent failures. 

my call back was:
```python
     async def help_request(msg):
        subject = msg.subject
[...]

PR #170, opened 2020-09-12
provide a default error callback

brianshannan:
See commit message for details, @wallyqs let me know if you had something else in mind. The error handling for synchronous callbacks (IMO support should definitely be removed for) and callbacks run as tasks (are currently deprecated, which makes sense to me to me as unbounded concurrency seems bad) isn't perfect, and I didn't do anything to them as they're deprecated.

Resolves https://github.com/nats-io/nats.py/issues/91, https://github.com/nats-io/nats.py/issues/92

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -21,6 +21,7 @@ import base64
 from random import shuffle
 from urllib.parse import urlparse
 import sys
+import logging
 
 from nats.aio.errors import *
 from nats.aio.utils import new_inbox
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks @brianshannan, I think a default internal logger is the way to go as well. This is still backwards compatible so I think ok to merge already, also agree that should remove [the catch all](https://github.com/nats-io/nats.py/blob/master/nats/aio/client.py#L756-L760) for callback errors.

PR #175, opened 2020-09-22
make api more consistent with other clients

brianshannan:
This is a lot of breaking changes, so let me know if you're opposed to any of them.

This commit is an attempt to make the api of this client more closely
match the api of nats clients in other languages. Below is a non
exhaustive list of the changes made.

- Return subscriptions from `subscribe()` instead of a subscription id
- Allow iterating over messages in subscription with `async for`. To use, don't specify a callback when subscribing
- Move `unsubscribe()` and `drain(sid=sid)` to the subscription class
- Remove `publish_request()`, `publish()` now has a `reply` kwarg
- Remove `auto_unsubscribe()`, `subscribe()` already had a kwarg for this
- Remove support for executing synchronous callbacks
  - Remove `subscribe_async()`
  - Remove `is_async` kwarg from `subscribe()`
- Remove support for executing callbacks as tasks
- Remove multiple response functionality from `request()`
- Remove `timed_request()`, name is very misleading as all requests are timed. This was really old style requests, moved to a kwarg of `request()` called `old_style`
- Made connection context manager async
- Various internal code reoganizations
- Various doc cleanups
- Cleaned up example code

@wallyqs @charliestrawn


wallyqs (approved):
LGTM! I think these are all very welcome changes for the v1 client.

PR #176, opened 2020-09-24
Improvements for nats.connect 

wallyqs:
Many of the examples were using `from nats.aio.client import Client as NATS` which was too verbose, so for the v1 branch trying to emphasize simpler usage like:

```python
import asyncio
import nats

async def run():
    nc = await nats.connect("demo.nats.io")

    sub = await nc.subscribe("bar")
    await nc.publish("bar", b'First')
    await nc.publish("bar", b'Second')

    async for msg in sub.messages:
        print(f"Received a message on '{msg.subject} {msg.reply}': {msg.data.decode()}")

if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run())
    loop.close()
```

This changeset also removes a few redundant perf scripts, updates the readme and changes the `nats-pub`, `nats-sub`, `nats-req` so that they are closer to how the Go tools work.


readme.md
```diff
@@ -285,7 +288,7 @@ if __name__ == '__main__':
 ## TLS
 
 TLS connections can be configured with an [ssl context](https://docs.python.org/3/library/ssl.html#context-creation)
-
+pp
```
brianshannan, line 291: Remove this?

readme.md
```diff
@@ -48,27 +52,35 @@ async def run():
...
     # Use queue named 'workers' for distributing requests
     # among subscribers.
     sub = await nc.subscribe("help", "workers", help_request)
 
     # Send a request and expect a single response
-    # and trigger timeout if not faster than 1 second.
+    # and trigger timeout if not faster than 500 ms.
     try:
-        response = await nc.request("help", b'help me', timeout=1)
+        response = await nc.request("help", b'help me', 0.5)
```
brianshannan, line 83: nit: this should still be passed as a kwarg, as if the kwarg ordering in the function definition was changed this would stop working. Not a big deal.

readme.md
```diff
@@ -170,30 +174,29 @@ async def run():
...
+    await asyncio.sleep(1)
 
     # Drain gracefully closes the connection, allowing all subscribers to
     # handle any pending messages inflight that the server may have sent.
     await nc.drain()
 
 if __name__ == '__main__':
     loop = asyncio.get_event_loop()
-    loop.run_until_complete(run())
+    loop.run_until_complete(run(loop))
```
brianshannan, line 185: Should we remove `loop`? I don't think it's needed and it might be good not to encourage passing explicit loops around

brianshannan (approved):

PR #177, opened 2020-09-24
Updates for async for handling

wallyqs:
Added tests for example usage of the `async for` style of consuming messages plus a bug fix where unsubscribe was not updating the limit of max_msgs from a subscription.

Signed-off-by: Waldemar Quevedo <wally@synadia.com>

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -182,6 +182,7 @@ class Subscription:
         if self._conn.is_draining:
             raise ErrConnectionDraining
 
+        self._max_msgs = limit
         if limit == 0 or self._received >= limit:
             self._stop_processing()
             self._conn._remove_sub(self._id)
@@ -235,7 +236,8 @@ class _SubscriptionMessageIterator:
         self._unsubscribed_future = asyncio.Future()
 
     def _cancel(self):
-        self._unsubscribed_future.set_result(True)
+        if not self._unsubscribed_future.done():
+            self._unsubscribed_future.set_result(True)
 
     def __aiter__(self):
         return self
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -422,6 +422,100 @@ async def iterator_func(sub):
...
+        await nc.connect()
+
+        # Make subscription that only expects a couple of messages.
+        sub = await nc.subscribe('tests.>')
+        await nc.flush()
+
+        # Simple iterator to get next msg.
+        async def next_msg():
+            async for msg in sub.messages:
+                return msg
```
wallyqs, line 490: Think it would be cool to have a `sub.next_msg` like in the Go client as well, but if the future here times out then the message is lost, not sure right now but maybe there is a better way...

nats/aio/client.py
```diff
@@ -233,20 +234,32 @@ class _SubscriptionMessageIterator:
...
+            self._unsubscribed_future.set_result(True)
 
     def __aiter__(self):
         return self
 
     async def __anext__(self):
+        if self._iterator_future.done():
+            raise StopAsyncIteration
+
+        if self._unsubscribed_future.done():
```
brianshannan, line 250: I'm not really sure why two futures are necessary, the wait should be interrupted if the future is completed so this won't block for forever. If we want to exit as soon as possible I think checking the one future should be enough. Not sure what I'm missing?

tests/test_client.py
brianshannan: The future could potentially be stored on the subscription along with a flag indicating whether the result has been retrieved or not (if not, return/await on the future result, if not, create a new future and return/await on that)

nats/aio/client.py
wallyqs: You are right, not needed now so took the extra future out. I was getting a task pending error before but they went away once I added the `elif self._unsubscribed_future.done()` check below:

```python
Task was destroyed but it is pending!
...
  File "/python/nats.py/nats/aio/client.py", line 255, in __anext__
    get_task = asyncio.create_task(self._queue.get())
```

brianshannan (approved):

PR #178, opened 2020-09-24
Add next_msg to subscription

wallyqs:
Adds `next_msg` to subscriptions with optional timeout parameter.

```python
        sub = await nc.subscribe('tests.>')
        await nc.flush()

        for i in range(0, 2):
            await nc.publish(f"tests.{i}", b'bar')

        # Retrieve single message
        msg = await sub.next_msg()
```

Signed-off-by: Waldemar Quevedo <wally@synadia.com>


nats/aio/client.py
```diff
@@ -120,6 +120,26 @@ def messages(self):
...
+        """
+        future = asyncio.Future()
+
+        async def _next_msg():
+            msg = await self._pending_queue.get()
+            future.set_result(msg)
+            
+        task = asyncio.create_task(_next_msg())
+        try:
+            msg = await asyncio.wait_for(future, timeout)
```
wallyqs, line 136: For the implementation was looking at maybe using an async generator but for since there is a timeout involved went with getting directly from the pending queue instead for now.

PR #181, opened 2020-10-06
add type hints to publicly available apis

brianshannan:
@wallyqs @charliestrawn

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -22,6 +22,7 @@ from random import shuffle
 from urllib.parse import urlparse
 import sys
 import logging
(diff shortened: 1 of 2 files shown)


nats/aio/client.py
```diff
@@ -258,7 +259,13 @@ async def __anext__(self):
 class Msg:
```
charliestrawn, line 259: pre-existing: this would make good use case for a [dataclass](https://docs.python.org/3/library/dataclasses.html) potentially (available since python 3.6)

charliestrawn (approved):
LGTM 👀

wallyqs (approved):
LGTM

charliestrawn (approved):

PR #183, opened 2020-10-08
use fstrings where possible

brianshannan:
Now that this is 3.6+, we can use fstrings. This change uses fstrings where possible. Also, I extracted some of the protocol logic to the protocol package, as it was duplicated in client.py. Not sure what your feelings on this are.

@wallyqs @charliestrawn


wallyqs (approved):
Seems that fstrings are also the fastest way to concatenate this. LGTM!

PR #188, opened 2020-11-10
Tests - allow tests to run from any location

charliestrawn:
I noticed while try to run an individual test within PyCharm that it would fail to find the path of the config files. I believe other editors like vim and emacs have plugins that let you easily run a single test at a time from within the file. This change allows for that. You can test this out by running `cd tests && pytest test_client.py`.

I removed the requires from Pipfile because that only applies to local development, and I don't think we want to lock devs into 3.7.

I also removed NatsTestCasethe because it seemed to be only used to print the name of the test which is easily achievable with `pytest` and the `-v` flag i.e. `pytest tests/test_client.py -v`. The `if __name__ == '__main__':` also become somewhat redundant for the same reason.


wallyqs (approved):
LGTM! 👍

PR #196, opened 2021-02-25
Support set pending size to client connect options

blackstorm:
For small payload and small amount message publish. 
Support set `pending_size`not only `DEFAULT_PENDING_SIZE` can be fast publish to nats server


wallyqs (approved):
LGTM

PR #214, opened 2021-10-07
Update __main__.py

pturanyi:
Authentication by token against NATS. It is needed by Argo event bus.

### DIFF
--- a/examples/nats-pub/__main__.py
+++ b/examples/nats-pub/__main__.py
@@ -43,6 +43,7 @@ async def run(loop):
     parser.add_argument('-d', '--data', default="hello world")
     parser.add_argument('-s', '--servers', default="")
     parser.add_argument('--creds', default="")
+    parser.add_argument('--token', default="")
     args, unknown = parser.parse_known_args()
 
     data = args.data
@@ -62,6 +63,9 @@ async def run(loop):
 
     if len(args.creds) > 0:
         options["user_credentials"] = args.creds
+
+    if args.token.strip() != "":
+        options["token"] = args.token.strip()
 
     try:
         if len(args.servers) > 0:


charbonnierg:
I don't really feel this is the place to have this kind of discussion but as I started to answer I might as well finish.

There's no need to install Go or build anything in order to use `nats` CLI. It is available for several architectures including `armv7`, `arm64` and `amd64`. In most cases, installing `nats` CLI is as simple as downloading a file into the correct path.

- When building Docker images, it's even simpler. You just need to add `nats` executable from [`synadia/nats-box`](https://hub.docker.com/r/synadia/nats-box) Docker image:

> Dockerfile doc about `COPY --from`: https://docs.docker.com/develop/develop-images/multistage-build/#use-an-external-image-as-a-stage. Note that you cannot use variables in the image name provided as `--from` value, so version must be hardcoded in Dockerfile.
```
COPY --from=synadia/nats-box:latest  /usr/local/bin/nats /usr/local/bin/nats
```

- Or assuming you have `bash`, `wget` and `unzip` installed, you can use a script that can be used accross different platforms if needed:

```bash
#!/usr/bin/env bash

set -eou pipefail

install_nats() {
  # Arch can be specified as first argument
  ARCH="${1:-amd64}"
  # Version can be specified as second argument
  VERSION="${2:-0.0.26}"
  # The name of the release file without the ".zip" prefix
  RELEASE="nats-${VERSION}-linux-${ARCH}"
  # Remove both archive and directory on exit and error
  trap "rm -Rf ${RELEASE}*" err exit
  # Construct URL of release file
  URI="https://github.com/nats-io/natscli/releases/download/${VERSION}/${RELEASE}.zip"
  # Download using wget (you could use curl or any other CLI http client)
  wget -q --show-progress "$URI"
  # Extract content
  unzip "${RELEASE}.zip"
  # Ensure nats binary is executable
  chmod +x "${RELEASE}/nats"
  # You may want to use sudo here
  mv "${RELEASE}/nats" /usr/local/bin/
}

install_nats "$@"
```

I can use this script to install on:

> No need to use sudo if you're running the script as root user

- On arm64 machines, like my laptop:
```bash
sudo ./install_nats.sh amd64
```

- On an arm64 machine, like a Raspberry Pi 4:

```bash
sudo ./install_nats.sh arm64
```

- Or an arm7 development board:

```bash
sudo ./install_nats.sh arm7
```

wallyqs (approved):

Issue #187, opened 2020-11-10
Rare case of collision when do multiples request/reply


YiuRULE:
Hi !

We are currently using the library (and thanks for the work of the contributor of the project !), and it seems when we do multiples REQ/REPLY on the same time from the client, similar at:

```
async def a(nc):
    msg = nc.request(subject="foo", payload={})
    # Do something with msg

async def b(nc):
    msg = nc.request(subject="bar", payload={})
    # Do something with msg  

await asyncio.gather(
    a(nc),
    b(nc),
)
```

We could send, on a really rare case, the same reply id to the server, who make the function `b`, receive the response from the request `a`.

Concerning our log, from the server side:

```
2020-11-09 14:17:19 ... Message(subject='foo', reply='_INBOX.Kjq4GobYoPqdOsTvlGKcbf.Kjq4GobYoPqdOuTvlGKcbf', ...)
2020-11-09 14:17:19 ... Message(subject='bar', reply='_INBOX.Kjq4GobYoPqdOsTvlGKcbf.Kjq4GobYoPqdOuTvlGKcbf', ...)
```

Concerning the environment we found the id collision, it was during internal benchmark for our application, where we had our API connecting with a nats server, as a reply and as a publisher, and an another server who spawn 1 new client each second and do two request at startup, while receiving message continuously.

YiuRULE:
(if it can help, here a bit the stacktrace of the issue I had recently for the same issue)

```
  File "/home/guillaume/.virtualenvs/pykouze/lib/python3.8/site-packages/nats/aio/client.py", line 943, in request
    del self._resp_map[token.decode()]
KeyError: 'qnaUryOVxMvt2AiVJpzTik'
```

YiuRULE:
Sorry, again me :)

Got a bit of time, and worker on a code who could reproduce the bug, who reproduce a similar environment on how we use the library.

client.py
```
import asyncio
from nats.aio.client import Client as NATS


async def run():
    nc = NATS()
    await nc.connect("nats://localhost:4222")

    await nc.request("hello", b'', timeout=1)
    await asyncio.gather(
        nc.request("foo", b'', timeout=1),
        nc.request("bar", b'', timeout=1),
    )

asyncio.run(run())
```

server.py
```
import asyncio
from nats.aio.client import Client as NATS


async def run():
    nc = NATS()
    await nc.connect("nats://localhost:4222")

    async def dummy(msg):
        await nc.publish(msg.reply, b'')

[...]

PR #228, opened 2021-11-24
Fixed NUID generator to produce the same value in subsequent calls

chiselko6:
The issue is that for `seq` being large enough (close to `MAX_SEQ`) with `inc` being equal to 62 (for 62x2 it works as well, but for 62xK for K > 2 the minimum required value for `seq` to reproduce the issue is more than `MAX_SEQ`) two subsequent calls to `NUID.next` produce the same value. This happens due to float division losing precision.

### DIFF
--- a/nats/nuid.py
+++ b/nats/nuid.py
@@ -53,7 +53,7 @@ class NUID:
         suffix = bytearray(SEQ_LENGTH)
         for i in reversed(range(SEQ_LENGTH)):
             suffix[i] = DIGITS[int(l) % BASE]
-            l /= BASE
+            l //= BASE
 
         prefix.extend(suffix)
         return prefix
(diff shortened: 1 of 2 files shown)


wanfranck:
Great job @chiselko6! LGTM!

vshender:
This solution actually solves the problem described in https://github.com/nats-io/nats.py/issues/187.

aduhovnik:
@chiselko6 good call!

Skyross:
Nice fix, @chiselko6!

wallyqs (approved):

wallyqs:
Thanks a lot for the fix @chiselko6 ! I will backport this to v0 branch as well.

wallyqs:
This is now available in v0.11.5 release as well. Thanks again!

PR #242, opened 2021-12-19
Add subject, sid, and sub to SlowConsumerError for the ease of debugging

KMilhan:
Thank you very much maintainers for giving us such a wonderful project.

One day we happened to have multiple thousands of error messages of 
```
nats: slow consumer, messages dropped
```
while we using 2.0.0-rc3.

It was difficult to debug as the messages didn't have the subject and sid. This small change may save us in the future.


wallyqs (approved):
LGTM

PR #244, opened 2021-12-19
Accept config and parameters in add_stream

KMilhan:
This change make add_stream accept both `config` and `params`.
Current code expects only one side, so it caused unexpected
behavior. For example,
```python
>>> await add_stream(config=common_config, name="this_subject")

ValueError("nats: stream name is required")
```

This change allows the above code.

Thank you for your review maintainers.


wallyqs (approved):
LGTM

PR #245, opened 2021-12-19
Add missing `self` argument and `classmethod` wrapper

KMilhan:
_Sorry for many numbers of PR :pray:_

This change adds missing `cls` or `self` to JetStream client at `nats/js/client.py` to fix below.

```python
>>> active_subscription = await js.pull_subscribe(
    subject=subject,
    stream=stream,
    durable=durable,
)
>>> await active_subscription.unsubscribe()
TypeError: unsubscribe() takes 0 positional arguments but 1 was given
```

### DIFF
--- a/nats/js/client.py
+++ b/nats/js/client.py
@@ -375,7 +375,8 @@ class JetStream:
            api.StatusHdr in msg.header:
             return msg.header[api.StatusHdr]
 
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks for the many fixes!

KMilhan:
Thank you for the quick review and merge :+1:

PR #248, opened 2021-12-30
Fix minor type hinting including `Msg`

KMilhan:
`Msg` had a type hinting `str` for `data`. Which misled my understanding at some point.
I ran `mypy` and found similar errors and here's a small fix.
Thank you very much for your review.


wallyqs (approved):
Thanks!

PR #253, opened 2022-01-12
Fix type errors in /nats/js/

orsinium:
The PR makes `mypy ./nats/js/` pass without errors. The diff may seem scary but there should be no backward-incompatible changes, except merging the `JetStream` class into `JetStreamContext`.

I want to bring later more extensive type annotations.

BTW, flake8 now passes too.

### DIFF
--- a/nats/js/manager.py
+++ b/nats/js/manager.py
@@ -25,7 +25,6 @@ class JetStreamManager:
     """
     JetStreamManager exposes management APIs for JetStream.
     """
-
     def __init__(
         self,
         conn,
--- a/nats/js/manager.py
+++ b/nats/js/manager.py
@@ -13,33 +13,26 @@
 #
 
 import json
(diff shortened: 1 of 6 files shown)


nats/js/manager.py
```diff
@@ -197,34 +189,11 @@ async def delete_consumer(self, stream=None, consumer=None):
         )
         return resp['success']
 
-    async def get_last_msg(
```
orsinium, line 200: I've moved it into `JetStreamContext` because it uses `self._jsm`

wallyqs: sounds good, thanks!

wallyqs (approved):
LGTM!

orsinium:
Oh wow, already merged. Thank you, that's a pleasure to contribute to your project :)

PR #255, opened 2022-01-13
Improve type annotations in nats/js/

orsinium:
1. Add all missed type annotations
2. Convert some classes into `dataclass`
3. replace `JetStreamContext.delete_key_value` by `KeyValue.destroy`
4. Merge `KeyValueManager` into `JetStreamContext`. It uses a lot of things from `JetStreamContext` and so cannot exist on its own. I'd decouple some bits of `JetStreamContext` but it needs an actual decoupling, probably with a smart composition.
5. Remove `class _JS()`, it's not used anywhere.

### DIFF
--- a/nats/aio/msg.py
+++ b/nats/aio/msg.py
@@ -14,18 +14,28 @@
 
 import datetime
 from dataclasses import dataclass
-from typing import Optional
+from typing import TYPE_CHECKING, Dict, List, Optional
 from nats.errors import Error, NotJSMessageError, MsgAlreadyAckdError
 
 
+if TYPE_CHECKING:
+    from nats import NATS
+
+
+@dataclass
 class Msg:
     """
     Msg represents a message delivered by NATS.
     """
-    __slots__ = (
-        'subject', 'reply', 'data', 'sid', '_client', 'headers', '_metadata',
-        '_ackd'
-    )
+    client: "NATS"
+    subject: str = ''
+    reply: str = ''
+    data: bytes = b''
+    sid: int = 0
+    headers: Optional[Dict[str, str]] = None
+
+    _metadata: Optional["Metadata"] = None
+    _ackd: bool = False
 
     class Ack:
         Ack = b"+ACK"
@@ -57,24 +67,6 @@ class Msg:
         #
         V2TokenCount = 12
 
-    def __init__(
-        self,
-        subject: str = '',
-        reply: str = '',
-        data: bytes = b'',
-        sid: int = 0,
-        client=None,
-        headers: dict = None
-    ) -> None:
-        self.subject = subject
-        self.reply = reply
-        self.data = data
-        self.sid = sid
-        self.headers = headers
-        self._client = client
-        self._metadata = None
-        self._ackd = False
-
     @property
     def header(self) -> Optional[dict]:
         """
@@ -82,31 +74,31 @@ class Msg:
         """
         return self.headers
 
-    async def respond(self, data: bytes):
+    async def respond(self, data: bytes) -> None:
         """
         respond replies to the inbox of the message if there is one.
         """
         if not self.reply:
             raise Error('no reply subject available')
-        if not self._client:
+        if not self.client:
             raise Error('client not set')
 
-        await self._client.publish(self.reply, data, headers=self.headers)
+        await self.client.publish(self.reply, data, headers=self.headers)
 
     async def ack(self) -> None:
         """
         ack acknowledges a message delivered by JetStream.
         """
         self._check_reply()
-        await self._client.publish(self.reply)
+        await self.client.publish(self.reply)
         self._ackd = True
 
-    async def ack_sync(self, timeout: float = 1.0):
+    async def ack_sync(self, timeout: float = 1.0) -> "Msg":
         """
         ack_sync waits for the acknowledgement to be processed by the server.
         """
         self._check_reply()
-        resp = await self._client.request(self.reply, timeout=timeout)
+        resp = await self.client.request(self.reply, timeout=timeout)
         self._ackd = True
         return resp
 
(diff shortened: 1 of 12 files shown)


nats/aio/msg.py
```diff
@@ -14,17 +14,27 @@
...
 import datetime
 from dataclasses import dataclass
+from typing import TYPE_CHECKING, Dict, List, Optional
 from nats.errors import Error, NotJSMessageError, MsgAlreadyAckdError
 
+if TYPE_CHECKING:
+    from nats import NATS
 
+
+@dataclass
```
wallyqs, line 24: I really like the `dataclass` annotations, do you think there might be some performance impact/benefits from using it for `Msg`?

nats/aio/msg.py
```diff
@@ -14,17 +14,27 @@
...
 class Msg:
     """
     Msg represents a message delivered by NATS.
     """
-    __slots__ = (
-        'subject', 'reply', 'data', 'sid', '_client', 'headers', '_metadata',
-        '_ackd'
-    )
+    client: "NATS"
```
wallyqs, line 29: I think fine either way, but why the rename of `_client` into `client`?

nats/aio/msg.py
```diff
@@ -113,29 +106,31 @@ async def nak(self) -> None:
...
         term terminates a message delivered by JetStream and disables redeliveries.
         """
         self._check_reply()
 
-        await self._client.publish(self.reply, Msg.Ack.Term)
+        await self.client.publish(self.reply, Msg.Ack.Term)
         self._ackd = True
 
+    # TODO(@orsinium): use a cached_property. Available in functools since 3.8,
+    # as a package (backports.cached-property), or can be just copy-pasted in the project.
```
wallyqs, line 131: sounds good to constrain to 3.8 since that is the version where `loop` parameter got deprecated.

nats/js/kv.py
```diff
@@ -183,81 +178,9 @@ async def status(self) -> BucketStatus:
...
-            stream=stream.name,
-            pre=f"$KV.{config.bucket}.",
-            js=self,
-        )
-
-    async def delete_key_value(self, bucket: str):
-        """
-        delete_key_value deletes a JetStream KeyValue store by destroying
-        the associated stream.
+    async def destroy(self) -> bool:
```
wallyqs, line 183: I think this is a good addition but we should also let `delete_key_value` to more or less conform to the APIs in the Go client just in case: https://github.com/nats-io/nats.go/blob/b337a5c01afdab41a22becdf3759e038f9948889/kv.go#L35

wallyqs (commented):
Overall LGTM, just a couple of questions.

nats/aio/msg.py
orsinium: Great question! The dataclasses have `slots` but, unfortunately, only From 3.10, which is not an option just yet for an OSS. So, we can use it only without `slots`. The slotted classes are a little bit faster at attribute lookup and take a bit less memory (minus one dict, basically). You can get a similar benefit by using `typing.NamedTuple`. However, there are a few minor differences:

1. NamedTuple is immutable. You can do the same for dataclasses by setting `frozen=True` but you cannot switch it off for a NamedTuple. It is convenient, sometimes.
2. NamedTuple doesn't allow to use `functools.cached_property`. And this is **very** convenient thing, I can't overstate how nicer and faster it makes the code.
3. NamedTuple behaves like a tuple. It generates `__iter__` and some methods like `count` which may be undesired and confusing for a public API.

Now, back to "the old approach vs dataclasses". My motivation was mostly the convenience of the development since I find the associated performance impact negligible. Dataclasses provide us with:

1. Autogeneration of `__init__`, `__repr__`, and `__eq__`.
2. Again, the ability to use `cached_property`

Now, let's talk about benchmarks. There is  the code I've tested:

```python
from dataclasses import dataclass
class A:
    __slots__ = ['a']
    def __init__(self, a):
        self.a = a
@dataclass
class B:
    a: int

a = A(1)
b = B(1)
```

And there are my results:

```python
%timeit A(1) # 152 ns ± 4.98 ns per loop
%timeit B(1) # 228 ns ± 4.13 ns per loop
%timeit a.a  # 30.8 ns ± 1.1 ns per loop
%timeit b.a  # 33.9 ns ± 1.12 ns per loop
%timeit a.a = 2  # 33.9 ns ± 3.27 ns per loop
%timeit b.a = 2  # 43.7 ns ± 3.87 ns per loop
```

As you can see, there is indeed a difference. Is it negligible? Definitely yes for the lookup operation, and we don't care about assignment one, AFAIK the message is not supposed to be changed. The instantiation, however, gives a noticeable difference, at least in my example. Is it tolerable? I don't know, actually. Depends on the philosophy of the project and where it's heading. For example, I expect it to be even faster if compiled with mypyc. But dataclasses are definitely shorter.

nats/aio/msg.py
orsinium: Dataclasses don't allow you to underscore whatever fields you want to pass in `__init__`. You either keep it internal-only ("private", in some sense) or make it available for bot outside initialization and outside access. If we decide to go with the old approach and drop dataclasses, I'll revert it back as well.

nats/js/kv.py
orsinium: My motivation to change it was that `delete_key_value` didn't work and I wasn't sure how it was supposed to 🙃 Thank you for pointing to the Go client, I'll just have a look at it and do the same here.

nats/aio/msg.py
orsinium: I don't want to bring so radical changes just yet. Let's break things in a separate PR 😉

nats/aio/msg.py
wallyqs: Thanks for the explanation!

nats/aio/msg.py
wallyqs: I'd rather keep it 'private' as `_client` in this case, maybe just set it outside of `build_message` and remove it as a parameter for `__init__`?
```
diff --git a/nats/aio/client.py b/nats/aio/client.py
index a41bfb9..e42f348 100644
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1413,6 +1413,7 @@ class Client:
                 return
 
         msg = self._build_message(subject, reply, data, hdr)
+        msg._client = self
         if not msg:
             return
 
@@ -1513,7 +1514,6 @@ class Client:
             reply=reply.decode(),
             data=data,
             headers=headers,
-            client=self
         )
```

orsinium: Great news! I was wrong. Turned out, you can pass underscored attributes into a dataclass. Did I confuse it with [attrs](https://attrs.org/)? 🤔 Anyway, I've fixed it now.

wallyqs (approved):
LGTM! Thanks again

PR #258, opened 2022-01-18
Run flake8 and mypy on CI

orsinium:
The PR adds [mypy](http://mypy-lang.org/) and [flake8](https://flake8.pycqa.org/en/latest/) on CI.

+ **mypy** is a type checker. It finds some type misuse bugs (I've fixed a lot of issues in my previous PRs there, so it's actually helpful) and makes sure you don't lie in your type annotations (and type annotations are important because there are user-facing and part of the library API).
+ **flake8** finds some trivial bugs and makes sure you follow [PEP-8](https://www.python.org/dev/peps/pep-0008/). For example, I previously fixed there bare `try-except` blocks (by replacing `except:` with `except Exception:`) so it won't suppress `sys.exit`, kill signals, and KeyboardInterrupt (and you never want to suppress any of it in a library).

Currently, they check only `nats/js` because this is the only part of the project I fixed so far. And the only part that allows going a bit wild since it's not in a major release yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ test:
 	yapf --recursive --diff $(SOURCE_CODE)
 	yapf --recursive --diff tests
 	mypy
+	flake8 ./nats/js/
 	pytest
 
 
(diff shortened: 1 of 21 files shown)


wallyqs (approved):
LGTM, thanks :)

PR #259, opened 2022-01-19
Improve JS API

orsinium:
1. Expand starred imports in nats/aio/
2. Turn Status and Header constants in nats/js into Enums
3. Simplify a bit some methods in nats/js/client.py
4. Use SCREAMING_SNAKE_CASE for Enum instances, [as in the official documentation](https://docs.python.org/3/library/enum.html).
5. Remove some unused variables


wallyqs (approved):
LGTM

PR #261, opened 2022-01-21
Improve JS API

orsinium:
1. The Nats API uses nanoseconds for time, Python uses seconds. So, some classes in `nats/js` had time in nanoseconds, some in seconds. I've made them all store time in seconds and convert it into nanoseconds in `as_dict` result.
2. `Base.loads` silently drops unknown fields. While it is great for API responses (an open-world assumption), it can be harmful for the cases where the user passes a value explicitly. So, I renamed the method into `from_response` and use it only for parsing API responses. The `evolve` method now uses `dataclasses.replace` instead. Everything that was in `__post_init__` also migrated into `from_response` since it is also applicable only for API responses, not for user-provided data. It also helped to avoid double-conversion of time from nanoseconds.

### DIFF
--- a/nats/js/client.py
+++ b/nats/js/client.py
@@ -623,7 +623,7 @@ class JetStreamContext(JetStreamManager):
             )
             return info
 
-        async def fetch(self, batch: int = 1, timeout: int = 5) -> List[Msg]:
+        async def fetch(self, batch: int = 1, timeout: float = 5) -> List[Msg]:
             """
             fetch makes a request to JetStream to be delivered a set of messages.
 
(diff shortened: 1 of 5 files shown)


nats/js/client.py
```diff
@@ -243,12 +242,6 @@ async def cb(msg):
                 config = api.ConsumerConfig(
                     ack_policy=api.AckPolicy.EXPLICIT,
                 )
-            elif isinstance(config, dict):
-                config = api.ConsumerConfig.loads(**config)
-            elif not isinstance(config, api.ConsumerConfig):
-                raise ValueError("nats: invalid ConsumerConfig")
-            assert config is not None
```
orsinium, line 250: I decided to drop type checks there. If the type annotation says that `config` is a `ConsumerConfig`, it is the responsibility of the user (with the help of mypy) to ensure that's true.

wallyqs: I liked being able to pass a dict for the ConsumerConfig instead, but I agree and instead need to add more options like in the Go client in case want to customize without using the type.

wallyqs (approved):
LGTM 👍

nats/js/client.py
orsinium: Well, we can always support both and annotate it as `Union[ConsumerConfig, Dict[str, Any]]`. I'd keep it simple, though. Accepting `ConsumerConfig` makes it type-safe, gives the user a good autocomplete, points to the documentation, etc. A bit more verbose but it is Python, not Perl, here we keep things verbose if it is better for the reader. "Explicit is better than implicit".

In many places, we still accept `config, **params`, though, so the shorter version is still here, just not for this method. But again, having `**params: Any` will allow passing any arguments without type errors from the type checker and failing at runtime, which is not cool. So, I think we should have `**params` only for the methods that don't accept any other arguments except `config`, so passing any keyword arguments is a deliberate choice rather than a typo.

PR #262, opened 2022-01-24
Sort imports

orsinium:
1. Add [isort](https://pycqa.github.io/isort/index.html) config
2. Make it compatible with yapf
3. Sort all imports


wallyqs (approved):
LGTM

PR #266, opened 2022-02-02
JetStream: decouple pull_subscribe_bind from pull_subscribe

orsinium:
Now we're getting to the root issue that motivated me to contribute into the library ;) The `pull_subscribe` method does a bit too much. It requests the consumer info from the server just to check if a consumer is created. If not, it creates one. There are multiple issues with this approach:

1. It sends one more request than it could. Since I'm explicitly creating the consumer before, it never has an effect in my code but I cannot avoid it.
1. The `config` argument is used only if the consumer doesn't exist.
1. The `subject` argument is used only if the consumer doesn't exist or stream is not passed.
1. The `stream` argument is kinda optional but not passing it implicitly sends a request.
1. If the `config` is passed explicitly, some of its parameters will be overwritten in place.

The method is easy to use incorrectly, and when you do, it sends 4 requests instead of just one.

To mitigate all these issues, I've decoupled the last step of the method into a separate method, called `pull_subscribe_fast`. It requires you to explicitly create a consumer beforehand but it does much fewer requests and has a much cleaner signature.

The name isn't good, though. What is even "fast"? So, I'm open to better ideas. Or alternatively we could just drop the original `pull_subscribe`, delegating the consumer creation logic to the caller. It's more verbose but "explicit is better than implicit".


orsinium:
Oh no, 2.0.0 is already released, I'm late for breaking changes 😱 Well, let's keep it backward-compatible then. This PR is 100% backward-compatible as it is now.

nats/js/client.py
```diff
@@ -362,22 +357,54 @@ async def main():
...
             config.filter_subject = subject
             config.durable_name = durable
             await self._jsm.add_consumer(stream, config=config)
 
-        # FIXME: Make this inbox prefix customizable.
-        deliver = api.InboxPrefix[:]
-        deliver.extend(self._nc._nuid.next())
+        return await self.pull_subscribe_fast(durable=durable, stream=stream)
+
+    async def pull_subscribe_fast(
```
wallyqs, line 368: There is a somewhat similar option in the nats.go client called `nats.Bind`: https://github.com/nats-io/nats.go/blob/main/js.go#L2195-L2197
How about renaming to `pull_subscribe_bind` to keep the semantics more or less the same?

wallyqs (approved):
LGTM, just a suggestion on renaming to something similar to Go client option.

nats/js/client.py
orsinium: Great idea! Fixed.

PR #274, opened 2022-02-10
JetStream: decouple subscribe_bind from subscribe

orsinium:
Do the same as #266 but for JetStream.subscribe.

### DIFF
--- a/nats/__init__.py
+++ b/nats/__init__.py
@@ -11,11 +11,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from typing import List, Union
 
 from .aio.client import Client as NATS
 
 
-async def connect(servers=["nats://localhost:4222"], **options) -> NATS:
+async def connect(
+    servers: Union[str, List[str]] = ["nats://localhost:4222"],
+    **options
+) -> NATS:
     """
     :param servers: List of servers to connect.
     :param options: NATS connect options.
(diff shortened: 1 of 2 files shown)


nats/js/client.py
```diff
@@ -271,17 +274,36 @@ async def cb(msg):
...
+        return await self.subscribe_bind(
+            cb=cb,
+            stream=stream,
+            config=config,
+            manual_ack=manual_ack,
+            ordered_consumer=ordered_consumer,
+            consumer=consumer,
+        )
+
+    async def subscribe_bind(
```
wallyqs, line 290: Think we need a test just for when using `subscribe_bind`, using `add_consumer` to create a durable consumer then bind to it to receive messages.

orsinium: Could you have a look at https://github.com/nats-io/nats.py/pull/274/commits/4d83ef72814c6c25b4fc211d002498e6c2671c42 , please? The subscriber doesn't receive any messages and dies with TimeoutError and I'm not sure how to debug it. Any advice?

wallyqs: Thanks, will take a look

orsinium: Thank you. Just in case, I've added you write access to the fork. This time, I forgot to create a branch for PR, so you'll need explicit access to write into `main`.

wallyqs: Hi @orsinium, I've just merged the branch with the fixed test at https://github.com/nats-io/nats.py/commit/d644c581560b838acada65a75c40cf07e8cdb369
What was needed is to setup the `deliver_subject` as part of the consumer config so that bind would succeed later.

PR #275, opened 2022-02-11
Use pytest-style assertions in test_js.py

orsinium:
1. Use pytest-style `assert` assertions instead of unittest-style `self.assertEqual` assertions. You already use pytest for running tests, so why don't make one more step towards beautiful tests?
2. Drop some `try: ... except: pass` from tests. These are tests, why silently suppress errors? Tests should detect errors, not ignore. If any of these errors are supposed to be ignored and will pop up, it's never too late to return these suppressions but with the expected error specified more precisely, like `except nats.errors.TimeoutError:`.
3. Drop some unused variables and imports.


wallyqs (approved):
Looks great! LGTM

PR #276, opened 2022-02-15
Add inbox prefix as a configuration variable

YiuRULE:
We can configure the inbox prefix on many other nats client implementation, including the go client (https://github.com/nats-io/nats.go/blob/161162cae0b290287b5e2246732c0075e3724497/nats.go#L1143-L1152) or the java client (https://github.com/nats-io/nats.java/blob/a551760a10e620049fb5f4848d104e96fba4ed0f/src/main/java/io/nats/client/Options.java#L572).

The point of this PR is to reproduce the same behaviour on the python client


wallyqs (approved):
LGTM! Thanks for the contribution

PR #278, opened 2022-02-21
nats: add typing to main connect function

squat:
Currently, the `connect` functions of the main NATS modules are not
fully typed or typed slightly incorrectly. The examples all show that a
single string can be given for the `servers` parameter, but the `Client`
class does not declare that it accepts this type, which causes static
analysis issues when the module is used directly.

This commit extends the signatures for `connect` so that `servers` can
also be a `str`, effectively giving it the same type as what is expected
by `_setup_server_pool`.

Signed-off-by: Lucas Servén Marín <lserven@gmail.com>

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -219,7 +219,7 @@ class Client:
 
     async def connect(
         self,
-        servers: Union[str, List[str]] = ["nats://127.0.0.1:4222"],
+        servers: Union[str, List[str]] = ["nats://localhost:4222"],
         error_cb: Optional[ErrorCallback] = None,
         disconnected_cb: Optional[Callback] = None,
         closed_cb: Optional[Callback] = None,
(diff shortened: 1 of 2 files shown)


nats/aio/client.py
```diff
@@ -219,7 +219,7 @@ def __init__(self) -> None:
 
     async def connect(
         self,
-        servers: List[str] = ["nats://127.0.0.1:4222"],
+        servers: Union[str, List[str]] = ["nats://127.0.0.1:4222"],
```
squat, line 222: btw, what do you think about adding another commit so that both the `connect` methods use the same default, either `localhost` or `127.0.0.1`?

wallyqs: Thanks, maybe let's use `localhost` for both? wdyt?

wallyqs (approved):
LGTM

nats/aio/client.py
squat: Agreed, this way we are not defaulting to IPv4. I'll add another commit to address that.

squat: ok ready :+1:

wallyqs (approved):

PR #285, opened 2022-03-05
Fix connected_url to return ParseResult

nosammai:
Fixes connected_url property to return the ParseResult object instead of a string representation like `'ParseResult(scheme=\'nats\', netloc=\'localhost:4222\', path=\'\', params=\'\', query=\'\', fragment=\'\')'`

This was broken in a recent update, and makes the examples that print `nc.connected_url.netloc` throw an Exception: `AttributeError: 'str' object has no attribute 'netloc'`

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -985,9 +985,9 @@ class Client:
             raise errors.TimeoutError
 
     @property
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM!

Issue #292, opened 2022-04-08
ssl_context referenced before assignment


Basketballforce:
On python 3.9.

I was testing the quickstart/examples found on nat.io and the github repo but, ran into this issue:

When running the example, an UnboundLocalError occurs in nats/aio/client.py stating that local variable ssl_context referenced before assignment

The odd thing is the specific line from the error traceback if from line 1741 in which 

`if ssl_context is None`

Any insight on why this is occuring?

Thank you!

Basketballforce:
Unrelated Sidenote: If running the python client in windows and you get a "event loop closed error" you can resolve this by adding

`asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())`

Before calling asyncio functions/calls

wallyqs:
Fixed via #293

Basketballforce:
Thanks! :)

PR #293, opened 2022-04-16
fix('#292: ssl_context referenced before assignment')

marianojabdala:
This commit fix the issue with ssl_context and also other issues reported by yapf


wallyqs (approved):

wallyqs:
Thanks for the contribution!

Issue #308, opened 2022-05-08
TypeError: __init__() missing 1 required positional argument: 'cluster' when connecting to KV


olgeni:
Hello,

I was trying to connect to a KV store for a quick check (names redacted):

    import asyncio
    import ssl

    import nats

    ssl_ctx = ssl.create_default_context()  # will not work without

    async def test():
        nc = await nats.connect('nats://my.public.hostname,
                                tls=ssl_ctx,
                                user_credentials='/path/to/creds')
        js = nc.jetstream()
        kv = await js.key_value('presence')

    async def loop():
        await test()

        while True:
            await asyncio.sleep(5)

    asyncio.run(loop())

But I get this exception:

    Traceback (most recent call last):
      File "/Users/olgeni/src/ntest/ntest/nats_test.py", line 24, in <module>
        asyncio.run(loop())
      File "/usr/local/Cellar/python@3.9/3.9.12/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
        return loop.run_until_complete(main)
      File "/usr/local/Cellar/python@3.9/3.9.12/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
        return future.result()
      File "/Users/olgeni/src/ntest/ntest/nats_test.py", line 18, in loop
        await test()
      File "/Users/olgeni/src/ntest/ntest/nats_test.py", line 14, in test
        kv = await js.key_value('presence')
      File "/Users/olgeni/src/ntest/.venv/lib/python3.9/site-packages/nats/js/client.py", line 919, in key_value
        si = await self.stream_info(stream)
      File "/Users/olgeni/src/ntest/.venv/lib/python3.9/site-packages/nats/js/manager.py", line 68, in stream_info
        return api.StreamInfo.from_response(resp)
      File "/Users/olgeni/src/ntest/.venv/lib/python3.9/site-packages/nats/js/api.py", line 287, in from_response
        cls._convert(resp, 'config', StreamConfig)
      File "/Users/olgeni/src/ntest/.venv/lib/python3.9/site-packages/nats/js/api.py", line 66, in _convert
        resp[field] = type.from_response(data)
      File "/Users/olgeni/src/ntest/.venv/lib/python3.9/site-packages/nats/js/api.py", line 241, in from_response
        cls._convert(resp, 'placement', Placement)
[...]

PR #310, opened 2022-05-08
Make 'cluster' optional in 'Placement' class.

olgeni:
Fix for issue #308, where you would get a `TypeError: __init__() missing 1 required positional argument: 'cluster'` when connecting to a KV bucket ☺️


wallyqs:
Thanks!

PR #321, opened 2022-06-09
Fix uncaught asyncio.TimeoutError

### LINKED ISSUE #320: Client fails to reconnect to the cluster on some cases
Hi, I'm using nats-py 2.1.3 and I discovered an issue where the client fails to reconnect to the NATS cluster.

My cluster has 3 servers: `nats-1`, `nats-2` and `nats-3`. When the client is connected to the server, I send a SIGSTOP signal to the `nats-1` server. I expect the client to reconnect to one of the other servers in the cluster, but it only works 50% of the time.

In the `_attempt_reconnect` function, the client attempts to reconnect to the next server in the pool. The server pool looks like this: `['nats://172.20.0.2:4222', 'nats://localhost:4222', 'nats://172.20.0.4:4222', 'nats://172.20.0.3:4222']` (but the order is random, note that both `nats://172.20.0.2:4222` and `nats://localhost:4222` correspond to the stopped `nats-1` server).

Here there are two cases: 
- the next server picked by the client corresponds to `nats-2`, or `nats-3`, everything works as expected, the client reconnects to another server of the cluster
- the next server picked by the client corresponds to `nats-1`, which is stopped. The client tries to connect here https://github.com/nats-io/nats.py/blob/a649850f33d3ed39874e687208e70c95f2574763/nats/aio/client.py#L1760. But it fails, and a `asyncio.Timeout` exception is raised.

The problem is that `_attempt_reconnect` should catch this exception and retry connecting to the next server in the pool, and I think that was the intended behavior.

https://github.com/nats-io/nats.py/blob/a649850f33d3ed39874e687208e70c95f2574763/nats/aio/client.py#L1377
Here we can see that `_attempt_reconnect` catches `TimeoutError`, but not `asyncio.TimeoutError`, which is why the reconnection doesn't work.

## How to reproduce the bug

I have a simple script that connects to a NATS server : 
```python
import asyncio
import time
import nats

async def main():
    nc = await nats.connect(
        "localhost:4222",
        ping_interval=1,
        max_outstanding_pings=2,
        disconnected_cb=on_disconnect,
        reconnected_cb=on_reconnect,
        closed_cb=on_closed,
        error_cb=on_error,
    )
    
    while True:
        try:
            await asyncio.sleep(1)
        except KeyboardInterrupt:
            break

    await nc.close()

async def on_disconnect():
    print("disconnected")

async def on_reconnect():
    print("reconnected")

async def on_closed():
    print("closed")

async def on_error(e: Exception):
    print("error", e)

if __name__ == '__main__':
    asyncio.run(main())
```

My NATS cluster setup with docker-compose : 
```yaml
version: '3'

services:
  nats-1:
    image: nats:2.8.2
    restart: unless-stopped
    volumes:
      - ./config.conf:/etc/nats/config.conf
    command: --config /etc/nats/config.conf --name nats-1
    ports:
    - 127.0.0.1:4222:4222
  nats-2:
    image: nats:2.8.2
    restart: unless-stopped
    volumes:
      - ./config.conf:/etc/nats/config.conf
    command: --config /etc/nats/config.conf --name nats-2
    ports:
    - 127.0.0.1:4223:4222
  nats-3:
    image: nats:2.8.2
    restart: unless-stopped
    volumes:
      - ./config.conf:/etc/nats/config.conf
    command: --config /etc/nats/config.conf --name nats-3
    ports:
    - 127.0.0.1:4224:4222
```

With this `config.conf`: 
```
listen: 0.0.0.0:4222
http: 0.0.0.0:8222

cluster {
  name: "nats_cluster"
  listen: 0.0.0.0:6222

  routes = [
    "nats://nats-1:6222",
    "nats://nats-2:6222",
    "nats://nats-3:6222"
  ]
}

jetstream {
  store_dir=/var/lib/nats/data
}
```

ozon2:
Fix #320.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1374,7 +1374,7 @@ class Client:
                 self._err = e
                 await self.close()
                 break
-            except (OSError, errors.Error, TimeoutError) as e:
+            except (OSError, errors.Error, asyncio.TimeoutError) as e:
                 self._err = e
                 await self._error_cb(e)
                 self._status = Client.RECONNECTING


wallyqs:
Thanks for the investigation and the fix!

PR #326, opened 2022-07-06
Add "num_replicas" field for consumer

goku321:
This adds `num_replicas` field to allow configuring the number of replicas for a durable consumer.

### DIFF
--- a/tests/test_js.py
+++ b/tests/test_js.py
@@ -1590,7 +1590,7 @@ class ConsumerReplicasTest(SingleJetStreamServerTestCase):
 
 
         # Create consumer
-        config = nats.js.api.ConsumerConfig(num_replicas=1)
+        config = nats.js.api.ConsumerConfig(num_replicas=1, durable_name="mycons")
         cons = await js.add_consumer(stream="TESTREPLICAS", config=config)
 
         assert cons.config.num_replicas == 1
--- a/tests/test_js.py
+++ b/tests/test_js.py
@@ -1577,3 +1577,22 @@ class KVTest(SingleJetStreamServerTestCase):
 
         with pytest.raises(BadBucketError):
             await js.key_value(bucket="TEST3")
+
+class ConsumerReplicasTest(SingleJetStreamServerTestCase):
+    @async_test
+    async def test_number_of_consumer_replicas(self):
+        nc = await nats.connect()
+
+        js = nc.jetstream()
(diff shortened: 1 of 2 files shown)


goku321:
Hello @wallyqs . Would really appreciate if you can review this. A bit of urgent for us as we are moving to production and would want to be able to create R1 consumers.

tests/test_js.py
```diff
@@ -1577,3 +1577,22 @@ async def error_handler(e):
...
+
+class ConsumerReplicasTest(SingleJetStreamServerTestCase):
+    @async_test
+    async def test_number_of_consumer_replicas(self):
+        nc = await nats.connect()
+
+        js = nc.jetstream()
+        await js.add_stream(name="TESTREPLICAS", subjects=["test.replicas"])
+        for i in range(0, 10):
+            await js.publish("test", f'{i}'.encode())
```
wallyqs, line 1589: so that it receives messages, has to be following subject:
```suggestion
            await js.publish("test.replicas", f'{i}'.encode())
```

goku321: Thanks. Fixed the subject name

wallyqs:
Thanks, tests are passing for my locally but there is format issue that making the CI fail. I can take a look later also before merging and releasing later today.

tests/test_js.py
```diff
@@ -1577,3 +1577,22 @@ async def error_handler(e):
...
         with pytest.raises(BadBucketError):
             await js.key_value(bucket="TEST3")
+
+class ConsumerReplicasTest(SingleJetStreamServerTestCase):
+    @async_test
+    async def test_number_of_consumer_replicas(self):
+        nc = await nats.connect()
+
+        js = nc.jetstream()
+        await js.add_stream(name="TESTREPLICAS", subjects=["test.replicas"])
```
shiv4289, line 1587: Hi @wallyqs Is there a way to make this stream R3 here?
This would strengthen the test so the streams is R3 and consumer is R1 and we can be 100% sure the replication factor for consumer is not inherited from the stream.

wallyqs: I'd like that too, would have to make the tests run in cluster mode and right now they only run in single node (no helper for that in the tests atm)

wallyqs:
Included with format fixes as part of #327 
Thanks for the contribution!

PR #328, opened 2022-07-07
Fix header line to be compatible again with other clients

wallyqs:
CRLF required in other clients parsers so this helps with compatibility with the spec.

Signed-off-by: Waldemar Quevedo <wally@nats.io>


pohlt:
Thanks a lot, @wallyqs ! 🥳

PR #331, opened 2022-08-11
[WIP]: Initial ObjectStore client implementation

domderen:
This pull request contains a JetStream ObjectStore client implementation. It is an early draft that at the moment contains following functionalities:

- Create Object Store,
- Get Object Store,
- Delete Object Store,
- Put Object,
- Get Object,

It was re-implemented from the golang code in [nats.go](https://github.com/nats-io/nats.go) repo. 

**This code is not complete yet!** I'm sharing it in this form to gather some initial feedback if this is the right direction for this implementation. Any feedback is welcome :)

I think it would be good to implement ObjectStore calls around Bytes.IO types, to allow handling streams, same as the golang version does, but I went with simpler bytes implementation for now. I can try to extend it later on if it will look like a right direction.


domderen:
Hey @wallyqs it would be great to hear your thoughts on this PR.

wallyqs:
Thanks @domderen for the contribution and the rebase. Sorry for the delay, I will start looking today.

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+            info = api.ObjectInfo(
+                name=meta.name,
+                description=meta.description,
+                headers=meta.headers,
+                options=meta.options,
+                bucket=self._name,
+                nuid=id.decode(),
+                size=0,
+                chunks=0,
+                mtime=0
```
tothandras, line 266:
```python
from datetime import datetime, timezone
# ...
```

```suggestion
                mtime=datetime.now(timezone.utc).isoformat(),
```

tothandras (commented):
I've started to check the cross language compatibility of the implementations, first with Go and I plan to do the same with JavaScript.
(I have a use case where I write objects in Python and read them from JavaScript.)

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+        stream: str,
+        js: "JetStreamContext",
+    ) -> None:
+        self._name = name
+        self._stream = stream
+        self._js = js
+
+    def __sanitize_name(self, name: str) -> str:
+        name = name.replace(".", "_")
+        return name.replace(" ", "_")
```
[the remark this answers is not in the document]
tothandras: this might run multiple times, fix needed

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+            result.data += msg.data
+            h.update(msg.data)
+
+            # Check if we are done.
+            if tokens[8] == OBJ_NO_PENDING:
+                await sub.unsubscribe()
+
+                # Make sure the digest matches.
+                sha = h.digest()
+                digest_str = info.digest.replace(OBJ_DIGEST_TYPE, "")
```
tothandras, line 215:
```suggestion
                digest_str = info.digest.replace(OBJ_DIGEST_TYPE, "").replace(OBJ_DIGEST_TYPE.upper(), "")
```

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def __sanitize_name(self, name: str) -> str:
+        name = name.replace(".", "_")
+        return name.replace(" ", "_")
+
+    async def get_info(self, name: str) -> api.ObjectInfo:
+        """
+        GetInfo will retrieve the current information for the object.
+        """
+        obj = self.__sanitize_name(name)
```
tothandras, line 143:
```suggestion
```

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+
+        if not key_valid(obj):
+            raise InvalidObjectNameError
+
+        meta = OBJ_META_PRE_TEMPLATE.format(bucket=self._name, obj=obj)
```
tothandras, line 148:
```suggestion
        if not key_valid(name):
            raise InvalidObjectNameError

        meta = OBJ_META_PRE_TEMPLATE.format(bucket=self._name, obj=name)
```

nats/js/object_store.py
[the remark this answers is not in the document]
tothandras: looking more into this, name was correct, only the meta subject is base64 encoded

nats/js/object_store.py
tothandras, line 148:
```suggestion
        meta = OBJ_META_PRE_TEMPLATE.format(bucket=self._name, obj=base64.urlsafe_b64encode(bytes(obj, "utf-8")).decode())
```

nats/js/object_store.py
```diff
@@ -0,0 +1,328 @@
...
+            einfo = await self.get_info(meta.name)
+        except NotFoundError:
+            pass
+
+        # Create a random subject prefixed with the object stream name.
+        id = self._js._nc._nuid.next()
+        chunk_subj = OBJ_CHUNKS_PRE_TEMPLATE.format(
+            bucket=self._name, obj=id.decode()
+        )
+        meta_subj = OBJ_META_PRE_TEMPLATE.format(bucket=self._name, obj=obj)
```
tothandras, line 251:
```suggestion
        meta_subj = OBJ_META_PRE_TEMPLATE.format(bucket=self._name, obj=base64.urlsafe_b64encode(bytes(obj, "utf-8")).decode())
```

domderen:
Hey @tothandras thanks for your comments. I'm on holidays at the moment, but will look into it once I'm back in ~2 weeks.

tothandras:
@domderen enjoy your holiday!

nats/js/object_store.py
domderen: Hey, could you elaborate, why somthing like this?

domderen:
Hey @tothandras I updated this PR with your comments. I think I included them all, but would you mind taking a look?

nats/js/object_store.py
tothandras: I've seen both `"digest": "sha-256=xxx"` and `"digest": "SHA-256=xxx"`. Probably we should change this line instead: https://github.com/tothandras/nats.py/blob/object_store/nats/js/object_store.py#L50. The Go client uses `SHA`: https://github.com/nats-io/nats.go/blob/f4102422df7df83b2032318fd14a248c45f113e1/object.go#L232

tothandras:
@domderen If you could rebase on master I could try it out. I'm probably less suited for reviewing the code itself. I lack the deep knowledge of Python. :)

domderen:
Hey I think the rebase should be done, though I'm not sure if I didn't mess something up... Looks like a lot of changes now...

wallyqs:
@domderen I think you can try something like the following below to squash into a single commit and avoid the merge conflicts using your current branch:

```sh
# Checkout the object store branch
git checkout object_store

# Take a diff against latest main, we'll apply this diff as a patch.
git diff main > obj.diff

# Rename current local object_store branch.
git branch -m object-store-backup

# Redo object_store branch
git checkout main
git branch object_store
git checkout object_store

# Apply the diff as a patch on top of object_store branch and commit.
git apply -3 obj.diff

# Fix formatting
make format

git commit -s -m "Add Object Store implementation"

# Force push the object_store branch that is now squashed and rebased.
git push main object_store -f
```

I gave it a try at squashing into a single commit via a diff the result can be found here: https://github.com/nats-io/nats.py/compare/obj 
When running the tests I'm running into a ` binascii.Error: Incorrect padding` error though: 
https://app.travis-ci.com/github/nats-io/nats.py/jobs/585091380#L508

tothandras:
I just run into an issue with `__sanitize_name`, we should definitely write some integration tests to make sure all language implementations are compatible with each other. I hope to have time for it in the next 2-3 weeks.

domderen:
Hey @wallyqs thanks for the help! I followed your steps and force pushed just those changes that looked like my original implementation. It should make it easier to review. Let's see if the build passes now, if not I'll take a look at tests.

domderen:
@tothandras I think I tried to take this implementation of __sanitize_name from the golang implementation. Are they different now?

domderen:
Hmm there are two tests failing, but those are two different tests in two different builds, both don't seem to be related to my changes... Should I expect some flakiness in those tests?

tothandras:
> @tothandras I think I tried to take this implementation of __sanitize_name from the golang implementation. Are they different now?

Sorry, I was wrong. It seems I didn't apply all the patches on my fork.

danielfernandez-igz:
Hi everyone!

Are there any plans to resume this PR in the short term? I noticed it was reviewed six months ago, but has been stale for some time now.

Also, looks like the Travis build failed due to a flaky test but it's been excluded from `pytest` runs, so it shouldn't be a concern anymore (see https://github.com/nats-io/nats.py/pull/424).

Thank you all!

domderen:
Hey, I was updating this PR occasionally when I needed to change something in the Object Store implementation for my own needs. I'm happy to keep going on this PR, but I'm not sure if owners see it as valuable, or if they would prefer a different direction. 

It is also still missing some of the APIs that Object Store has in golang version, and I'm not sure what would be the minimal required scope for it to get merged.

PR #337, opened 2022-08-29
Add flush in Hello World example

HYAR7E:
Adding flush in the HelloWorld example would give it way more visibility than adding it to the API reference, and it will enforce the practice of executing a `flush` before closing NATS connection


wallyqs:
Excellent, thanks!

PR #341, opened 2022-08-31
Modernize the way we package and distribute nats.py by using PEP621

4383:
Hello,

Here are a couple of commits that I propose to modernize the way nats.py is packaged (by using PEP621).

For more details about the implementation please have a look to my given commit messages.

I Split my changes into 3 commits to properly isolates topics, however, if you prefer, I'd be ok to squash all my commits into a single one commit.


pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nats-py"
+version = "2.1.7"
```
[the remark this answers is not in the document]
charbonnierg: It should be the other way around, I.E, the `pyproject.toml` file can use dynamic metadata: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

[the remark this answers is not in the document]
4383: Done

4383:
> If you want developers to be able to perform editable install (PEP 660 - `pip install -e .`), then it is necessary to add a `setup.py` file as indicated [in the setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#:~:text=If%20compatibility%20with%20legacy%20builds%20or%20versions%20of%20tools%20that%20don%E2%80%99t%20support%20certain%20packaging%20standards%20(e.g.%20PEP%20517%20or%20PEP%20660)%2C%20a%20simple%20setup.py%20script%20can%20be%20added%20to%20your%20project%20%5B1%5D%20(while%20keeping%20the%20configuration%20in%20pyproject.toml)%3A):
> 
> > If compatibility with legacy builds or versions of tools that don’t support certain packaging standards (e.g. [PEP 517](https://peps.python.org/pep-0517/) or [PEP 660](https://peps.python.org/pep-0660/)), a simple setup.py script can be added to your project [[1]](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setupcfg-caveats) (while keeping the configuration in pyproject.toml):
> 
> ```python
> from setuptools import setup
> 
> setup()
> ```

Done

4383:
To check the generated package:
```
$ python -m build
$ tar -axf ./dist/nats-py-2.1.7.tar.gz nats-py-2.1.7/PKG-INFO -O
```

setup.py
```diff
@@ -1,31 +1,3 @@
...
-    url='https://github.com/nats-io/nats.py',
-    author='Waldemar Quevedo',
-    author_email='wally@synadia.com',
-    license='Apache 2 License',
-    packages=['nats', 'nats.aio', 'nats.protocol', 'nats.js'],
-    package_data={"nats": ["py.typed"]},
-    zip_safe=True,
-    extras_require=EXTRAS
-)
+setup()
```
wallyqs, line 3: Not very clear at the moment whether GitHub supports it yet for its dependency graph: https://github.com/community/community/discussions/6456

Really like the changes, but also wonder if we keep some of the basic metadata here still for a bit like Flask is doing:

https://github.com/pallets/flask/blob/main/setup.py

So maybe reduce it to something like this?

```python
# Metadata goes in pyproject.toml. These are here for GitHub's dependency graph.
setup(
    name="nats-py",
    extras_require={
        'nkeys': ['nkeys']
    }
)
```

4383: Github's dependency graph support `pyproject.toml`, please see https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems

However, if you prefer, it I'm not against managing the `extras_require` the way Flask is doing it.

4383: We should notice that Flask doesn't rely on `pyproject.toml` but rather on `setup.cfg`.

Concerning the dependency graph it seems to require a `poetry.lock` file, which is a related the poetry package manager. We don't rely on poetry here, so, if we want to keep this github feature available I'd also suggest to keep `extras_require` in the `setup.py` file.

I'll submit a new patch set to restore this point.

Thanks for the heads up.

4383: Done

4383: > Not very clear at the moment whether GitHub supports it yet for its dependency graph: [community/community#6456](https://github.com/orgs/community/discussions/6456)
> 
> Really like the changes, but also wonder if we keep some of the basic metadata here still for a bit like Flask is doing:
> 
> https://github.com/pallets/flask/blob/main/setup.py
> 
> So maybe reduce it to something like this?
> 
> ```python
> # Metadata goes in pyproject.toml. These are here for GitHub's dependency graph.
> setup(
>     name="nats-py",
>     extras_require={
>         'nkeys': ['nkeys']
>     }
> )
> ```

You should notice that currently those `extras_require` available in `setup.py` aren't listed in your current dependency graph https://github.com/nats-io/nats.py/network/dependencies.

Only those given by the pipenv artifacts are listed.

4383: Instead I'd suggest to use the extras mechanisms of `requirements.txt` defined by the PEP 508 [1] to manage those extras require.
`requirements.txt` is a well know mechanism where extras require can be defined.

It would lead us to:

```
nkeys [auth] >= 0.1.0
```

Instead of naming this extras `auth`, we could name it `security` by example, given the nature of the nkeys lib.
[1] https://peps.python.org/pep-0508/

4383: Github's dependency graph properly detect `the requirements.txt` file. Here is an example with one of my other project:
https://github.com/openuado/niet/network/dependencies

wallyqs (approved):
LGTM, adopting this starting from v2.2.0 release. Thanks!

PR #342, opened 2022-09-01
Makes connections examples more configurable

4383:
These changes propose to use argparse to allow users to set their servers for the given examples rather than modifying code base to adapt to our own needs.

Arguments for these examples are more centralized and, in a second time, could be more improved by also integrate `nats-[req|sub|pb]`.

These changes also rename existing options to names (from `servers` to `server`) for `nats-pub`, `nats-sub`, and `nats-req` as these examples can only consume one server at time.

### DIFF
--- a/examples/basic.py
+++ b/examples/basic.py
@@ -3,15 +3,12 @@ import asyncio
 import nats
 from nats.errors import ConnectionClosedError, NoServersError, TimeoutError
 
+from common import args
+
 
 async def main():
-    # It is very likely that the demo server will see traffic from clients other than yours.
-    # To avoid this, start your own locally and modify the example to use it.
-    nc = await nats.connect("nats://demo.nats.io:4222")
-
-    # You can also use the following for TLS against the demo server.
-    # 
-    # nc = await nats.connect("tls://demo.nats.io:4443")
+    arguments, _ = args.get_args("Run a basic example.")
+    nc = await nats.connect(arguments.servers)
 
     async def message_handler(msg):
         subject = msg.subject
--- a/examples/client.py
+++ b/examples/client.py
@@ -4,6 +4,8 @@ from datetime import datetime
 from nats.aio.client import Client as NATS
 from nats.errors import ConnectionClosedError, TimeoutError
 
+from common import args
+
 
 class Client:
     def __init__(self, nc):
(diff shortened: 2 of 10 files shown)


examples/nats-pub/__main__.py
```diff
@@ -43,7 +43,7 @@ async def run():
     # e.g. nats-pub -s demo.nats.io hello "world"
     parser.add_argument('subject', default='hello', nargs='?')
     parser.add_argument('-d', '--data', default="hello world")
-    parser.add_argument('-s', '--servers', default="")
+    parser.add_argument('-s', '--server', default="")
```
4383, line 46: This example can consume only one server at time.

examples/nats-req/__main__.py
```diff
@@ -43,7 +43,7 @@ async def run():
     # e.g. nats-req -s demo.nats.io hello world
     parser.add_argument('subject', default='hello', nargs='?')
     parser.add_argument('-d', '--data', default="hello world")
-    parser.add_argument('-s', '--servers', default="")
+    parser.add_argument('-s', '--server', default="")
```
4383, line 46: This example can consume only one server at time.

examples/nats-sub/__main__.py
```diff
@@ -42,7 +42,7 @@ async def run():
 
     # e.g. nats-sub hello -s nats://127.0.0.1:4222
     parser.add_argument('subject', default='hello', nargs='?')
-    parser.add_argument('-s', '--servers', default="")
+    parser.add_argument('-s', '--server', default="")
```
4383, line 45: This example can consume only one server at time.

wallyqs (approved):
LGTM

4383:
> LGTM

Please hold it, I was wrong when I said that `nats[pub|sub|req]` only consume one server at once. I'll submit soon a new patch set.

See the example in the doc: https://docs.nats.io/running-a-nats-service/configuration/clustering

4383:
> > LGTM
> 
> Please hold it, I was wrong when I said that `nats[pub|sub|req]` only consume one server at once. I'll submit soon a new patch set.
> 
> See the example in the doc: https://docs.nats.io/running-a-nats-service/configuration/clustering

Done. The latest patch set drops the `examples/nats-[sub|pub|req]/__main__.py` files as they can't rely on an external module as added by these changes. Else the doc examples won't work anymore.

https://docs.nats.io/running-a-nats-service/configuration/clustering

examples/nats-pub/__main__.py
4383: You are right, the client can handle more than one server, so I decided to revert this part of my changes.
Also changing this name would require changing the doc outside the scope of this repository, so I prefer to leave it as it is now.
https://docs.nats.io/running-a-nats-service/nats_docker/ngs-docker-python

4383:
This pull request is ready to go. I applied my latest to comments above into the latest submitted patch set.

Thanks for your time.

PR #343, opened 2022-09-01
Support sending requests with reconnected client

charbonnierg:
This pull request fixes a bug (or at least a non documented behaviour). It brings very little change to the codebase.

## Problem description

- [x] It is not possible to send requests using a client which has been connected twice.

## Reproducible example:

```python
from nats import NATS

nc = NATS()

await nc.connect()
await nc.close()
await nc.connect()
# We expect a NoRespondersError but we get a TimeoutError without this PR
await nc.request("some.unknown.subject", timeout=0.1)
```

## Objectives of the PR

- [x] Let users send requests with clients which were already closed before

## Proposal implementation

- [x] Reset `Client._resp_sub_prefix` at the end of the `Client._close` method.
  
I saw that `Client._client_id` is already resetted at that time, so I though that it was the right place to reset `Client._resp_sub_prefix` too.

## How was this PR tested ?

I added a single test which is very similar to the example above. If we can talk to NATS server, then a request on a subject without subjecription must raise a `NoRespondersError`, if not, a `TimeoutError` is raised and test fails.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -638,8 +638,9 @@ class Client:
             if self._closed_cb is not None:
                 await self._closed_cb()
 
-        # Set the client_id back to None
+        # Set the client_id and subscription prefix back to None
         self._client_id = None
+        self._resp_sub_prefix = None
 
     async def drain(self) -> None:
         """
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -918,6 +918,19 @@ class ClientTest(SingleServerTestCase):
         self.assertEqual(0, err_count)
 
     @async_test
+    async def test_connect_after_close(self):
+        nc = await nats.connect()
+        with self.assertRaises(nats.errors.NoRespondersError):
+            await nc.request("missing", timeout=0.01)
+        await nc.close()
+        await nc.connect()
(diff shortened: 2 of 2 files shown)


charbonnierg:
When I run `yapf` on my branch, I get a lot of modifications:

```bash
        modified:   nats/aio/subscription.py
        modified:   nats/errors.py
        modified:   nats/js/api.py
        modified:   nats/js/client.py
        modified:   nats/js/errors.py
        modified:   nats/js/kv.py
        modified:   nats/js/manager.py
        modified:   nats/nuid.py
        modified:   nats/protocol/parser.py
        modified:   tests/test_client.py
        modified:   tests/test_client_async_await.py
        modified:   tests/test_client_nkeys.py
        modified:   tests/test_client_v2.py
        modified:   tests/test_js.py
        modified:   tests/test_nuid.py
        modified:   tests/test_parser.py
        modified:   tests/utils.py
```
so I did not run `yapf` before the commit

wallyqs (approved):
LGTM, some clients (like Go) do not expose `connect` on the instance type to avoid reusing, but in the case of Python this approach would be more convenient since we do allow creating a NATS client instance to connect besides the `nats.connect` method from the module.

PR #356, opened 2022-09-20
Feature/implement websockets

### LINKED ISSUE #270: [question] Will nats.py support websockets
As the subject says. Interested in knowing if there are any intentions to enable use of the websockets port from the nats.py client.

### LINKED ISSUE #353: how to connect with websockets?
is there a way to connect with websockets?

it is possible to write my own code to connect to the nats? What I should consider?
Is there a way this libary will support nats.ws?

bvanelli:
Open points:

- [x] Implement and test websockets transport for nats.
   - [x] Test old behaviour
   - [x] Test ws
   - [x] Test wss
- [x] Need to review the changes made to the core of nats
   - [x] Created a ssl_context implementation as property of client https://github.com/nats-io/nats.py/blob/c72de758b7c053292b8cd005ba3fd5a5fb5f8e7c/nats/aio/client.py#L1169
   - [x] Added an increase in _pending_data_size. To me this seems like a bug, where the flush would be skipped if the transport used previously (stream) didn't write bytes immediately https://github.com/nats-io/nats.py/blob/c72de758b7c053292b8cd005ba3fd5a5fb5f8e7c/nats/aio/client.py#L1958
   - [x] Added transport instance inside client. https://github.com/nats-io/nats.py/blob/c72de758b7c053292b8cd005ba3fd5a5fb5f8e7c/nats/aio/client.py#L1278

- [x] Need to figure out what to do when transport needs replacement (i.e. when mixing both ws and nats connections on the server list)
- [x] Make tests run
- [x] Optional: include aiohttp as a optional dependency, similar to nkeys
- [ ] Optional: write more tests to test the transport

Here is the test client to test this functionality:

```python
import asyncio
import nats


async def main():
    nc = await nats.connect("wss://demo.nats.io:8443")

    async def message_handler(msg):
        subject = msg.subject
        reply = msg.reply
        data = msg.data.decode()
        print("Received a message on '{subject} {reply}': {data}".format(
            subject=subject, reply=reply, data=data))

    # Simple publisher and async subscriber via coroutine.
    sub = await nc.subscribe("foo", cb=message_handler)

    await nc.publish("foo", b'Hello')
    await nc.publish("foo", b'World')
    await nc.publish("foo", b'!!!!!')

    await asyncio.sleep(5)  # wait for messages to arrive

    await sub.unsubscribe()

    # Terminate connection to NATS.
    await nc.drain()
    await nc.close()


if __name__ == '__main__':
    asyncio.run(main())

```

Closes #353
Closes #270

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1206,6 +1206,8 @@ class Client:
                     # Closer to how the Go client handles this.
                     # e.g. nats://localhost:4222
                     uri = urlparse(connect_url)
+                elif "ws://" in connect_url or "wss://" in connect_url:
+                    uri = urlparse(connect_url)
                 elif ":" in connect_url:
                     # Expand the scheme for the user
                     # e.g. localhost:4222
@@ -1263,14 +1265,13 @@ class Client:
             try:
                 s.last_attempt = time.monotonic()
                 if not self._transport:
-                    if s.uri.hostname.startswith("ws"):
+                    if s.uri.scheme == "ws":
                         self._transport = WebsocketTransport()
                     else:
                         # use TcpTransport as a fallback
                         self._transport = TcpTransport()
                 await self._transport.connect(
-                    s.uri.hostname,
-                    s.uri.port,
+                    s.uri,
                     buffer_size=DEFAULT_BUFFER_SIZE,
                     connect_timeout=self.options['connect_timeout']
                 )
@@ -1869,8 +1870,8 @@ class Client:
 
             # connect to transport via tls
             await self._transport.connect_tls(
+                self._current_server.uri,
                 ssl_context,
-                hostname,
                 DEFAULT_BUFFER_SIZE,
                 self.options['connect_timeout']
             )
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1245,6 +1245,14 @@ class Client:
                     self._server_pool.append(Srv(uri))
             except ValueError:
                 raise errors.Error("nats: invalid connect url option")
+            # make sure protocols aren't mixed
+            if not (all(server.uri.scheme in ("nats", "tls")
+                        for server in self._server_pool)
+                    or all(server.uri.scheme in ("ws", "wss")
+                           for server in self._server_pool)):
+                raise errors.Error(
+                    "nats: mixing of websocket and non websocket URLs is not allowed"
+                )
         else:
             raise errors.Error("nats: invalid connect url option")
 
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -56,7 +56,7 @@ from .subscription import (
     DEFAULT_SUB_PENDING_MSGS_LIMIT,
     Subscription,
 )
-from .transport import Transport, TcpTransport, WebsocketTransport
+from .transport import Transport, TcpTransport, WebSocketTransport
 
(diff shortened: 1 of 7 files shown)


nats/aio/transport.py
```diff
@@ -0,0 +1,231 @@
...
+        return self._io_writer.close()
+
+    def at_eof(self):
+        return self._io_reader.at_eof()
+
+    def __bool__(self):
+        return bool(self._io_writer) and bool(self._io_reader)
+
+
+class WebsocketTransport(Transport):
```
wallyqs, line 168: how about `WebSocket` instead of `Websocket` to follow same style as aiohttp:
```suggestion
class WebSocketTransport(Transport):
```

wallyqs:
Just gave it a try, awesome work!
```
python3 examples/basic.py -s wss://demo.nats.io:8443
Received a message on 'foo ': Hello
Received a message on 'foo ': World
Received a message on 'bar ': First
Received a message on 'bar ': Second
Received a message on 'help _INBOX.GQhnv0uOPDvMfLg1OziGsD.GQhnv0uOPDvMfLg1OziGxXe814': help me
```

wallyqs:
it looks like needs to add this to .travis.yml for the tests to run in CI:

```
install:
  - pip install -e .[nkeys]
  - pip install aiohttp
```

nats/aio/client.py
```diff
@@ -1973,11 +1948,12 @@ async def _process_connect_init(self) -> None:
...
 
     async def _send_ping(self, future: asyncio.Future = None) -> None:
-        assert self._io_writer, "Client.connect must be called first"
+        assert self._transport, "Client.connect must be called first"
         if future is None:
             future = asyncio.Future()
         self._pongs.append(future)
-        self._io_writer.write(PING_PROTO)
+        self._transport.write(PING_PROTO)
+        self._pending_data_size += len(PING_PROTO)
```
wallyqs, line 1956: yes looks like this was a bug...

bvanelli:
One of the jobs failed https://app.travis-ci.com/github/nats-io/nats.py/jobs/583496905 because of https://github.com/google/yapf/issues/936. I included a patch for the CI as a workaround to have tests passing. Just need to figure out when transport needs replacement and it looks in a good state to me.

bvanelli:
I wanted to do something like:

```python

**Comment by wallyqs** [comment]
Looking at the Go client, when passing the options on `connect`, it should be checked whether it is being attempted to mix websockets and nats schemes and thrown an error:
```
nats: error: mixing of websocket and non websocket URLs is not allowed
```

bvanelli:
> Looking at the Go client, when passing the options on `connect`, it should be checked whether it is being attempted to mix websockets and nats schemes and thrown an error:
> 
> ```
> nats: error: mixing of websocket and non websocket URLs is not allowed
> ```
> 
> https://github.com/nats-io/nats.go/blob/715a5917c806d2858f6fc80663fbf0910ad58ee3/nats.go#L1589-L1597

Sounds fair, I added a check for that on connect: https://github.com/nats-io/nats.py/pull/356/commits/7ac8e8e18024550aa25ba31eb3f2103ec93b1ab7

wallyqs (approved):
LGTM! Thanks for this excellent contribution! 🎉

Issue #360, opened 2022-10-04
Unexpected port when connecting through secure websocket (wss)


bvanelli:
There are some small issues left when connecting with a secure websocket (wss) without specifying a port, as it gets treated by the following else statement:

https://github.com/nats-io/nats.py/blob/main/nats/aio/client.py#L1234

That then changes the protocol unexpectedly. The correct behavior should be to instead use port 443 as default, assuming websocket is being served in the default tls port.

Also, this replacements do not support proxies (i.e. traefik) as the path name would be omitted.

I'll send a mr for this one 😄

PR #371, opened 2022-10-08
Fix issue where protocol was changed for ws and wss if no port was provided.

bvanelli:
Fixes https://github.com/nats-io/nats.py/issues/360


wallyqs (approved):
LGTM

Issue #361, opened 2022-10-04
Untreated error callbacks when using websockets


bvanelli:
Turns out some errors are not handled and show full traceback when using websockets. For example this one, when I restart the nats server:

```python
ERROR:nats.aio.client:nats: encountered error
Traceback (most recent call last):
  File "C:\Users\brunno.vanelli\PycharmProjects\base-python\venv\lib\site-packages\nats\aio\[client.py](http://client.py/)", line 2035, in _read_loop
    await self._ps.parse(b)
  File "C:\Users\brunno.vanelli\PycharmProjects\base-python\venv\lib\site-packages\nats\protocol\[parser.py](http://parser.py/)", line 93, in parse
    self.buf.extend(data)
TypeError: can't extend bytearray with int
``` 

Or this one, that I could not reproduce but I guess is related to reconnect:

```
ERROR:nats.aio.client:nats: encountered error
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x0000022EB5AE54C0>
ERROR:asyncio:Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x0000022EB34A3130>
```

If anyone finds more of those please append to the issue.

I'll submit a merge request once I do more testing and find more edge cases.

Rajiv91:
Hey @bvanelli i'm seing the same error with the websocket that you reported:

`
TypeError: can't extend bytearray with int
`
But this doesn't trigger the error nor the disconnection cb, it just stops the loop that parses the messages from the server.
The code says that if falls in some of those parse exceptions : it will stop running and its task has to be rescheduled.
But i'm not seing such a thing happening to reschedule within client.py, i'm talking specifically about the task:

`
        self._reading_task = asyncio.get_running_loop().create_task(
            self._read_loop()
        )
`
So at the end when this happens my nats connection hangs and i have to reboot my python process to stablish it again and send the messages successfully. Do you know if this is a known issue that someone else is working on? or when they say that "its task has to be rescheduled" they mean  the user should mess with the client.py code to reschedule the task in some place, because I can't find where they do it.

bvanelli:
@Rajiv91 Which version are you using? Do you have some short code describing the issue?

Also, it is a known issue that if you have an error callback that raises an exception, it could cause problems you described.

Rajiv91:
Hi  @bvanelli thanks for your reply.
[...]

PR #375, opened 2022-10-14
Untreated error callbacks when using websockets

bvanelli:
There are two things:

- The parsing issue was caused by readline returning the actual error code (in this case 1001). I made it return `b''` instead and EOF will be triggered in the next call
- `__bool__` was returning false every time it was disconnected. I'm returning now if `self._client` exists instead.

If you want to test, simply use the script

```python
import asyncio
import nats


async def main():
    async def cb(msg):
        print(msg)
    ws = await nats.connect("ws://localhost", error_cb=cb)
    await ws.subscribe("hello", cb=cb)
    for i in range(100):
        await ws.publish("hello", f"hello world {i}".encode())
        await asyncio.sleep(5)


if __name__ == '__main__':
    asyncio.run(main())
```

And turn off the connection to NATS. This will cause now EOF in both implementations.

Closes https://github.com/nats-io/nats.py/issues/361

### DIFF
--- a/nats/aio/transport.py
+++ b/nats/aio/transport.py
@@ -250,6 +250,9 @@ class WebSocketTransport(Transport):
 
     async def readline(self):
         data = await self._ws.receive()
+        if data.type == aiohttp.WSMsgType.CLOSE:
+            # if the connection terminated abruptly, return empty binary data to raise unexpected EOF
+            return b''
         return data.data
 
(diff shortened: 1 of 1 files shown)


wallyqs (approved):

bvanelli:
@wallyqs do you have further points to address?

Issue #373, opened 2022-10-13
CancelledError suppression issue


chris62837:
Hi, we are observing a problem with cancelling tasks when using nats-py
 
As an example:
```
import asyncio
import nats
import nats.errors
 
def main():
    asyncio.run(foo())
 
async def foo():
 
    connection = nats.NATS()
 
    await connection.connect(...)
 
    subscription = await connection.jetstream().pull_subscription_bind(...)
 
    while True:
        try:
            msgs = await subscription.fetch(batch=1, timeout=5)
            msg = msgs[0]
        except nats.errors.TimeoutError:
            continue
 
        # process msg
 
        await msg.ack()
```
 
Upon a `KeyboardInterrupt`, or `SIGINT` (e.g. on Ctrl-c or when halting a docker container of our application), we expect the behaviour to be a `asyncio.CancelledError` raised by `subscription.fetch`, however, the `CancelledError` exception is being caught at [https://github.com/nats-io/nats.py/blob/main/nats/aio/subscription.py#L174](https://github.com/nats-io/nats.py/blob/main/nats/aio/subscription.py#L174) and replaced with a `TimeoutError`. Which will, in the code above, result in the error being not handled correctly. (fetch is just retried).
 
We have implemented the following workaround, but think this would ideally not be necessary.
```
async def bar():
 
    connection = nats.NATS()
 
    await connection.connect(...)
 
    subscription = await connection.jetstream().pull_subscription_bind(...)
 
    async def do_fetch():
        return await subscription.fetch(batch=1, timeout=5)
 
    while True:
        try:
            try:
                task = asyncio.create_task(do_fetch())
                msgs = await asyncio.wait_for(asyncio.shield(task))
                msg = msgs[0]
            except asyncio.CancelledError:
                task.cancel()
                try:
                    await asyncio.wait_for(task)
                except nats.errors.TimeoutError:
                    pass
                raise
        except nats.errors.TimeoutError:
            continue
 
        # process msg
 
        await msg.ack()
 ```
 
We are proposing changing [https://github.com/nats-io/nats.py/blob/main/nats/aio/subscription.py#L178](https://github.com/nats-io/nats.py/blob/main/nats/aio/subscription.py#L178)

From: raise errors.TimeoutError
To: raise
 
This is the behaviour recommended by the asyncio python docs: [https://docs.python.org/3/library/asyncio-exceptions.html](https://docs.python.org/3/library/asyncio-exceptions.html)
 
[...]

PR #378, opened 2022-10-26
Raise asyncio.CancelledError back

charbonnierg:
## Motivation

> This PR addresses the issue #373

When a SIGINT is received by an python application running a pull subscribe consumer, `next_msg()` coroutine may suppress the exception.

This PR raises back the exception once it's catched.

## State of the PR

- [x] This PR fixes the issue mentionned by #373 

- [ ] This PR includes comments which should not be kept. I did a little `Ctrl+F` with `except asyncio.CancelledError` and saw several except blocks. I wonder if they could also cause suppression of the exception.

- [ ] The message commit of this PR should be modified (current message is "wip: add comments regarding cancel exceptions")

I hope it can be useful to discuss what changes are needed to avoid suppressions exceptions :)

### DIFF
--- a/nats/aio/subscription.py
+++ b/nats/aio/subscription.py
@@ -174,8 +174,7 @@ class Subscription:
         except asyncio.CancelledError:
             future.cancel()
             task.cancel()
-            # Call timeout otherwise would get an empty message.
-            raise errors.TimeoutError
+            raise
 
     def _start(self, error_cb):
         """
@@ -230,6 +229,7 @@ class Subscription:
             # Subscription is done and won't be receiving further
             # messages so can throw it away now.
             self._conn._remove_sub(self._id)
+        # QUESTION: Can this except block swallow external cancellations ?
         except asyncio.CancelledError:
             # In case draining of a connection times out then
             # the sub per task will be canceled as well.
@@ -288,6 +288,7 @@ class Subscription:
                 try:
                     # Invoke depending of type of handler.
                     await self._cb(msg)
+                # QUESTION: Can this except block swallow external cancellations ?
                 except asyncio.CancelledError:
                     # In case the coroutine handler gets cancelled
                     # then stop task loop and return.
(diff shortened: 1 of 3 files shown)


nats/aio/subscription.py
```diff
@@ -288,6 +288,7 @@ async def _wait_for_msgs(self, error_cb) -> None:
                 try:
                     # Invoke depending of type of handler.
                     await self._cb(msg)
+                # QUESTION: Can this except block swallow external cancellations ?
                 except asyncio.CancelledError:
```
charbonnierg, line 292: If user has a long running cb, then cancellation might be suppressed here too

chris62837:
Thanks for opening this, I was also on the cusp of doing the same (minus you open questions for the potential other suppression points!)

I had written a test which verifies that the CancelledError is now raised as expected from this function with some mocking, may or may not be useful:

```
    @async_test
    async def test_fetch_cancelled_errors_raised(self):
        nc = NATS()
        await nc.connect()

        js = nc.jetstream()

        await js.add_stream(name="test", subjects=["test.a"])
        await js.add_consumer(
            "test",
            durable_name="a",
            deliver_policy=nats.js.api.DeliverPolicy.ALL,
            max_deliver=20,
            max_waiting=512,
            max_ack_pending=1024,
            filter_subject="test.a"
        )

        sub = await js.pull_subscribe_bind("a", stream="test")

        with self.assertRaises(asyncio.CancelledError):
            with unittest.mock.patch("asyncio.wait_for", unittest.mock.AsyncMock(side_effect=asyncio.CancelledError)):
                await sub.fetch(batch=1, timeout=0.1)
```

charbonnierg:
Thanks @chris62837 , I added the unit test to the PR

willish:
I wonder if `nats/js/client.py` needs some changes to handle the new possibility of `asyncio.CancelledError` being raised during the invocation of `next_msg()`

NeutrinoXY:
Any update on this? This is an important change for proper exception handling. @wallyqs

PR #384, opened 2022-11-18
Fix all new mypy errors

orsinium:
I've been away for 10 months, and you've already disabled mypy 🙃 

The PR fixes all new mypy errors, to my best effort. In follow-up PRs I'm going to get mypy back on CI and upgrade it to the latest version.

Please, notice what I did with `TcpTransport.connect_tls`. Mypy tells me this code doesn't work, and I agree. Well, I've thrown an assertion for now that I'm pretty sure won't work, but it silents mypy for now. If I understand correctly, this code is not used, not tested, and is just a draft. But I might be wrong.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -127,7 +127,7 @@ class ServerVersion:
         self._patch_version: Optional[int] = None
         self._dev_version: Optional[str] = None
 
-    # TODO(@orisinum): use cached_property
+    # TODO(@orsinium): use cached_property
     def parse_version(self) -> None:
         v = (self._server_version).split('-')
         if len(v) > 1:
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1954,7 +1954,9 @@ class Client:
             self._flusher()
         )
 
-    async def _send_ping(self, future: Optional[asyncio.Future] = None) -> None:
+    async def _send_ping(
+        self, future: Optional[asyncio.Future] = None
+    ) -> None:
         assert self._transport, "Client.connect must be called first"
         if future is None:
             future = asyncio.Future()
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -18,7 +18,6 @@ import ipaddress
 import json
 import logging
 import ssl
-import sys
 import time
 from secrets import token_hex
 from dataclasses import dataclass
@@ -121,14 +120,15 @@ class Srv:
 
 class ServerVersion:
 
-    def __init__(self, server_version: str):
+    def __init__(self, server_version: str) -> None:
         self._server_version = server_version
-        self._major_version = None
-        self._minor_version = None
-        self._patch_version = None
-        self._dev_version = None
+        self._major_version: Optional[int] = None
+        self._minor_version: Optional[int] = None
+        self._patch_version: Optional[int] = None
+        self._dev_version: Optional[str] = None
 
-    def parse_version(self):
+    # TODO(@orisinum): use cached_property
+    def parse_version(self) -> None:
         v = (self._server_version).split('-')
         if len(v) > 1:
             self._dev_version = v[1]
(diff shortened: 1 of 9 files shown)


orsinium:
Did I break a test? 👀 I'll have a look on Sunday.

nats/aio/transport.py
```diff
@@ -139,6 +139,13 @@ async def connect_tls(
         buffer_size: int,
         connect_timeout: int,
     ):
+        # TODO(@orsinium): this assertion is wrong but so the implementation,
+        # and IDK how to fix it.
+        assert self._io_writer
```
wallyqs, line 144: what is the error reported by mypy? This would have been used here: https://github.com/nats-io/nats.py/blob/48059d5191fc8626e6abc85509c54413319c3585/nats/aio/client.py#L1893-L1898

orsinium: `self._io_writer` is defined as `Optional[StreamWriter]` and is defined in 3 places:

1. In `__init__` as None
2. In `connect` using `asyncio.open_connection`.
3. In `connect_tls` at line 169 but AFTER it is used at line 156.

So, either line 156 fails with AttributeError or `self._io_writer` is defined somewhere else. Is it defined? Is `connect` supposed to be called before `connect_tls`?

orsinium:
All tests pass locally for me (except websocket ones, but I guess my nats-server isn't new enough). Is CI flaky?

nats/aio/transport.py
wallyqs: I think so, first `connect` would happen in `select_next_server` which would establish the TCP connection, then after getting the initial `INFO` it would upgrade with `connect_tls`:
https://github.com/nats-io/nats.py/blob/48059d5191fc8626e6abc85509c54413319c3585/nats/aio/client.py#L1302-L1308

wallyqs:
yes, some of the `PullSubscribeTest` looks like are flaky...

nats/aio/transport.py
orsinium: Thank you! Fixed: https://github.com/nats-io/nats.py/pull/384/commits/a826001f5a2e13ba34ffc1e158979d232370a1d2

nats/aio/client.py
```diff
@@ -121,14 +120,15 @@ class Srv:
...
-        self._minor_version = None
-        self._patch_version = None
-        self._dev_version = None
+        self._major_version: Optional[int] = None
+        self._minor_version: Optional[int] = None
+        self._patch_version: Optional[int] = None
+        self._dev_version: Optional[str] = None
 
-    def parse_version(self):
+    # TODO(@orisinum): use cached_property
```
wallyqs, line 130: typo on handle 😉 
```suggestion
    # TODO(@orsinium ): use cached_property
```

nats/aio/transport.py
```diff
@@ -139,6 +139,11 @@ async def connect_tls(
         buffer_size: int,
         connect_timeout: int,
     ):
+        assert self._io_writer, f'{type(self).__name__}.connect must be called first'
+
+        # TODO(@orsinium): nats.py doesn't support Python below 3.7,
+        # we don't need this check.
```
wallyqs, line 145: yes think we can remove the check at this point.

wallyqs (approved):
LGTM! will merge after format fixes shown on CI, but tests and mypy looks good to me locally 👍 ✅  
```
pipenv run mypy
Success: no issues found in 21 source files
```
Thanks for the contribution again!

nats/aio/transport.py
```diff
@@ -140,44 +137,25 @@ async def connect_tls(
...
+        protocol = asyncio.StreamReaderProtocol(reader)
+        transport_future = asyncio.get_running_loop().start_tls(
+            self._io_writer.transport,
+            protocol,
+            ssl_context,
+            # hostname here will be passed directly as string
+            server_hostname=uri if isinstance(uri, str) else uri.hostname
+        )
+        transport = await asyncio.wait_for(transport_future, connect_timeout)
+        assert isinstance(transport, asyncio.WriteTransport)
```
wallyqs, line 154: This fails because apparently for uvloop, `uvloop.loop._SSLProtocolTransport` is not an instance of `asyncio.WriteTransport`. I will remove these asserts before merging for now, tests passing locally otherwise.

wallyqs: oh I see but then with mypy start to get the following...
```
nats/aio/transport.py:170: error: Item "None" of "Optional[StreamReader]" has no attribute "read"  [union-attr]
```

nats/js/client.py
```diff
@@ -291,7 +290,7 @@ async def cb(msg):
                 config.ack_wait = 22 * 3600  # 22 hours
                 config.idle_heartbeat = idle_heartbeat
                 config.num_replicas = 1
-                config.memory_storage = True
+                config.mem_storage = True
```
wallyqs, line 293: this bug would have been caught by mypy...
```
nats/js/client.py:295: error: "ConsumerConfig" has no attribute "memory_storage"; maybe "mem_storage"?  [attr-defined]
```

wallyqs:
Ok merged manually including this to fix the uvloop issue: https://github.com/nats-io/nats.py/commit/c99fdf16de2a9dc712ba36f21f8b4baf788ba2f4
Thanks again!

PR #385, opened 2022-11-19
Sort imports

orsinium:
Sort imports using [isort](https://github.com/PyCQA/isort)


wallyqs (approved):
LGTM! I had not heard about isort, will give it a try from now.

PR #386, opened 2022-11-19
Make all type annotations lazy

orsinium:
[PEP 563](https://peps.python.org/pep-0563/) introduced postponed evaluation of type annotations. It makes importing modules faster and allows to use the latest Python features in older Python releases and to use forward references and type-checking-only objects without quoting annotations. In other words, a lot of cool stuff. The feature can be activated since Python 3.7 by adding in the file `from __future__ import annotations`. The feature was supposed to be the default behavior since Python 3.10 or so, but was postponed indefinitely because of some corner cases when it doesn't work well (runtime type annotation evaluation for closures).

### DIFF
--- a/nats/__init__.py
+++ b/nats/__init__.py
@@ -11,6 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from __future__ import annotations
+
 from typing import List, Union
 
 from .aio.client import Client as NATS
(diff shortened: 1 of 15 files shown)


orsinium:
CI is just being a drama queen again. `test_fetch_n` needs some love.

wallyqs (approved):
LGTM

PR #388, opened 2022-11-22
Use the new type annotations syntax

orsinium:
PR #386 made all [type annotations lazy](https://peps.python.org/pep-0563/), meaning that they are not evaluated in runtime. That means, now we can use the [new syntax for type annotations](https://peps.python.org/pep-0604/) without breaking compatibility with older Python versions.

Before:

```python
from typing import List, Optional

a: Optional[List[str]]
```

After:

```python
a: list[str] | None
```

### DIFF
--- a/nats/__init__.py
+++ b/nats/__init__.py
@@ -13,14 +13,11 @@
 #
 from __future__ import annotations
 
-from typing import List, Union
-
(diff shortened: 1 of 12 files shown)


wallyqs (approved):
LGTM! 👍

PR #389, opened 2022-11-23
Make Transport.connect a classmethod

orsinium:
We have a number of places where we expect methods to be called in a specific order. It mostly concerns `connect` methods that are expected to be called as soon as the class is instantiated. You can find places like this by looking for `assert self._something` statements at the beginning of methods.

One such place is `Transport` which doesn't work if `Transport.connect` is not called first. The PR converts `Transport.connect` into a classmethod. So, now it's a constructor that returns an instantiated and connected `Transport`. It guarantees that any `Transport` instance is always connected. Well, or disconnected, but at least we can guarantee that `TcpTransport._io_reader` is never `None`.


orsinium:
CI is a drama queen.

caspervonb:
Stale, although valid going to close this out. nats-core takes this approach.

Issue #392, opened 2022-11-23
js pull consumer lost messages


ppd0705:
I am using jetstream pull subscription to fetch messages, I found some messages lost. I think the [code](https://github.com/nats-io/nats.py/blob/main/nats/aio/subscription.py#L161) below may case the problem.
```python
        future: asyncio.Future[Msg] = asyncio.Future()

        async def _next_msg() -> None:
            msg = await self._pending_queue.get()
            self._pending_size -= len(msg.data)
            future.set_result(msg)

        task = asyncio.get_running_loop().create_task(_next_msg())
        try:
            msg = await asyncio.wait_for(future, timeout)
            return msg
        except asyncio.TimeoutError:
            future.cancel()
            task.cancel()
            raise errors.TimeoutError
        except asyncio.CancelledError:
            future.cancel()
            task.cancel()
            # Call timeout otherwise would get an empty message.
            raise errors.TimeoutError
```

when `wait_for` timeout, it would cancel the future, but cancel future will use the [function](https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L544)  below
```python 
async def _cancel_and_wait(fut, loop):
    """Cancel the *fut* future or task and wait until it completes."""

    waiter = loop.create_future()
    cb = functools.partial(_release_waiter, waiter)
    fut.add_done_callback(cb)

    try:
        fut.cancel()
        # We cannot wait on *fut* directly to make
        # sure _cancel_and_wait itself is reliably cancellable.
        await waiter
    finally:
        fut.remove_done_callback(cb)
```

after `fut.cancel()`, there is a coroutine `await waiter` called, then event loop may transfer to `_next_msg` task.
In `_next_msg`,  a msg may got, but the msg can't be result of the future, because the future canceled already.

PR #393, opened 2022-11-23
js: fix msg losing

ppd0705:
as [issue 392](https://github.com/nats-io/nats.py/issues/392) said, js pull consumer may lose messages


ppd0705:
@wallyqs  could you have a look? thank you.

wallyqs:
Fixed via #446

PR #394, opened 2022-11-24
Move configs from setup.cfg into pyproject.toml

orsinium:
Cool kids use pyproject.toml for tool configs. TOML is nice and typed, with syntax highlighting, linters, formatters, all that stuff. Only flake8 stays in setup.cfg because they're boring.


wallyqs (approved):
LGTM

PR #396, opened 2022-11-25
mypy: check untyped defs

orsinium:
I've set `check_untyped_defs = true` for mypy and fixed whatever it reported. I didn't commit the updated config, though, because otherwise, it throws 14 more errors that I've fixed in #389

### DIFF
--- a/nats/aio/subscription.py
+++ b/nats/aio/subscription.py
@@ -84,9 +84,9 @@ class Subscription:
         self._pending_queue: asyncio.Queue[Msg] = asyncio.Queue(
             maxsize=pending_msgs_limit
         )
-        self._pending_size = 0
-        self._wait_for_msgs_task = None
-        self._message_iterator = None
+        self._pending_size: int = 0
+        self._wait_for_msgs_task: asyncio.Task | None = None
+        self._message_iterator: _SubscriptionMessageIterator | None = None
 
         # For JetStream enabled subscriptions.
         self._jsi: JetStreamContext._JSI | None = None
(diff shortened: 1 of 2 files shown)


nats/aio/subscription.py
```diff
@@ -302,7 +304,8 @@ async def _wait_for_msgs(self, error_cb) -> None:
                     self._pending_queue.task_done()
 
                 # Apply auto unsubscribe checks after having processed last msg.
-                if self._max_msgs > 0 and self._received >= self._max_msgs and self._pending_queue.empty:
+                if self._max_msgs > 0 and self._received >= self._max_msgs and self._pending_queue.empty(
+                ):
```
orsinium, line 308: I know it's not the right place to discuss this, but yapf is a very bad autoformatter. What the hell is this? How on the Earth this is readable?

wallyqs: sorry for the late reply, I'm ok with removing yapf if you want or use a replacement

orsinium: I was just complaining out loud, I didn't expect a reply. But I'm glad you are open to alternatives. I maintain the [awesome-python-code-formatters](https://github.com/life4/awesome-python-code-formatters) list, so I'd love to bring some good alternative code formatters into the project :) I'll do that when we get Taskfile merged, so I can provide you with a good automation for that.

nats/aio/subscription.py
```diff
@@ -84,9 +84,9 @@ def __init__(
         self._pending_queue: asyncio.Queue[Msg] = asyncio.Queue(
             maxsize=pending_msgs_limit
         )
-        self._pending_size = 0
-        self._wait_for_msgs_task = None
-        self._message_iterator = None
+        self._pending_size: int = 0
+        self._wait_for_msgs_task: asyncio.Task | None = None
```
Lancetnik, line 88: Annotation with `|` are supported since Py3.8
Cuz the library still supports 3.7 for any time, this code should use Optional, Union, etc
Please, setup your formatter to check 3.7 compatibility. All of these constuction can be easely upgraded by [this tool](https://github.com/asottile/pyupgrade) one time, but for now we should follow the Py3.7 syntax

orsinium: The file has `from __future__ import annotations` import at the beginning which makes all type annotations lazy and allows us to use the annotation syntax from future Python versions. See #386 and #388.

Lancetnik: Well, it was a smth new 4 me, thx

PR #397, opened 2022-11-29
Refactor a bit PullSubscription.fetch

orsinium:
1. Reuse the same code for pulling from the queue in both `_fetch_one` and `_fetch_n`.
2. Don't get from queue more than `batch` messages.
3. Refactor this loop a bit: don't check if the queue is empty before pulling from it. Instead, try pulling and exiting on the QueueEmpty exception raised. "Ask for forgiveness, not permission". And since before it had `except Exception: pass`, it used to have 3 function calls instead of one now (because it was going on the next iteration loop on failure instead of exiting the loop immediately).
4. Drop `except Exception: pass`, we don't want to suppress all ranges of exceptions when pulling from the queue.

### DIFF
--- a/nats/js/client.py
+++ b/nats/js/client.py
@@ -47,6 +47,8 @@ DEFAULT_JS_SUB_PENDING_BYTES_LIMIT = 256 * 1024 * 1024
 # Max history limit for key value.
 KV_MAX_HISTORY = 64
 
+NANOSECOND = 10**9
+
 
 class JetStreamContext(JetStreamManager):
     """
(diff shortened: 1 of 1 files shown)


orsinium:
I am the second person in the list of nats.py contributors. I love the project and I gave several public talks promoting it. But then something happened and my PRs stopped being reviewed. I'm glad that many years ago someone got time to bring it at least some closure, but I'm disappointed that this resolution is "closing as stale" instead of continued support of the most active outside contributor.

Oh well, that's probably for better. A lot has changed in 2 years.

PR #399, opened 2022-12-03
Send credentials even if no_auth_required is set on INFO

wallyqs:
At nats-server v2.9.8 there was an issue with `auth_required` not being sent by the server in some configurations.  Go client was not affected because it was not checking this property on the INFO line, so this changes the logic to be more similar to the Go client and always send the credentials whenever they are configured.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1854,6 +1854,7 @@ class Client:
             connection_completed, self.options["connect_timeout"]
         )
         if INFO_OP not in info_line:
+            # FIXME: Handle PING/PONG arriving first as well.
             raise errors.Error(
                 "nats: empty response from server when expecting INFO message"
             )
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -243,7 +243,7 @@ class Client:
         self._resp_sub_prefix: bytearray | None = None
         self._nuid = NUID()
         self._inbox_prefix = bytearray(DEFAULT_INBOX_PREFIX)
-        self._auth_configured = None
+        self._auth_configured: bool = False
 
         # NKEYS support
         #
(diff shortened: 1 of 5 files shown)


nats/aio/client.py
```diff
@@ -250,6 +250,7 @@ def __init__(self) -> None:
         self._resp_sub_prefix: Optional[bytearray] = None
         self._nuid = NUID()
         self._inbox_prefix = bytearray(DEFAULT_INBOX_PREFIX)
+        self._auth_configured = None
```
orsinium, line 253:
```suggestion
        self._auth_configured: bool = False
```

nats/aio/client.py
```diff
@@ -1860,10 +1863,15 @@ async def _process_connect_init(self) -> None:
...
         try:
             srv_info = json.loads(info.decode())
+            self._server_info = srv_info
         except Exception:
             raise errors.Error("nats: info message, json parse error")
 
-        self._server_info = srv_info
+        # In case 'auth_required' is part of INFO, then need to send credentials.
+        auth_required = srv_info.get("auth_required")
+        if auth_required:
```
orsinium, line 1872:
```suggestion
        if srv_info.get("auth_required", False):
```

tests/test_client.py
```diff
@@ -2620,6 +2621,33 @@ async def test_protocol_mixing(self):
...
+            error_cb=err_cb,
+        )
+        sub = await nc.subscribe("foo")
+        await nc.publish("foo", b'hello')
+        await nc.flush()
+        await asyncio.wait_for(fut, 1)
+        err = fut.result()
+        self.assertEqual(
+            str(err), 'nats: permissions violation for subscription to "foo"'
+        )
```
orsinium, line 2647:
```suggestion
        assert str(err) == 'nats: permissions violation for subscription to "foo"'
```

tests/utils.py
```diff
@@ -511,6 +511,25 @@ def tearDown(self):
         self.loop.close()
 
 
+class NoAuthUserServerTestCase(unittest.TestCase):
```
orsinium, line 514: Since you use pytest to run tests anyway, I think it would be better to use pytest fixtures instead of unittest-style hooks.

```python
@pytest.fixture
def no_auth_user_server():
    server = NATSD(...)
    yield server
    server.stop()
    
```

PR #401, opened 2022-12-16
Forbid None for Msg.Metadata fields

orsinium:
1. Remove `| None` from `Msg.Metadata` fields
2. Move the code constructing the Metadata into a constructor of Metadata (`Metadata._from_reply`). Let the class make itself.
3. Move token count constants on the module level, [as recommended by PEP 8](https://peps.python.org/pep-0008/#constants). There are more constants in `Ack` that shouldn't be there, I'll take care of them later.
4. Make `Msg.Metadata` immutable.


wallyqs (approved):
LGTM

PR #405, opened 2023-01-02
Convert `republish` value into dataclass

orsinium:
I haven't checked but it seems like the `republish` field in the API response is not converted to `RePublish` type from dict. I'll later see if I can add [typeguard](https://github.com/agronholm/typeguard) into tests to avoid such issues in the future.


orsinium:
Seems like there is a flaky test, I don't think I broke uvloop.

wallyqs (approved):

PR #419, opened 2023-02-14
Add custom inbox prefix support for JS pull subscribe

alparslanavci:
With [#276](https://github.com/nats-io/nats.py/pull/276), custom inbox prefix support is added to core NATS features, but not JS. This PR addresses this issue.

This is an important problem since it affects authorization permissions. If the user cannot specify a custom inbox prefix, the user's messages can be received by the non-authorized users who subscribed to the default inbox subject.


wallyqs (approved):
LGTM

casix:
Maybe I'm late but why do not use the inbox_prefix defined in the client class when the connection is created as a default?

PR #420, opened 2023-02-15
Fixes incorrect formatting in `.travis.yml`.

WillCodeCo:
The intention seems to be for Travis to allow the following build jobs to fail, without causing the whole build operation to break:
- where `python="3.8"`
- where `python="3.9"`
- where `name="Python: 3.9/uvloop"`

However the YAML file is incorrectly formatted to communicate this, meaning that builds are failing. This commit addresses this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,6 @@ jobs:
     install:
       - pip install uvloop
   allow_failures:
-    - python:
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
Thanks!!

PR #426, opened 2023-02-24
fix nkeys seed read \n caused invalid seed

aohan237:
if you read nkey files, the origin method will read \n , so it will cause nkey seed load error,  invalid nkey seed.

just strip the \n will do the work

this request only add a seed.strip() likes line before nkey.from_seed to pure the seed

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -25,13 +25,7 @@ from dataclasses import dataclass
 from email.parser import BytesParser
 from random import shuffle
 from secrets import token_hex
-from typing import (
-    Any,
-    Awaitable,
-    Callable,
-    Tuple,
-    Union,
-)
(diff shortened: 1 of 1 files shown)


nats/aio/client.py
```diff
@@ -561,6 +563,7 @@ def sig_cb(nonce: str) -> bytes:
                             # into the pre allocated bytearray.
                             user_seed = bytearray(nkey_size)
                             f.readinto(user_seed)  # type: ignore[attr-defined]
+                user_seed = user_seed.strip()
```
wallyqs, line 566: I think this is the fix?

aohan237: yes, all the seed  related with nkeys will all need this strip,  otherwise still cause invalid seed

PR #431, opened 2023-03-02
support parsing headers with fast-mail-parser

RouquinBlanc:
Hi,

While experimenting with a NATS subscriber with headers, I noticed that by adding a few headers to my publications, I ended up getting almost double CPU load on the subscriber side.

Looking more closely at what is happening, it looks like we end up spending almost 40% of the active time inside `email.parser.BytesParser`

Screenshot 2023-03-02 at 12 51 38

This PR is aimed at raising the awareness that this part is quite costly, and it exist a [rust-based equivalent](https://github.com/namecheap/fast_mail_parser/) doing the work way faster:

```
In [4]: data = b'tkid: fb9ac7f6942d0c7ccda526913af8f012\r\nconnclass: ads\r\nconnid: adsbRQB'

In [1]: from email.parser import BytesParser
In [2]: p = BytesParser()
In [5]: %timeit p.parsebytes(d)
31.5 µs ± 595 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [8]: from fast_mail_parser import parse_email
In [9]: %timeit parse_email(data)
2.63 µs ± 44.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

Please excuse the test not passing, there are some differences in the way the rust lib deals with empty values and I thought it was too preliminary to optimize at this point, without knowing if you think this is interesting enough to be implemented.

Before:
Screenshot 2023-03-02 at 15 32 53

After:
Screenshot 2023-03-02 at 15 33 04

Hoping this is useful!

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -34,6 +34,11 @@ from typing import (
 )
 from urllib.parse import ParseResult, urlparse
 
+try:
+    from fast_mail_parser import parse_email
+except ImportError:
+    parse_email = None
+
 import nats.js
 from nats import errors
 from nats.nuid import NUID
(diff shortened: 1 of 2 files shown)


bruth:
@RouquinBlanc Thanks for the analysis and contribution!

It looks like the CI failed due to [linting](https://app.travis-ci.com/github/nats-io/nats.py/jobs/597162669#L313). Do you mind reworking that list comprehension to reduce the line length?

wallyqs:
Thanks for the contribution! Merged with some fixes at: https://github.com/nats-io/nats.py/commit/ad5f2bf9515216951448d1a85d0a371a57a83f3e

Issue #434, opened 2023-03-14
Client.client_id is a number instead of string?


kkqy:
nats\aio\client.py
```
@property
    def client_id(self) -> Optional[str]:
        """
        Returns the client id which we received from the servers INFO
        """
        return self._client_id
```
The type of property "client_id" should be string according to the defination,but it returns a number in fact.
I am a newbie of nats, and I don't know the standard type of the client id of nats.

kkqy:
http monitor (http://127.0.0.1:8222/connz):
```
{
  "server_id": "NA2V3OLX4SIK4XFHAE4SGLJB7MNBIHFOJUQIXBPDD65375IN5ZL23EXC",
  "now": "2023-03-14T18:35:25.2339645Z",
  "num_connections": 3,
  "total": 3,
  "offset": 0,
  "limit": 1024,
  "connections": [
    {
      "cid": 84,
      "kind": "Client",
      "type": "nats",
      "ip": "127.0.0.1",
      "port": 63449,
      "start": "2023-03-14T18:23:52.5351742Z",
      "last_activity": "2023-03-14T18:23:59.3198083Z",
      "rtt": "1ns",
      "uptime": "11m32s",
      "idle": "11m25s",
      "pending_bytes": 0,
      "in_msgs": 0,
      "out_msgs": 2,
      "in_bytes": 0,
      "out_bytes": 80,
      "subscriptions": 1,
      "lang": "go",
      "version": "1.24.0"
    },
    {
      "cid": 85,
      "kind": "Client",
      "type": "nats",
      "ip": "::1",
      "port": 63491,
      "start": "2023-03-14T18:23:59.3006298Z",
      "last_activity": "2023-03-14T18:23:59.3011593Z",
      "rtt": "569µs",
      "uptime": "11m25s",
      "idle": "11m25s",
      "pending_bytes": 0,
      "in_msgs": 1,
      "out_msgs": 0,
      "in_bytes": 40,
      "out_bytes": 0,
      "subscriptions": 0,
      "lang": "python3",
      "version": "2.2.0"
    },
    {
      "cid": 86,
      "kind": "Client",
      "type": "nats",
      "ip": "::1",
      "port": 63494,
      "start": "2023-03-14T18:23:59.3192799Z",
      "last_activity": "2023-03-14T18:23:59.3198083Z",
      "rtt": "1ns",
      "uptime": "11m25s",
      "idle": "11m25s",
      "pending_bytes": 0,
      "in_msgs": 1,
[...]

PR #435, opened 2023-03-14
change the type of client_id to int

kkqy:
https://github.com/nats-io/nats.py/issues/434

The type of property "client_id" should be int instead of str.


wallyqs (approved):
LGTM

PR #457, opened 2023-06-08
fix #456: use uuid4 to gen unique task names

### LINKED ISSUE #456: Python 3.7 is not functional in 2.3.1
v2.3.1 with [Python 3.7](https://github.com/nats-io/nats.py/blob/main/pyproject.toml#L14) produces the following error.

```
AttributeError: '_asyncio.Task' object has no attribute 'get_name'
```

It looks like `get_name` was introduced in [Asyncio 3.8](https://docs.python.org/3.8/library/asyncio-task.html?highlight=task#asyncio.Task.get_name). `get_name` is only used once [here](https://github.com/nats-io/nats.py/blob/v2.3.1/nats/aio/subscription.py#L181).

Lancetnik:


wallyqs (approved):
LGTM

PR #459, opened 2023-06-15
Consumers info offset

raprek:
# ISSUE
Current js.consumers_info() method is able to return only 256 consumers. 
# SOLUTION
Nats server has method [jsConsumerListRequest](https://github.com/nats-io/nats-server/blob/367d8576128d2170a457a0c826fc8a04a936da85/server/jetstream_api.go#L4011), according code the final consumer list restricted by `JSApiListLimit` [code](https://github.com/nats-io/nats-server/blob/367d8576128d2170a457a0c826fc8a04a936da85/server/jetstream_api.go#L4094),  [JSApiListLimit](https://github.com/nats-io/nats-server/blob/367d8576128d2170a457a0c826fc8a04a936da85/server/jetstream_api.go#L420).
 However, it is possible to use `offset` param in body of `CONSUMER.LIST`, so that obtain all info manually by js.consumers_info(stream, offset)

### DIFF
--- a/nats/js/manager.py
+++ b/nats/js/manager.py
@@ -226,7 +226,7 @@ class JetStreamManager:
         )
         return resp['success']
 
-    async def consumers_info(self, stream: str, offset: Optional[int]) -> List[api.ConsumerInfo]:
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #460, opened 2023-06-18
Coherently use f-strings

floscha:
This PR simply replaces the rather outdated string formatting with f-strings which are already being used in other parts of the example code. Thus, this PR changes no functionality but simply makes the code more readable and coherent.

### DIFF
--- a/examples/basic.py
+++ b/examples/basic.py
@@ -49,7 +49,7 @@ async def main():
     # and trigger timeout if not faster than 500 ms.
     try:
         response = await nc.request("help", b'help me', timeout=0.5)
-        print("Received response: {response.data.decode()}")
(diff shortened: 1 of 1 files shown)


examples/basic.py
```diff
@@ -50,8 +49,7 @@ async def help_request(msg):
     # and trigger timeout if not faster than 500 ms.
     try:
         response = await nc.request("help", b'help me', timeout=0.5)
-        print("Received response: {message}".format(
-            message=response.data.decode()))
+        print("Received response: {response.data.decode()}")
```
charbonnierg, line 52: This string is missing the `f` prefix

floscha: My bad. Fixed with baf159b4de178a2479045badf2771b85daa740ec.

PR #463, opened 2023-06-21
Fix websocket connection when not using TLS

wallyqs:
`tls_required` shown on INFO was making the a non TLS client try to upgrade even though it was not configured.


bruth (approved):
LGTM. Noting this is a workaround for the 2.9.19 server..

PR #483, opened 2023-08-13
Add consumer backoff parameter support

csuriano23:
Since NATS v2.7.1 a backoff consumer config is available. This PR makes the setting available on python client


strangemonad:
Any chance this can get approved and merged? This seems like a useful option to expose (e.g. for workqueue patterns)

PR #494, opened 2023-09-11
Package metadata fixes

wallyqs:
Updates for some of the pypi metadata, plus now adding explicit version in package.


bruth (approved):
LGTM

PR #500, opened 2023-09-20
fix: KeyWatcher.updates() returns type of Entry | None

tekumara:
vscode/pyright infers the type of KeyWatcher.updates() as `Unknown`:

CleanShot 2023-09-20 at 21 10 01@2x

with this fix the type is now inferred as `Entry | None`:

CleanShot 2023-09-20 at 21 11 08@2x

in this example `e` could be `None`, so the type checker is now correctly telling us we haven't checked the `None` case (hence the red squiggly lines)

### DIFF
--- a/nats/js/kv.py
+++ b/nats/js/kv.py
@@ -298,7 +298,7 @@ class KeyValue:
 
         def __init__(self, js):
             self._js = js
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks for the fix!

PR #521, opened 2023-12-21
Added support to export subjects when fetching js.stream_info()

bsmith1180:
I added support for the python client to query subjects held in a stream.

The PR adds a field `subjects` to the js StreamState class (js/api.py)
and populates it through the JetStreamManager.stream_info() method.  It also adds a `subject_filter` keyword parameter to stream_info()

### DIFF
--- a/nats/js/manager.py
+++ b/nats/js/manager.py
@@ -68,12 +68,13 @@ class JetStreamManager:
             raise NotFoundError
         return info['streams'][0]
 
-    async def stream_info(self, name: str) -> api.StreamInfo:
+    async def stream_info(self, name: str, subjects_filter: str = '>') -> api.StreamInfo:
         """
         Get the latest StreamInfo by stream name.
         """
+        req_data = json.dumps({"subjects_filter": subjects_filter})
         resp = await self._api_request(
-            f"{self._prefix}.STREAM.INFO.{name}", timeout=self._timeout
+            f"{self._prefix}.STREAM.INFO.{name}", req_data.encode(), timeout=self._timeout
         )
         return api.StreamInfo.from_response(resp)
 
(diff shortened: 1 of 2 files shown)


nats/js/manager.py
```diff
@@ -68,12 +68,13 @@ async def find_stream_name_by_subject(self, subject: str) -> str:
             raise NotFoundError
         return info['streams'][0]
 
-    async def stream_info(self, name: str) -> api.StreamInfo:
+    async def stream_info(self, name: str, subjects_filter: str = '>') -> api.StreamInfo:
```
wallyqs, line 71: I think we want `subjects_filter` to be an optional parameter by default so something like:
```suggestion
    async def stream_info(self, name: str, subjects_filter: str = None) -> api.StreamInfo:
```

nats/js/manager.py
```diff
@@ -68,12 +68,13 @@ async def find_stream_name_by_subject(self, subject: str) -> str:
...
         """
         Get the latest StreamInfo by stream name.
         """
+        req_data = json.dumps({"subjects_filter": subjects_filter})
         resp = await self._api_request(
-            f"{self._prefix}.STREAM.INFO.{name}", timeout=self._timeout
+            f"{self._prefix}.STREAM.INFO.{name}", req_data.encode(), timeout=self._timeout
```
wallyqs, line 77: would prefer to send the subjects_filter JSON only when explicitly specifying a subject.

wallyqs:
Thanks for the PR, I left a couple of comments in case you want to address, I will include this fix in the next release.

wallyqs:
Commit included as part of this set of changes: https://github.com/nats-io/nats.py/pull/527

Issue #414, opened 2023-01-22
Question: Deterministic Subject token Partitioning example


alexreznikoff:
Hi! Сan you add an example for python how to write publishers and subscribers using  [Deterministic Subject token Partitioning](https://docs.nats.io/nats-concepts/subject_mapping#deterministic-subject-token-partitioning)

if i created 3 partitions should i create 3 subscribers (one for each partition)?
Is there no automatic rebalancing of partitions by subscribers?

jnmoyne:
There is nothing to do on the publishers, the Core NATS subject mappings are configured either in the server config or in the account JWT (depending on the security mode you are using).

If you partition a message stream into 3 (for example) then you could create 3 streams, each one capturing messages on a partition, and then you would run 3 consuming applications (for example) each subscribing to it's own stream consumer. You can modify the number of partitions in the subject mapping transform at any time (i.e. change the server config and signal the server to reload the config file, or push an updated account JWT), but up to you to start/stop the partitions' consuming applications according to your change.

Starting with NATS 2.10 you will be able to define subject mapping transform directly in the streams (i.e. the transform is defined as part of the stream definition)

wallyqs:
Added in #536

hungiito:
> Starting with NATS 2.10 you will be able to define subject mapping transform directly in the streams (i.e. the transform is defined as part of the stream definition)
@jnmoyne This is exciting! How can I subscribe to be notified when this feature is released?

wallyqs:
@hungiito you can find an example here: #536 release of client should be out later today

PR #536, opened 2024-02-20
Add subject_transform field to StreamConfig

charbonnierg:
# Problem

The `StreamConfig` dataclass does not define a field named `subject_transform`.

The subject transform feature has been mentioned in issue https://github.com/nats-io/nats.py/issues/414

# Proposed changes

Add the `subject_transform` field to the `StreamConfig` dataclass:

```python
from nats.js.api import SubjectTransform


cfg = StreamConfig(
    name="demo",
    subjects=["one.two.three"],
    subject_transform=SubjectTransform(src=">", dest="uno.>")
)
```

>Example is taken from the [the ADR](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-30.md#example-transforms)

### DIFF
--- a/nats/js/api.py
+++ b/nats/js/api.py
@@ -223,6 +223,13 @@ class RePublish(Base):
 
 
 @dataclass
+class SubjectTransform(Base):
+    """Subject transform to apply to matching messages."""
+    src: str
+    dest: str
+
+
+@dataclass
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

Issue #520, opened 2023-12-21
Add compression field for stream config


ierehon1905:
### Proposed change

in `StreamConfig` a field called `compression` may be added. Its type according to documentation would be 
```py
compression : Optional[Literal["s2"]] = None
```

### Use case

We use compression in our streams and would like to move from defining our streams in infra as code to just code approach

PR #537, opened 2024-02-20
Add compression field in StreamConfig

charbonnierg:
# Problem

The `StreamConfig` dataclass does not define a field named `compression`.

The compression feature has been mentioned in issue https://github.com/nats-io/nats.py/issues/520

# Proposed changes

Add the `compression` field to the `StreamConfig` dataclass:

```python
from nats.js.api import StoreCompression, StreamConfig


cfg = StreamConfig(
    name="demo",
    subjects=["one.two.three"],
    compression=StoreCompression.S2
)
```

According to the docs:

> If file-based and a compression algorithm is specified, the stream data will be compressed on disk. Valid options are nothing (empty string) or s2 for Snappy compression.

This PR introduces the `StoreCompression` enum which can take two values:

```python
class StoreCompression(str, Enum):
    NONE = ""
    S2 = "s2"
```

For backward compatibility, this field is optional and set to `None` by default.

### DIFF
--- a/nats/js/api.py
+++ b/nats/js/api.py
@@ -211,6 +211,19 @@ class DiscardPolicy(str, Enum):
     NEW = "new"
 
 
+class Compression(str, Enum):
+    """
+    If stream is file-based and a compression algorithm is specified,
+    the stream data will be compressed on disk.
+
+    Valid options are nothing (empty string) or s2 for Snappy compression.
+    Introduced in nats-server 2.10.0.
+    """
+
+    NONE = ""
+    S2 = "s2"
+
+
 @dataclass
 class RePublish(Base):
     """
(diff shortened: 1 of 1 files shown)


nats/js/api.py
```diff
@@ -211,6 +211,19 @@ class DiscardPolicy(str, Enum):
     NEW = "new"
 
 
+class Compression(str, Enum):
```
wallyqs, line 214:
```suggestion
class StoreCompression(str, Enum):
```

nats/js/api.py
```diff
@@ -260,6 +273,9 @@ class StreamConfig(Base):
     # Allow higher performance and unified direct access for mirrors as well.
     mirror_direct: Optional[bool] = None
 
+    # Allow compressing messages.
+    compression: Optional[Compression] = None
```
wallyqs, line 277:
```suggestion
    compression: Optional[StoreCompression] = None
```

wallyqs (approved):
LGTM, just changing type from Compression to StoreCompression so that it is closer to the Go client: https://github.com/nats-io/nats.go/blob/f7c1fe99926cd7a2b394709f7693add6e438cc37/js.go#L3789-L3818

PR #539, opened 2024-02-22
Add metadata field to StreamConfig and ConsumerConfig

charbonnierg:
# Problem

The `StreamConfig` dataclass does not define a field named `metadata`, neither does the `ConsumerConfig` dataclass.

# Proposed changes

Add the `metadata` field to the `StreamConfig` dataclass and `ConsumerConfig` dataclass:

```python
from nats.js.api import ConsumerConfig, StreamConfig


stream_cfg = StreamConfig(
    name="demo",
    subjects=["one.two.three"],
    metadata={"foo": "bar"}
)
```

As stated in the [schemas](https://github.com/nats-io/jsm.go/blob/2f16a88ec1aedc16995c3fa4370fd327c372372f/schemas/jetstream/api/v1/stream_configuration.json#L408) the `metadata` field is typed as `Optional[Dict[str, str]]`


wallyqs (approved):
LGTM

PR #546, opened 2024-03-15
Allow passing file contents as user_credentials, not just file paths

roguh:
Implemented with tempfile.NamedTemporaryFile, open to new ideas. We can potentially get rid of tempfile and use the contents directly without relying on `open()`.

https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html

Hi, from PowerFlex.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -30,11 +30,13 @@ from typing import (
     Any,
     Awaitable,
     Callable,
-    Tuple,
-    Union,
-    List,
-    Optional,
     Dict,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+    TypedDict,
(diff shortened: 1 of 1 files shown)


wallyqs:
Thanks for the PR, I was looking a how we could avoid the tempfile to support this usage. What do you think of something like this? https://github.com/nats-io/nats.py/commit/e1c692485405aa96802eb184031916700349447b
This way usage would be:

```python
await nats.connect(user_credentials=RawCredentials("<creds file contents as string>"))
```

roguh:
That's look good. I would also prefer avoiding temp files, but wasn't sure how much of this code to change.

This also works well with type checkers and Python 3.7 I assume.

PR #568, opened 2024-06-12
Add test workflow

caspervonb:


.github/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches:
+      - main
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.7, 3.8, 3.9, 3.10, 3.11, 3.12]
```
wallyqs, line 17: we can remove 3.7 since EOL

PR #569, opened 2024-06-12
Add async publishing to JetStream

caspervonb:
- Add a `JetStreamContext.publish_async` method which does not await for ack.
- Add a `JetStreamContext.publish_async_completed` method which awaits for all pending publish acks.
- Add a `JetStreamContext.publish_async_pending` method which returns the number of pending acks.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1023,21 +1023,23 @@ class Client:
             msg = await self._request_new_style(
                 subject, payload, timeout=timeout, headers=headers
             )
+        if msg.headers and msg.headers.get(nats.js.api.Header.STATUS
+                                           ) == NO_RESPONDERS_STATUS:
+            raise errors.NoRespondersError
         return msg
 
-    async def _request_future(
+    async def _request_new_style(
         self,
         subject: str,
         payload: bytes,
         timeout: float = 1,
         headers: Optional[Dict[str, Any]] = None,
-    ) -> asyncio.Future[Msg]:
+    ) -> Msg:
         if self.is_draining_pubs:
             raise errors.ConnectionDrainingError
 
         if not self._resp_sub_prefix:
             await self._init_request_sub()
-
         assert self._resp_sub_prefix
 
         # Use a new NUID + couple of unique token bytes to identify the request,
@@ -1052,38 +1054,22 @@ class Client:
             subject, payload, reply=inbox.decode(), headers=headers
         )
 
-        async def wait_for_future() -> Msg:
+        # Wait for the response or give up on timeout.
+        try:
+            msg = await asyncio.wait_for(future, timeout)
+            return msg
+        except asyncio.TimeoutError:
             try:
-                msg = await asyncio.wait_for(future, timeout)
-                if msg.headers and msg.headers.get(nats.js.api.Header.STATUS
-                                                   ) == NO_RESPONDERS_STATUS:
-                    raise errors.NoRespondersError
+                # Double check that the token is there already.
+                self._resp_map.pop(token.decode())
+            except KeyError:
+                await self._error_cb(
+                    errors.
+                    Error(f"nats: missing response token '{token.decode()}'")
+                )
 
-                return msg
-            except asyncio.TimeoutError:
-                try:
-                    # Double check that the token is there already.
-                    self._resp_map.pop(token.decode())
-                except KeyError:
-                    await self._error_cb(
-                        errors.Error(f"nats: missing response token '{token}'")
-                    )
-
-                future.cancel()
-                raise errors.TimeoutError
-
-        return asyncio.ensure_future(wait_for_future())
-
-    async def _request_new_style(
-        self,
-        subject: str,
-        payload: bytes,
-        timeout: float = 1,
-        headers: Optional[Dict[str, Any]] = None,
-    ) -> Msg:
-        future = await self._request_future(subject, payload, timeout, headers)
-        msg = await future
-        return msg
+            future.cancel()
+            raise errors.TimeoutError
 
     def new_inbox(self) -> str:
         """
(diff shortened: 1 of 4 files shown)


nats/js/client.py
```diff
@@ -136,6 +136,42 @@ async def publish(
             raise nats.js.errors.APIError.from_error(resp['error'])
         return api.PubAck.from_response(resp)
 
+    async def publish_async(
```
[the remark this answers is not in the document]
wallyqs: publish_async like the go client built-in sounds good for parity

nats/js/client.py
```diff
@@ -136,6 +136,42 @@ async def publish(
...
+        hdr = headers
+        if timeout is None:
+            timeout = self._timeout
+        if stream is not None:
+            hdr = hdr or {}
+            hdr[api.Header.EXPECTED_STREAM] = stream
+
+        fut = await self._nc._request_future(
+            subject, payload, timeout=timeout, headers=hdr
+        )
```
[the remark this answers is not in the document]
wallyqs: https://github.com/nats-io/nats.go/blob/5dbd825c2cf21d81bbc92cbfe5a4cb73aae7a169/jetstream/publish.go#L294-L303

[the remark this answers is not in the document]
wallyqs: we can try to do similar with 4K as the default: https://github.com/nats-io/nats.go/blob/main/js.go#L224-L225
this internal acks queue would be then bound to the state of each jetstream context instance

nats/aio/client.py
```diff
@@ -1023,23 +1023,21 @@ async def request(
             msg = await self._request_new_style(
                 subject, payload, timeout=timeout, headers=headers
             )
-        if msg.headers and msg.headers.get(nats.js.api.Header.STATUS
-                                           ) == NO_RESPONDERS_STATUS:
-            raise errors.NoRespondersError
```
wallyqs, line 1028: removing this here would make `await nc.request` not handle no responders correctly no?

nats/js/client.py
```diff
@@ -103,6 +111,50 @@ def _jsm(self) -> JetStreamManager:
...
+        print("PAF SUB CALLBACK", msg)
+        token = msg.subject[len(self._nc._inbox_prefix) + 22 + 2:]
+        try:
+            paf = self._paf_map.get(token)
+            if not paf:
+                return
+
+            # Handle no responders
+            if msg.headers and msg.headers.get(api.Header.STATUS) == NO_RESPONDERS_STATUS:
+                print("RAISING NO RESPONDER")
```
wallyqs, line 139: leftover print

tests/test_js.py
```diff
@@ -80,6 +80,49 @@ async def test_publish_verbose(self):
...
+            self.assertEqual(result.stream, "QUUX")
+            self.assertEqual(result.seq, seq)
+
+        # Check that stalled is raised and can be recovered
+        futures = []
+        while True:
+            try:
+               future = await js.publish_async("quux", b'bar:1')
+               futures.append(future)
+            except TooManyStalledMsgsError:
```
wallyqs, line 110: In the go client there is also a publish_async_complete to wait when the jetstream context is not stalled anymore, we could do something similar here (could be a different PR)

nats/js/client.py
```diff
@@ -136,6 +188,48 @@ async def publish(
...
+            timeout = self._timeout
+
+        if stream is not None:
+            hdr = hdr or {}
+            hdr[api.Header.EXPECTED_STREAM] = stream
+
+        # Use a new NUID + couple of unique token bytes to identify the request,
+        # then use the future to get the response.
+        token = self._nc._nuid.next()
+        token.extend(token_hex(2).encode())
```
wallyqs, line 221: like this approach 👍

nats/js/errors.py
```diff
@@ -133,6 +133,15 @@ def __str__(self) -> str:
         return "nats: no response from stream"
 
 
+class TooManyStalledMsgsError(Error):
+    """
+    Raised when too many outstanding async messages are waiting for ack.
```
wallyqs, line 138:
```suggestion
    Raised when too many outstanding async published messages are waiting for ack.
```

nats/js/client.py
```diff
@@ -136,6 +185,50 @@ async def publish(
...
+        payload: bytes = b'',
+        stream: Optional[str] = None,
+        headers: Optional[Dict] = None,
+    ) -> asyncio.Future[api.PubAck]:
+        """
+        emits a new message to JetStream and returns a future that can be awaited for acknowledgement.
+        """
+
+        if len(self._paf_map) > self._pending_acks_limit:
+            raise nats.js.errors.TooManyStalledMsgsError
```
wallyqs, line 200: Here instead of raising the error I think we should wait for a default stall time, then when that times out we would raise the error: 

https://github.com/nats-io/nats.go/blob/main/js.go#L951-L958

once it has been stall for that long then the error would be raised. 
The default stall wait in Go is 200ms but for Python it could be longer, thinking maybe a one second.

nats/js/client.py
```diff
@@ -136,6 +185,50 @@ async def publish(
...
+            subject, payload, reply=inbox.decode(), headers=hdr
+        )
+
+        return future
+
+    async def publish_async_completed(self) -> None:
+        """
+        waits for all currently pending async publishes to be completed.
+        """
+        await asyncio.gather(*self._paf_map.values(), return_exceptions=True)
```
wallyqs, line 230: might be good to also add `js.publish_async_pending` to get a snapshot of how many publish acks are inflight:
https://github.com/nats-io/nats.go/blob/main/js.go#L780-L784

wallyqs (approved):
LGTM! 👍

Issue #553, opened 2024-04-08
KV should raise an error if the key contains unacceptable characters


kevinhikaruevans:
### Proposed change

If the key contains unacceptable characters (e.g. whitespace), an error should be raised. Maybe `BadSubjectError` can be reused or a new error can be created (e.g. `BadKeyError`).

It currently gives an `UnexpectedEOF` error and causes a `TimeoutError`. This is fairly confusing because it doesn't necessarily imply the key is bad and could be remedied by a quick check on the key.

### Use case

I had somebody else feeding me some data, which was being fed into nats kv. I noticed that occasionally, an entry would cause an eof error and a timeout error and it wasn't very clear why. I finally realized that the user was putting spaces in the keys, which was causing problems in the underlying `publish()` call.

### Contribution

Sure. I would assume it would be putting in a little check in `put()`

skewty:
Is this not the Primitive Obsession code smell?

https://refactoring.guru/smells/primitive-obsession

```python
from typing import Final, Self
import re

class Key(str):
    PATTERN: Final = re.compile(r"^[A-Za-z0-9_-]+$")  # whatever you want / need

    def __new__(cls, *args, **kwargs) -> Self:
        value = super().__new__(cls, *args, **kwargs)
        # really any checks you require here but likely regex will suffice
        if cls.PATTERN.match(value) is None:
            raise ValueError(f'Key {value} contains invalid characters')
        return value
```

skewty:
IMO these Value classes should be used elsewhere in the code-base as well.  Example bucket name.

PR #571, opened 2024-06-17
Enforce valid identifiers for "key value keys"

caspervonb:
Closes https://github.com/nats-io/nats.py/issues/553

### DIFF
--- a/nats/js/kv.py
+++ b/nats/js/kv.py
@@ -131,11 +131,16 @@ class KeyValue:
         self._js = js
         self._direct = direct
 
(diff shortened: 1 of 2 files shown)


wallyqs:
We should have this but this client at least it needs to be opt-in at least for a few releases I think...

wallyqs:
change LGTM but need to add a way to be opt into previous behavior just in case, adding a validate_keys method on top of this PR here: https://github.com/nats-io/nats.py/pull/706

Issue #391, opened 2022-11-23
"Error: nats: missing response token '...'"


ronigober:
Hey :)
First, it's important to know that we initialize the NATS client with `error_cb=error_callback`, 
and `error_callback(msg)` raise the given message as an exception.

We get from time to time an exception of `Error: nats: missing response token '...'` when sending a request.
We think that it occurs when the connection between the NATS client to the NATS server is not in it's best...
When we search for the root cause of this error, we found that it could be a race condition issue in the NATS client.

We found that in the `_request_new_style` function, for every new request we generate a token that will be used as part of the inbox and will be added to the `_resp_map` dictionary to later access its message.
Then, while waiting for the future, if we get `asyncio.TimeoutError` exception, we try to pop the token from the `_resp_map` dictionary (_"# Double check that the token is there already."_),
and because it doesn't exists, `_error_cb` called with the mentioned `Error: nats: missing response token '...'` message.

```
    async def _request_new_style(
        self,
        subject: str,
        payload: bytes,
        timeout: float = 1,
        headers: Dict[str, Any] = None,
    ) -> Msg:
        if self.is_draining_pubs:
            raise errors.ConnectionDrainingError

        if not self._resp_sub_prefix:
            await self._init_request_sub()
        assert self._resp_sub_prefix

        # Use a new NUID + couple of unique token bytes to identify the request,
        # then use the future to get the response.
        token = self._nuid.next()
        token.extend(token_hex(2).encode())
        inbox = self._resp_sub_prefix[:]
        inbox.extend(token)
        future: asyncio.Future = asyncio.Future()
        self._resp_map[token.decode()] = future
        await self.publish(
            subject, payload, reply=inbox.decode(), headers=headers
        )

        # Wait for the response or give up on timeout.
        try:
            msg = await asyncio.wait_for(future, timeout)
            return msg
        except asyncio.TimeoutError:
            try:
                # Double check that the token is there already.
                self._resp_map.pop(token.decode())
            except KeyError:
                await self._error_cb(
                    errors.
                    Error(f"nats: missing response token '{token.decode()}'")
                )

            future.cancel()
            raise errors.TimeoutError
```

We saw that the only other place that we pop the token from the `_resp_map` dictionary is in the `_request_sub_callback` function. In it we pop the token if we got the future or we get one of the `syncio.CancelledError`, `asyncio.InvalidStateError` exceptions.

```
    async def _request_sub_callback(self, msg: Msg) -> None:
[...]

PR #573, opened 2024-06-17
Fix potential race condition in `nats.aio.Client.request`

caspervonb:
Eliminate the side effects of a timeout and callback racing to resolve a request future.
The race may still occur but shouldn't be harmful.

Fixes https://github.com/nats-io/nats.py/issues/391
Fixes https://github.com/nats-io/nats.py/issues/572


wallyqs (approved):
LGTM

PR #593, opened 2024-07-31
JetStreamManager: During a direct GET, do not conflate ''seq == 0'' and ''seq is None''

charles-dyfis-net:


charles-dyfis-net:
> Hey thank you for your contribution. Are you able to come up with a test for this?

Honestly, y'all -- as folks who understand the original intent of writing `not seq` -- are probably in a better place than I am to understand the server-side semantics and come up with a narrowly targeted test case.

Applying this patch was _part_ of resolving an issue I encountered in my client-side code, but I also started passing in `subject` at the same time; I'd need to go back and reverse-engineer the problem a bit.

wallyqs (approved):
LGTM

Issue #591, opened 2024-07-30
`streams_info` Method Only Returns First Page of Streams


koushamad:
### Observed behavior

The method only returns the first page of streams (up to 256 streams) and does not paginate through the remaining streams.

### Expected behavior

The `streams_info` method should continuously request stream information from the server, paginating through the results using an offset, and yield each stream's information as it is retrieved.

### Server and client version

nats.py: v2.8.0
nast-server: v2.10.18

PR #594, opened 2024-08-01
Fix stream info pagination and add test for empty stream list

koushamad:
Fix:  streams_info Method Only Returns First Page of Streams #591 

Observed behavior
The method only returns the first page of streams (up to 256 streams) and does not paginate through the remaining streams.

Expected behavior
This merge request adds a test to verify that the `streams_info` method correctly handles offsets, ensuring proper pagination when retrieving streams. Additionally, it implements the necessary changes to the `streams_info` method to support pagination.

- Added a test to verify the `streams_info` method handles offsets correctly.
- Implemented pagination in the `streams_info` method to support retrieving streams with an offset.
- Implemented `streams_info_iterator` method to retrieve a list of streams as an iterator.
- Created `StreamsListIterator` class to handle iteration over the list of streams.

### DIFF
--- a/nats/js/api.py
+++ b/nats/js/api.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, fields, replace
 from enum import Enum
-from typing import Any, Dict, Optional, TypeVar, List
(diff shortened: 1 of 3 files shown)


koushamad:
@caspervonb test fixed.

PR #597, opened 2024-08-09
Resolves #596

### LINKED ISSUE #596: Memory leak in ObjectStore.get due to uncancelled tasks
### Observed behavior

Growing number of active coroutines that are never garbage collected. Traces attached.

### Expected behavior

Coroutines gets garbage collected.

### Server and client version

nats: 2.10.18

### Host environment

_No response_

### Steps to reproduce

There is a memory leak in the `ObjectStore.get` method that causes the number of active coroutines to grow over time and never decrease. This issue occurs because the long-running tasks created during subscription (`_hbtask` and `_fctask`) are not being cancelled when the subscription is unsubscribed.

**Problem:**

1. When creating a subscription in `ObjectStore.get`, two long-running tasks may be created: `_hbtask` for idle heartbeat and `_fctask` for ordered consumers.
2. These tasks are not cancelled when sub.unsubscribe() is called.
3. As a result, these tasks continue to run and hold references to the subscription and other objects, preventing garbage collection.

**Impact:**

* The number of active coroutines can grow to very high numbers (observed up to 20K) over time.
* This leads to increased memory usage and potential performance degradation.

**Proposed Fix:**
Modify the `ObjectStore.get` method to explicitly cancel these tasks after unsubscribing:

Ref: https://github.com/nats-io/nats.py/blob/d6f4f97dc646d7a7c7cf1322f7aac899d4ddc6bf/nats/js/object_store.py#L228

Fix:
```python
if tokens[8] == OBJ_NO_PENDING:
    await sub.unsubscribe()

    # Cancel running tasks
    if sub._sub._jsi._hbtask:
        sub._sub._jsi._hbtask.cancel()
    if sub._sub._jsi._fctask:
        sub._sub._jsi._fctask.cancel()
```

**Additional Considerations:**

1. This fix should be applied to `ObjectStore.get`. 
2. Other methods in `ObjectStore` and `JetStreamContext` that create subscriptions should be reviewed for similar issues.
3. Consider modifying the `Subscription.unsubscribe` method to automatically cancel these tasks for a more comprehensive solution.

By implementing this fix, we expect to see a significant reduction in the number of active coroutines over time, leading to improved memory management and performance.

**Dumps:**
* [cpu_coroutines_dump.txt](https://github.com/user-attachments/files/16549977/cpu_coroutines_dump.txt)
* [io_coroutines_dump.txt](https://github.com/user-attachments/files/16549978/io_coroutines_dump.txt)

pySilver:
Fixes memory leak described in https://github.com/nats-io/nats.py/issues/596

Signed-off-by: Antoni Martyniuk

### DIFF
--- a/nats/js/client.py
+++ b/nats/js/client.py
@@ -942,6 +942,20 @@ class JetStreamContext(JetStreamManager):
                         await self._conn.publish(fc_reply)
                         self._sub._jsi._fcr = None
             return msg
+        
+        async def unsubscribe(self, limit: int = 0):
+            """
+            Unsubscribes from a subscription, canceling any heartbeat and flow control tasks,
+            and optionally limits the number of messages to process before unsubscribing.
+            """
+            await super().unsubscribe(limit)
+
+            if self._sub._jsi._hbtask:
+                self._sub._jsi._hbtask.cancel()
+
+            if self._sub._jsi._fctask:
+                self._sub._jsi._fctask.cancel()
+            
 
     class PullSubscription:
         """
--- a/tests/test_js.py
+++ b/tests/test_js.py
@@ -1730,6 +1730,43 @@ class SubscribeTest(SingleJetStreamServerTestCase):
 
         await nc.close()
 
+    @async_test
+    async def test_unsubscribe_coroutines(self):
+        nc = await nats.connect()
+        js = nc.jetstream()
+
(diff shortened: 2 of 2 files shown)


pySilver:
Unfortunately, I have no idea how to reformat file. When running `make format` it does format way to much code outside of my changes.

PR #602, opened 2024-09-04
feat(kv): implement KeysWithFilters method for key filtering in KV bucket

somratdutta:
- Added KeysWithFilters method to KeyValue store to allow retrieving keys that match specific filters.
- The method returns a list of keys from the KV bucket, applying the provided filters for pattern matching.
- Checks consumer info to ensure filters are properly supported by the NATS server, with warnings for server versions < 2.10.

Resolves: [#575](https://github.com/nats-io/nats.py/issues/575)

### DIFF
--- a/nats/js/kv.py
+++ b/nats/js/kv.py
@@ -359,6 +359,48 @@ class KeyValue:
 
         return keys
 
+    async def keys_with_filters(self, filters: List[str] = None, **kwargs) -> List[str]:
+        """
+        Returns a filtered list of keys in the bucket.
+        If no filters are provided, all keys are returned.
+        """
+        watcher = await self.watchall(
+            ignore_deletes=True,
+            meta_only=True,
+        )
+
+        keys = []
+
+        # Check consumer info and make sure filters are applied correctly
+        try:
+            consumer_info = await watcher.sub.consumer_info()
+            if consumer_info and filters:
+                # If NATS server < 2.10, filters might be ignored.
+                if consumer_info.config.filter_subject != ">":
+                    print("Warning: Server may ignore filters if version is < 2.10.")
+        except Exception as e:
+            print(f"Error fetching consumer info: {e}")
+
+        async for key in watcher:
+            # None entry is used to signal that there is no more info.
+            if not key:
+                break
+
+            # Apply filters if any were provided
+            if filters:
+                if any(f in key.key for f in filters):
+                    keys.append(key.key)
+            else:
+                # No filters provided, append all keys
+                keys.append(key.key)
+
+        await watcher.stop()
+
+        if not keys:
+            raise nats.js.errors.NoKeysError
+
+        return keys
+
     async def history(self, key: str) -> List[Entry]:
         """
         history retrieves a list of the entries so far.
--- a/examples/kv.py
+++ b/examples/kv.py
@@ -11,9 +11,19 @@ async def main():
 
     # Set and retrieve a value
     await kv.put('hello', b'world')
+    await kv.put('goodbye', b'farewell')
+    await kv.put('greetings', b'hi')
+    await kv.put('greeting', b'hey')
+
+    # Retrieve and print the value of 'hello'
     entry = await kv.get('hello')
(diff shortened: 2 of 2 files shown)


nats/js/kv.py
```diff
@@ -359,6 +359,48 @@ async def keys(self, **kwargs) -> List[str]:
 
         return keys
 
+    async def keys_with_filters(self, filters: List[str] = None, **kwargs) -> List[str]:
```
wallyqs, line 362: seems we can change this to be a param to `keys()` instead? 
```
keys(filters=[...])
```
no filter means just using `keys()`

nats/js/kv.py
```diff
@@ -359,6 +359,48 @@ async def keys(self, **kwargs) -> List[str]:
...
+
+        keys = []
+
+        # Check consumer info and make sure filters are applied correctly
+        try:
+            consumer_info = await watcher.sub.consumer_info()
+            if consumer_info and filters:
+                # If NATS server < 2.10, filters might be ignored.
+                if consumer_info.config.filter_subject != ">":
+                    print("Warning: Server may ignore filters if version is < 2.10.")
```
wallyqs, line 380: there is a logger that can be used instead

nats/js/kv.py
```diff
@@ -359,6 +359,48 @@ async def keys(self, **kwargs) -> List[str]:
...
+        except Exception as e:
+            print(f"Error fetching consumer info: {e}")
```
wallyqs, line 382: should we raise the error here instead?

wallyqs (changes requested):
Thanks for the contribution! I left some comments, before including we need to add a test for the feature.

nats/js/kv.py
somratdnutanix: @wallyqs I followed the convention as mentioned in this [issue](https://github.com/nats-io/nats.py/issues/575).
Should I change it as you mentioned?

somratdutta: @wallyqs I followed the convention as mentioned in this [issue](https://github.com/nats-io/nats.py/issues/575).
Should I change it as you mentioned?

wallyqs: yes I think for the nats.py client that has access to passing options that would be better

nats/js/kv.py
```diff
@@ -337,21 +340,41 @@ async def watchall(self, **kwargs) -> KeyWatcher:
...
             ignore_deletes=True,
             meta_only=True,
         )
         keys = []
 
+        # Check consumer info and make sure filters are applied correctly
+        try:
+            consumer_info = await watcher._sub.consumer_info()
+            if consumer_info and filters:
+                # If NATS server < 2.10, filters might be ignored.
```
wallyqs, line 358: there is a `connected_server_version` which can be used to check: https://github.com/nats-io/nats.py/blob/main/nats/aio/client.py#L1210C9-L1210C33

nats/js/kv.py
```diff
@@ -337,21 +340,41 @@ async def watchall(self, **kwargs) -> KeyWatcher:
...
+                if consumer_info.config.filter_subject != ">":
+                    logger.warning("Server may ignore filters if version is < 2.10.")
+        except Exception as e:
+            logger.error(f"Error fetching consumer info: {e}")
```
wallyqs, line 362: we can remove this log and just raise the exception

nats/js/kv.py
somratdutta: this can't be invoked without a `nc` I suppose?
I checked the usage of `connected_server_version` and this is how its been used
`server_version = nc.connected_server_version`

wallyqs: think it is part of the state already here, maybe via `self._js._nc`

somratdutta: no,
I have tried that already 😄

wallyqs: I see, thanks for checking

wallyqs (approved):
LGTM, thanks!

PR #605, opened 2024-09-19
Fix typing of `JetStreamContext.publish`

rijenkii:
Argument `headers` has type `Optional[Dict[Unknown, Unknown]]`, which makes pyright scream and shout in strict mode:
```
Type of "publish" is partially unknown
  Type of "publish" is "(subject: str, payload: bytes = b'', timeout: float | None = None, stream: str | None = None, headers: Dict[Unknown, Unknown] | None = None) -> Coroutine[Any, Any, PubAck]"PylancereportUnknownMemberType
```

### DIFF
--- a/nats/js/client.py
+++ b/nats/js/client.py
@@ -19,7 +19,7 @@ import json
 import time
 from email.parser import BytesParser
 from secrets import token_hex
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #607, opened 2024-09-20
Validate stream name in `add_stream` method

### LINKED ISSUE #471: Passing wrong stream name to `add_stream` throws `errors.TimeoutError`
This may be similar to #305 and/or #437.

Following code:
```python
import nats
import pytest

@pytest.mark.asyncio
async def test_js_wrong_name():
    nc = await nats.connect("localhost")
    js = nc.jetstream()
    await js.add_stream(name="stream.with.wrong.name")

    await nc.close()
```

raises timeout on `js.add_stream`:

```
>           raise errors.TimeoutError
E           nats.errors.TimeoutError: nats: timeout

../.venv/lib/python3.11/site-packages/nats/aio/client.py:1025: TimeoutError
```

If the stream name is correct it's ok.

ff137:
Fixes #471, and hopefully saves someone else time who would encounter the same issue. :-)

A less verbose error message / more succinct "valid name" rule may be preferable. So, feedback or suggestions welcome. I just went off of what I found in these docs: https://docs.nats.io/nats-concepts/jetstream/streams#configuration:

```md
Names cannot contain whitespace, `.`, `*`, `>`, path separators (forward or backwards slash), and non-printable characters.
```

### DIFF
--- a/nats/js/manager.py
+++ b/nats/js/manager.py
@@ -98,11 +98,11 @@ class JetStreamManager:
 
         # Validate stream name
         invalid_chars = set(".*>/\\")
-        if (
-            any(char in stream_name for char in invalid_chars) or  # No invalid characters
-            any(char.isspace() for char in stream_name) or  # No whitespace
-            not stream_name.isprintable()  # Must be printable
-        ):
+        has_invalid_chars = any(char in stream_name for char in invalid_chars)
+        has_whitespace = any(char.isspace() for char in stream_name)
+        is_not_printable = not stream_name.isprintable()
+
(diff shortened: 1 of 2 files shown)


ff137:
Thanks! 

Btw, being new to the nats python client, I only picked it up for this add stream method. So not sure where else needs name validation like this, or where to move such a method so it's deduplicated / more configurable.

I'm open to suggestions, to add it in other relevant places too. Or of course, that can be done separately

wallyqs (approved):
LGTM

PR #608, opened 2024-09-20
:art: clean up imports

ff137:
Noticed that pyproject.toml configures isort. But, imports weren't sorted.

So, just runs `isort` (ensuring the config is used), and ran `autoflake --remove-all-unused-imports` for good measure. Just some artsy cleanup.


ff137:
@caspervonb I've rebased 👌

I also tried adding an isort check to the cicd. 
For that, I need to add isort as a dev dependency to the Pipfile. 
And, define a `.isort.cfg` file which includes the pyproject.toml `[tool.isort]` settings, because they weren't being picked up. 
With this, I could get the isort check to pass in CICD.

But, I had to update the lock file. And it looks like it hasn't updated been in quite a while (lots of version bumps), and some tests seem to be failing as a result. So ... it's a bit too much to try squeeze in to this PR, so I'll make a separate one for that: #611 

The lockfile update and test fix should probably be done separately. I'll see if I can take a quick crack at it

ff137:
Oops. autoflake removed the `aiohttp` import in test_client_websocket.py:
```py
try:
    import aiohttp
    aiohttp_installed = True
except ModuleNotFoundError:
    aiohttp_installed = False
```
which broke one of the tests

ff137:
@caspervonb I already took a crack at it: #611

ff137:
rebased 👌

ff137:
@caspervonb I cherry-picked the commits from #611 here, and moved the cicd check to the new check.yaml

ff137:
> I've been thinking of replacing isort with ruff, as it'll serve as a replacement for flake, isort and yapf. But I'm okay with merging this in the meantime.

Good stuff, I actually wanted to recommend ruff instead of isort/pyflake/yapf, but didn't want to impose 😄

PR #611, opened 2024-09-24
Add isort check to CICD

ff137:
Pending #608

👷 Adds an isort check to the cicd.
➕ For that, I needed to add isort as a dev dependency to the Pipfile. 
⬆️ Updated lock file (_lots_ of version bumps! Last updated Jun 2023)
🔧 And, define `.isort.cfg`, which includes the pyproject.toml `[tool.isort]` settings, because they weren't being picked up.


ff137:
Including this in #608

PR #614, opened 2024-09-30
Implement `request_many`

caspervonb:


wallyqs:
@caspervonb let's give it a try at implementing the request mux version of the request many, the current nats.go implementation looks like in that direction: https://github.com/nats-io/nats.go/pull/1722/files

PR #623, opened 2024-10-18
Allow passing pathlib.Path derived types to user_credentials

johnweldon:
Allow passing Path like objects for the connect option `user_credentials`

Previously if you passed in a Path (or derived type like PosixPath), it would silently fail to actually read the credentials from that path.


wallyqs (approved):
LGTM

PR #632, opened 2024-11-09
Added Version to header which points to latest release

dudwe:
I have updated the docs to add the **version** specified in the Makefile as a link in the header.

The link will point to the release page for the given version:

image

Currently the version specified in the file is v2.0.0. I have made the docs point to this release: https://github.com/nats-io/nats.py/releases/tag/v2.0.0


caspervonb:
Closing as stale thank you, didn't make it to the hackathon due to health issues and never caught up with these.

Issue #287, opened 2022-03-27
nc.flush times out after keyboard interrupt


pohlt:
# Setup 
* nats-py: 2.0.0
* Python 3.10.3
* OS: Linux 5.16.16

# Test Code

```python
import asyncio
from nats.aio.client import Client

async def main():
    nc = Client()
    await nc.connect("nats://localhost:54100", name="test")
    sub = await nc.subscribe("subject")

    try:
        await asyncio.Future()
    except asyncio.CancelledError:
        print("cancelled")

    print("before unsubscribe")
    await sub.unsubscribe()
    print("before flush")
    await nc.flush()
    print("before close")
    await nc.close()
    print("done")


asyncio.run(main(), debug=True)
```

# Behavior

When I run this code, I can successfully connect to the NATS server and subscribe to the topic. The code will then wait for a keyboard interrupt which cancels the `Future`. During the shutdown, I unsubscribe from the topic, flush, and close the connection.

I experience two unexpected behaviors:
1. The `unsubscribe` never actually happens.
2. The `flush` times out.

Here is the output from the NATS server started with `-D -V`:

```
[1] 2022/03/27 08:32:09.155429 [DBG] [::1]:60536 - cid:49 - Client connection created
[1] 2022/03/27 08:32:09.156595 [TRC] [::1]:60536 - cid:49 - <<- [CONNECT {"echo": true, "headers": true, "lang": "python3", "name": "test", "no_responders": true, "pedantic": false, "protocol": 1, "verbose": false, "version": "2.0.0"}]
[1] 2022/03/27 08:32:09.156632 [TRC] [::1]:60536 - cid:49 - "v2.0.0:python3:test" - <<- [PING]
[1] 2022/03/27 08:32:09.156635 [TRC] [::1]:60536 - cid:49 - "v2.0.0:python3:test" - ->> [PONG]
[1] 2022/03/27 08:32:09.157726 [TRC] [::1]:60536 - cid:49 - "v2.0.0:python3:test" - <<- [SUB subject  1]
[1] 2022/03/27 08:32:11.513880 [DBG] [::1]:60536 - cid:49 - "v2.0.0:python3:test" - Client Ping Timer
[1] 2022/03/27 08:32:11.513941 [TRC] [::1]:60536 - cid:49 - "v2.0.0:python3:test" - ->> [PING]
[...]

PR #642, opened 2024-12-13
Don't break internal loops on CancelledError when the cancel is not triggered internally (avoid CTRL-C issue on python < 3.11)

diorcety:
Fix for #287
The issue comes from handling of before 3.11 (https://github.com/python/cpython/commit/f08a191882f75bb79d42a49039892105b2212fb9)
Before 3.10 all the tasks are cancelled when a KeyboardInterrupt is generated (https://github.com/python/cpython/blob/3.10/Lib/asyncio/runners.py#L47), leading that all internal loops are also cancelled, for example the client reader, flusher loops. Any call to nats client after a CTRL-PY will certainly timeout

In order to avoid such issue, this commit change the behaviour of CancelError inside the loop if this one is not due to internal cancel call.
The call to unsubscribe on PushSubscription is changed. Subscription loop is looking for is_closed from Subscription object, calling unsubscribe on super object of PushSubscription doesn't set correctly the flag _closed on the underlying Subscription.

I found during this work that the subscription inside _init_request_sub is never unsubscribed correctly.


nats/aio/subscription.py
```diff
@@ -333,7 +333,10 @@ async def _wait_for_msgs(self, error_cb) -> None:
                         and self._pending_queue.empty):
                     self._stop_processing()
             except asyncio.CancelledError:
-                break
+                if self._closed:
+                    break
+                else:
+                    continue
```
stankudrow, line 339: A redundant `else` branch.

nats/aio/client.py
```diff
@@ -2130,7 +2143,10 @@ async def _read_loop(self) -> None:
                 await self._process_op_err(e)
                 break
             except asyncio.CancelledError:
-                break
+                if self._status == Client.CLOSED:
+                    break
+                else:
+                    continue
```
stankudrow, line 2149: A redundant `else` branch.

nats/aio/client.py
```diff
@@ -2098,8 +2106,13 @@ async def _ping_interval(self) -> None:
...
                     return
                 await self._send_ping()
-            except (asyncio.CancelledError, RuntimeError, AttributeError):
+            except (RuntimeError, AttributeError):
                 break
+            except asyncio.CancelledError:
+                if self._status == Client.CLOSED:
+                    break
+                else:
+                    continue
```
stankudrow, line 2115: A redundant `else` branch.

nats/aio/client.py
```diff
@@ -2098,8 +2106,13 @@ async def _ping_interval(self) -> None:
...
             # except asyncio.InvalidStateError:
             #     pass
```
stankudrow, line 2117: To remove.

nats/aio/client.py
```diff
@@ -2068,23 +2071,28 @@ async def _flusher(self) -> None:
...
+                except (RuntimeError, AttributeError):
+                    # RuntimeError in case the event loop is closed
+                    break
+                finally:
+                    future.set_result(None)
+            except asyncio.CancelledError:
+                if self._status == Client.CLOSED:
+                    break
+                else:
+                    continue
```
stankudrow, line 2095: A redundant `else` branch.

diorcety: Fixed, thanks

nats/aio/client.py
diorcety: Fixed, thanks

nats/aio/client.py
diorcety: Fixed, thanks

nats/aio/subscription.py
diorcety: Fixed, thanks

PR #649, opened 2025-01-17
Add note on updating docs

wallyqs:


wallyqs:
@alexbozhenko maybe we could add a badge for the result of the GitHub actions?

PR #665, opened 2025-02-15
Update Makefile

stankudrow:


Makefile
```diff
@@ -1,7 +1,10 @@
 REPO_OWNER=nats-io
 PROJECT_NAME=nats.py
+
 SOURCE_CODE=nats
+TEST_CODE=tests
 
+.PHONY: help clean deps format lint test ci watch
```
stankudrow, line 7: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Makefile
```diff
@@ -10,33 +13,25 @@ help:
...
-	yapf -i --recursive $(SOURCE_CODE)
-	yapf -i --recursive tests
+	yapf -i --recursive $(SOURCE_CODE) $(TEST_CODE)
 
+lint:
+	yapf --recursive --diff $(SOURCE_CODE) $(TEST_CODE)
+	mypy $(SOURCE_CODE)
+	flake8 $(SOURCE_CODE)
 
 test:
```
stankudrow, line 31: I think the `lint` rule may not depend on the `lint` rule.

stankudrow:
@wallyqs , @caspervonb , do you really need the `ci` and `watch` rules?

stankudrow:
> Main intent is making them phony targets?

The key goals were actually two:

- yes, make these rules phony;
- introduce the "lint" rule (IMHO, it is useful);
- and, supplementary, to reconsider the file content.

Makefile
```diff
@@ -10,33 +13,25 @@ help:
...
-	yapf --recursive --diff $(SOURCE_CODE)
-	yapf --recursive --diff tests
-	mypy
-	flake8 ./nats/js/
 	pytest
 
-
 ci: deps
```
stankudrow, line 34: No dependency on either the "lint" rule and the "test" one because of different corresponding commands.

stankudrow:
In spite of these [checks](https://github.com/nats-io/nats.py/blob/main/.github/workflows/check.yml#L28-L33) in the "check.yml" workflow file, would it be useful to gather these checks under the "check" rule and just call `make check`?

stankudrow:
Hm, in another PR about grooming the workflow files, this [part](https://github.com/nats-io/nats.py/blob/main/.github/workflows/test.yml#L35-L37) in the "test.yml" can be replaced with the `make ci` command, so it is good this rule is already defined.

PR #666, opened 2025-02-17
Add `updates_only` mode to object store watch

caspervonb:
Alternative approach to #659


wallyqs (approved):
LGTM, needs some format fixes it seems

PR #672, opened 2025-03-09
Add an is_acked property to nats.aio.msg.Msg

charles-dyfis-net:
Provides a means to check whether a message has already been sent an ack without directly accessing nonpublic interfaces.


wallyqs (approved):
LGTM

PR #683, opened 2025-04-06
Resolve #682

### LINKED ISSUE #682: Add feature from nats 2.11
### Proposed change

Need add new feature from 2.11 version

### Use case

In Faststream we need to add support nats 2.11, but nats-py does not provide support

### Contribution

Yeap, I can take this on, but I'm going to need a little help with some issues.

Flosckow:
Add features from nats 2.11.

- [ ] Per-message ttl
- [ ] Pull consumer group

### DIFF
--- a/nats/js/api.py
+++ b/nats/js/api.py
@@ -13,7 +13,6 @@
 #
 
 from __future__ import annotations
-
 from dataclasses import dataclass, fields, replace
 from enum import Enum
 from typing import Any, Dict, Iterable, Iterator, List, Optional, TypeVar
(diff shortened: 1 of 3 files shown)


Flosckow:
@wallyqs, hi, can you check it?

Flosckow:
@Jarema, hi, сan you please help me with this?

Flosckow:
I wanted to clarify something about the pinned client. The ADR says that it can only be used for consume, not fetching. I'm a bit confused. Do I understand correctly that "consume" actually means "subscribe" and therefore, I need to add the "cb" parameter to "pull_subscribe" method?

Flosckow:
Can we add per message ttl and pull consumer group with overflow? Pinned client in work, but I don't have much time right now

Flosckow:
@wallyqs, sorry for the interruption, but I would like to finish this task.

oliverlambson:
Is there anything I could do to help move this along?

PR #705, opened 2025-06-27
fix streamsource serialization

cezarguimaraes:
The stream create API call currently silently fails to configure subject transforms when creating a sourcing stream. The nats server does warn about unknown field "subject_transform"

subject_transforms in go sdk with the correct json serialization key: https://github.com/nats-io/nats.go/blob/main/jsm.go#L282


wallyqs (approved):
LGTM

Issue #710, opened 2025-07-10
Unable to create jetstream consumer if filter_subjects is specified


nickolaj-jepsen:
### Observed behavior

When trying to use `filter_subjects` (plural), nats.py throws an error claiming that a consumer can't have both FilterSubject and FilterSubjects

```python
BadRequestError: nats: BadRequestError: code=400 err_code=10136 description='consumer cannot haveboth FilterSubject and FilterSubjects specified'
```

### Expected behavior

I'd expect this not to crash, as i don't provide a specific `filter_subject`, only `filter_subjects`

### Server and client version

Server: 2.11.6
nats-py: 2.10.0

### Steps to reproduce

Here's a minimal repro

```python
import nats
import asyncio
from nats.js.api import ConsumerConfig

async def main() -> None:
    connection = await nats.connect("localhost:4222")
    await connection.jetstream().add_stream(
        name="test_stream",
        subjects=["test.>"],
    )

    # Works
    await connection.jetstream().subscribe(
        subject="test.>",
        config=ConsumerConfig(
            filter_subject="test.foo",
        ),
    )

    # Crashes with "nats.js.errors.BadRequestError: nats: BadRequestError: code=400 err_code=10136 description='consumer cannot have both FilterSubject and FilterSubjects specified'"
    await connection.jetstream().subscribe(
        subject="test.>",
        config=ConsumerConfig(
            filter_subjects=["test.foo", "test.bar"],
        ),
    )

asyncio.run(main())
```

PR #711, opened 2025-07-10
Fix filter_subject overriding filter_subjects

nickolaj-jepsen:
Avoids setting filter_subject if filter_subjects is set, this avoids the server throwing this error when creating a consumer: "consumer cannot have both FilterSubject and FilterSubjects specified"

fixes: #710

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -48,3 +48,45 @@ jobs:
 
             - name: Run lint check
               run: uv run ruff check
+
+    spell:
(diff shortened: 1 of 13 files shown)


nickolaj-jepsen:
Hi @caspervonb, thanks for the review, I've added a test case for both the push and pull based subscriber. Let me know if you want any additional tests :slightly_smiling_face:

wallyqs (approved):
LGTM

PR #716, opened 2025-07-28
Claims JWT Support

arnavdas88:
Added Pythonic Interface for creating `Account Claims` and `User claims` as JWT. 

**Issue:** [Issue 717](https://github.com/nats-io/nats.py/issues/717)

Replicated the example at [NATS by Example - Programmatic NKeys and JWTs](https://natsbyexample.com/examples/auth/nkeys-jwts/go) in the test file [tests/contrib/test_jwt.py](https://github.com/arnavdas88/nats.py/blob/d4d648a92fa991543fc504e10a88cd8825f011bb/tests/contrib/test_jwt.py).

### Example Code Snippet
```python
operatorKP = from_seed(b'SOALU7LPGJK2BDF7IHD7UZT6ZM23UMKYLGJLNN35QJSUI5BNR4DJRFH4R4')
accountKP = from_seed(b'SAALXUEDN2QR5KZDDSH5S4RIWAZDM7CVDG5HNJI2HS5LBVYFTLAQCOXZAU')
userKP = from_seed(b'SUALJTG5JNRQCQKFE652DV4XID522ALOHJNQVHKKDJNVGWHCLHOEXEROEM')

account_claims = Claims(
    name="my-account",
    jti="PBFES33GGIFZM6UGC7NY5ARHRBFVFU4UD7FS2WNLZH3KPGWFVEFQ",
    iat=1678973945,
    iss=operatorKP.public_key.decode(),
    sub=accountKP.public_key.decode(),
    nats=Account(
        limits=OperatorLimits(
            nats_limits=NatsLimits(data = -1, payload = -1, subs = -1),
            account_limits=AccountLimits(exports = -1, imports = -1, wildcards = True, conn = -1, leaf = -1),
            jetstream_limits=JetStreamLimits(disk_storage=-1, mem_storage=-1)
        ),
        default_permissions=Permissions(),
        generic_fields=GenericFields(version=2, type=Types.Account)
    )
)

account_claims.to_dict()
```

### Some issues with the PR
- [ ] Directory and File structure is a mess in the `contrib/`.
- [ ] Test for Operator Claim is not written yet as the operator claims are not clear in [NATS by Example - Programmatic NKeys and JWTs](https://natsbyexample.com/examples/auth/nkeys-jwts/go) blog.
- [ ] `encode_user_claim()` and `encode_account_claim()` is not yet mature and only exists in the test file.
- [ ] The tests pass with the example shown in [NATS by Example - Programmatic NKeys and JWTs](https://natsbyexample.com/examples/auth/nkeys-jwts/go), but the resultant jwt has not been tested with a nats-server.
- [ ] Adding code documentation can be helpful.

> **Note :** Directory and File structure is a mess in the `contrib/` because following any other file structure was resulting in circular import error, and putting all the code in a single file was making it hard to read. Any suggestions / commits on how to have a better structure for the files in `contrib/` directory are welcome!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
             fail-fast: false
             matrix:
                 python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
-                nats_version: ["v2.10.29", "v2.11.6", "main"]
+                nats_version: ["v2.10.29", "v2.11.8", "main"]
                 include:
                     - nats_version: "main"
                       continue-on-error: true
--- a/tests/test_js.py
+++ b/tests/test_js.py
@@ -4057,12 +4057,10 @@ class ObjectStoreTest(SingleJetStreamServerTestCase):
             "TEST_LIST",
             config=nats.js.api.ObjectStoreConfig(description="listing", ),
         )
-        await asyncio.gather(
-            obs.put("A", b"AAA"),
-            obs.put("B", b"BBB"),
-            obs.put("C", b"CCC"),
-            obs.put("D", b"DDD"),
-        )
+        await obs.put("A", b"AAA")
+        await obs.put("B", b"BBB")
+        await obs.put("C", b"CCC")
+        await obs.put("D", b"DDD")
         entries = await obs.list()
         assert len(entries) == 4
         assert entries[0].name == "A"


alexbozhenko:
The test was fixed in https://github.com/nats-io/nats.py/pull/728. Could you please rebase?

arnavdas88:
@caspervonb 

1.) The Flattening Model is indeed over complicated (I have already written that in the PR). Also, it cannot load a dict / json right now. It can only convert the dataclass to a dict.

I have an alternate implementation that makes it a little bit simpler, while adding both serialization and serialization. But it will still add some complications. 


2.) There are Redundant Constructors because I took a lot of reference code from the golang package, as well as some code (for the constants and the jwt encoding and decoding) from the javascript/typescript package. 

For the Redundant Code as well as the Flattening Model, I am trying to keep the code format and classes as consistent as possible with the golang code. 


3.) About the inconsistent dataclass approach, dataclasses does make it easier for us to develop, but when inheriting multiple dataclasses, the classes will not have the nested structure (that is possible with golang's struct), and instead become a flat class (all the attributes are flat. I wanted to avoid that. 

Currently the classes that do not need to inherit any other model (source reference from go package) is replaced by dataclasses, whereas any model that inherits other classes use the `__init__`.


4.) utils.py is indeed a complete copy paste of the `asdict()` function with the additional `omitempty`. The other option was to dynamically injection this behaviour. Making it a bad approach. As I said, I have an alternative approach. I want to keep the asdict as similar as possible to the original function.


5.) Empty classes for claims - They *are* mostly used as placeholder right now, as most of the behaviour for the claims is a bit confusing to me! Specially the OperatorClaims. The placeholder can be modified by someone more experienced in the claims part.


6.) Manual JWTs were constructed because that part of the code was referenced from the typescript/javascript code where the jwt were manually constructed. It was necessary as the jwt the algorithm used in the jwt for claims encoding doesn't usually fall under the standard algorithms, and may or may not raise issues. Hence I kept those functions custom.

7.) Constants doesn't only not pass lint, most of the constants are not even used. But they exist and may help to fetch specific data from NATS server. So I kept all of them! The constants exist in js/ts package but not in the python package. I wonder why!


Would love to know your opinion on these!

Issue #490, opened 2023-08-25
_read_loop() inside nats/aio/client.py could become a never yield forever loop (nats-py-2.3.1.tar.gz, ssl)


3figs:
### What version were you using?

nats-py-2.3.1.tar.gz

### What environment was the server running in?

alpine linux - 3.18.2

### Is this defect reproducible?

Yes, 1 out 4 of changing certs.  after nats-server  restarts, client can fall into forever loop.  

### Given the capability you are leveraging, describe your expectation?

 Under no circumstance would _read_loop fall into a non-yield forever loop.  It blocked everything.

### Given the expectation, what is the defect you are observing?

Hi,
Our ssl nats client could fall into a forever loop when certs changed (nat-server restarted). Eventually we located the forever loop was inside _read_loop() inside aio/client/py:

   ```
while True:
      try:
           should_bail = self.is_closed or self.is_reconnecting
           if should_bail or self._transport is None:
                  break
          if self.is_connected and self._transport.at_eof():
                 err = errors.UnexpectedEOF()
                 await self._error_cb(err)
                 await self._process_op_err(err)
                 break
          b = await self._transport.read(DEFAULT_BUFFER_SIZE) <-----
          await self._ps.parse(b)
```

 b is always empty, eof is true.   When this issue happened, all other asyncio tasks were blocked because this loop never yield.
 
 We hit this issue when certs changes . After certs changed, nats-server restarts first. The nats-client using old certs should detect the failure and eventually reconnect using new certs. But nats-client falls into a forever loop and blocks other tasks from running.  This issue can be seen around 1 out 4 tries.
   
Have anyone ever hit similar issue? It seems a bug to me.  Under no circumstance would _read_loop fall into a non-yield forever loop. 

 
 Thanks
 Steven

3figs:
Forget about certs change. I can make it happen easily. It can happen anytime if the criteria is met.
 Actually I did see once when my app just restarted (no nats server restart).
 
When self._status == CONNECTING, self._transport.at_eof(), and b = await self._transport.read(..) is empty, _read_loop will fall into a never-yield-forever-loop.  Once this happened, it would block all other tasks from running. 

self._status can be CONNECTING at connect or attempt_reconnect().  It depends on when transport.at_eof() happens.

Below is the change I used. It just forces to set_eof in asyncio/sslproto.py. Remember to run "echo>/var/log/robot/set_eof"  before restart your app.


[...]

PR #719, opened 2025-08-06
Fix EOF processing while client is connecting.

ksenia-vazhdaeva:
Resolves https://github.com/nats-io/nats.py/issues/490

When the client status (`self._status`) is `CONNECTING` and the transport is eof (`self._transport.at_eof()`), the reading from the transport hangs indefinitely  (`b = await self._transport.read(DEFAULT_BUFFER_SIZE)` in `_read_loop`).

The solution is to process transport eof in the same way it is done when the client is connected.

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -1445,10 +1445,10 @@ class Client:
         try to switch the server to which it is currently connected
         otherwise it will disconnect.
         """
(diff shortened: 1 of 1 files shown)


caspervonb (approved):
Okay, turns out to be very hard to replicate, taking out my block on tests.
Can do it with a ton of mocks, after many runs, it does happen.

Seems sound, PTAL @wallyqs

PR #722, opened 2025-08-14
Add options to send custom WebSocket headers on connect 

saurabhojha:
This pull request introduces support for custom headers in WebSocket connections to the NATS server. The key changes include:

Provision for supplying custom WebSocket request headers — developers can now attach additional HTTP headers (e.g., authorization tokens or tracing metadata) when establishing a WebSocket connection via NATS py.

A reference to a related issue in the .NET client repository (https://github.com/nats-io/nats.net/issues/305) highlights the broader context and motivation for this enhancement.

Usage:
You can pass headers as a nats connection option ws_connection_headers.
```py
nc = await nats.connect(
            "wss://localhost:8081",
            ws_connection_headers={
            "Authorization": ["Bearer SecureToken"],
            "X-TLS-Client": ["secure-client-v1"]
             }
        )
```

### DIFF
--- a/nats/aio/client.py
+++ b/nats/aio/client.py
@@ -358,7 +358,7 @@ class Client:
         inbox_prefix: Union[str, bytes] = DEFAULT_INBOX_PREFIX,
         pending_size: int = DEFAULT_PENDING_SIZE,
         flush_timeout: Optional[float] = None,
-        ws_connection_headers: Optional[Dict[str,List[str]]] = None,
+        ws_connection_headers: Optional[Dict[str, List[str]]] = None,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


nats/aio/client.py
```diff
@@ -33,6 +33,8 @@
 from typing import Any, Awaitable, Callable, Dict, List, Optional, Tuple, Union
 from urllib.parse import ParseResult, urlparse
 
+from requests import options
```
caspervonb: CI is failing because of this import, needs to be added as a dependency for it to pass. 

Is this even used?

saurabhojha: I have removed this.  Thanks for pointing this out.

PR #726, opened 2025-09-03
Test more recent nats server versions

alexbozhenko:


.github/workflows/test.yml
```diff
@@ -18,7 +18,14 @@ jobs:
             fail-fast: false
             matrix:
                 python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
-                nats_version: ["v2.10.29", "v2.11.6", "main"]
+                nats_version:
+                    [
+                        "v2.10.29",
+                        "v2.11.8",
+                        "v2.11.9-RC.2",
```
wallyqs: can you change into release/v2.11.9 instead?

.github/workflows/test.yml
```diff
@@ -18,7 +18,14 @@ jobs:
...
+                        "v2.12.0-preview.2",
```
wallyqs, line 26: also punt on this one and let's just keep main

alexbozhenko: Let me switch back to draft. I am just trying to bisect to find out which server version stated to fail

PR #731, opened 2025-09-22
Add nats-server package

caspervonb:
This adds a nats.server package which provides utilities for spawning nats server processes.


.github/workflows/test.yml
```diff
@@ -42,6 +42,43 @@ jobs:
...
+                  pipenv run pytest -x -vv -s --continue-on-collection-errors ./tests
               env:
                   PATH: $HOME/nats-server:$PATH
+
+    project:
+        name: ${{ matrix.project }} (python-${{ matrix.python-version }}, nats-server-${{ matrix.nats-server-version }}, ${{ matrix.os }})
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
+                python-version: ["3.11", "3.12", "3.13"]
```
philpennock, line 54: Can we get a comment saying when this list should be updated and what the criteria are?

.github/workflows/test.yml
```diff
@@ -42,6 +42,43 @@ jobs:
...
+                os: ["ubuntu-latest", "macos-latest"]
+                nats-server-version: ["latest"]
+                project: ["nats-server"]
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v2
```
philpennock: Current version is `@v5`; v2 used Node 12, and Node 16 is currently being deprecated by GH.  I'm impressed this even works.

.github/workflows/test.yml
```diff
@@ -42,6 +42,43 @@ jobs:
...
+
+            - name: Set up Go
+              uses: actions/setup-go@v2
```
philpennock: Use `@v6` for: `actions/setup-go@v6`; v2 was node12.

.github/workflows/test.yml
```diff
@@ -42,6 +42,43 @@ jobs:
...
+              with:
+                  go-version: "1.23.4"
```
philpennock: Go 1.23 is no longer supported.  Since we're not using this to cut releases, we probably don't care about explicitly pinning to specific releases, so either use `stable` or have a comment indicating what environment we're trying to match.  If this needs to be older Go for a variety of nats-server versions, then `oldstable`?

.github/workflows/test.yml
```diff
@@ -42,6 +42,43 @@ jobs:
...
+
+            - name: Install NATS Server
+              run: |
+                  go install github.com/nats-io/nats-server/v2@${{ matrix.nats-server-version }}
+              shell: bash
+
+            - name: Set up Python ${{ matrix.python-version }}
+              uses: actions/setup-python@v2
```
philpennock: Python: `@v6`; v2 was node12.

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import re
+
+from typing_extensions import Self
```
philpennock: I take it this syntax for making `Self` available is because this code needs to run under Python 3.11 so this is backwards-compatibility?

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+
+# Set up logging
+logger = logging.getLogger(__name__)
+
+# Regex patterns for parsing server output
+READY_PATTERN = re.compile(r"Server is ready")
+LISTENING_PATTERN = re.compile(
+    r"Listening for client connections on ([^:]+):(\d+)"
```
philpennock: This looks like a bomb waiting for running on a node with IPv6 enabled.
Note that `nats-server -a ::` to listen on wildcard IPv6 results in output of:
```
Listening for client connections on [::]:4222
```

So this should probably use something like:
```python
LISTENING_PATTERN = re.compile(
    r"Listening for client connections on \[?([0-9a-f.:]+)\]?:(\d+)\b"
    )
```

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+)
+FATAL_PATTERN = re.compile(r"\[FTL\]\s+(.*)")
+INFO_PATTERN = re.compile(r"INFO\s+({.*})")
```
philpennock: This use of braces like this doesn't seem valid to me.  Also, the lines would be `\[INF\]` surely?  It doesn't look like `INFO_PATTERN` is used, so is this stale cruft?

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+class ServerError(Exception):
+    """Raised when a NATS server encounters an error during startup or operation."""
+
+
+class Server:
+    """Process controller for a NATS server instance."""
+
```
philpennock, line 85: declaring the self fields and their types here would probably help with type-checking, ensuring that it's more readily available than inference via __init__ types.

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+    @property
+    def client_url(self) -> str:
+        """Get the client url for the server, this can be used to connect to the server.
+
+        Returns:
+            The connect URL as a string.
+        """
+        host = self._host
+        if host == "0.0.0.0":
```
philpennock: Any particular reason for forcing `localhost` here, particularly when that might resolve back to `::` instead of `0.0.0.0`?  If so, perhaps:

```python
if host in ("0.0.0.0", "::"):
```

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+            port: The port the server is listening on.
+            jetstream: Whether JetStream is enabled.
+            connect_urls: List of additional server URLs that can be used for connection.
+        """
+        self._process = process
+        self._host = host
+        self._port = port
+        self._jetstream = jetstream
+        self._connect_urls = connect_urls or []
+        self._is_running = True
```
philpennock: This seems off: the property `is_running` is looking at entirely different fields.  Is this `_is_running` field cruft?

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+        """Get the client url for the server, this can be used to connect to the server.
+
+        Returns:
+            The connect URL as a string.
+        """
+        host = self._host
+        if host == "0.0.0.0":
+            host = "localhost"
+
+        return f"nats://{host}:{self._port}"
```
philpennock, line 137: That's ... making assumptions about `host`.  A URL library should probably be used instead of string concat, to get this right.  But minimally if sticking to this, then something like this immediately after the localhost transform:

```python
if ":" in host:
    host = "[" + host + "]"
```

nats-server/src/nats/server/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+        while True:
+            line = await process.stderr.readline()
+            if not line:
+                # EOF reached, process stderr for errors
+                await process.wait()
+
+                # If the process exited with an error code and we have stderr content
+                if process.returncode != 0:
+                    if error_lines:
+                        raise ServerError("\n".join(error_lines))
```
philpennock: This is losing the return-code, so that we only get the return-code information if there's nothing on stderr.  That seems unfortunate.

nats-server/tests/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+particularly for verifying that the required nats-server executable is available.
+"""
+
+import socket
+import subprocess
+
+import pytest
+
+
+def get_nats_server_version():
```
philpennock, line 14: Can this get a typing hint for the return type, `() -> str:` ?

nats-server/tests/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    Verify nats-server is installed and available before running any tests.
+
+    This fixture runs automatically at the start of the test session and
+    fails fast if the nats-server executable cannot be found.
+    """
+    get_nats_server_version()
+
+
+@pytest.fixture
+def available_port():
```
philpennock: Can this too get a typing hint?  I think it's returning int, not str?

nats-server/tests/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Get a free TCP port that can be used for testing.
```
philpennock, line 41: The name of the function doesn't say "a port which is available on localhost", but that's where the test is being done.  A port could be bound for a public IP but available for localhost, and we would randomly return that.  This should probably lead to the function being renamed to `available_port_for_localhost` or somesuch.

nats-server/pyproject.toml
```diff
@@ -0,0 +1,58 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "nats-server"
+version = "0.0.0"
+description = 'Python library for managing NATS servers for development and testing'
+readme = "README.md"
+requires-python = ">=3.8"
```
philpennock: Where is the 3.8 value coming from?  We're not testing that far back in the GH action.

nats-server/pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+  { name = "Casper Beyer", email = "casper@synadia.com" },
+]
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
```
philpennock, line 20: This is missing 3.13, which we do test against.

nats-server/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Asynchronous server control
+
+## Installation
+
+```bash
+pip install nats-server
+```
+
+Requires Python 3.8+ and the NATS server binary in your PATH.
```
philpennock: If we're asserting 3.8 compatibility then shouldn't our GH Actions test suite include that in the matrix?  We're only testing back to 3.11.

caspervonb: This should be bumped to 3.11

nats-server/pyproject.toml
caspervonb: Yes, stale, bumped to 3.11

.github/workflows/test.yml
caspervonb: Pinned it to "stable" instead.

nats-server/tests/conftest.py
caspervonb: Done

nats-server/tests/conftest.py
caspervonb: Done

nats-server/src/nats/server/__init__.py
caspervonb: Resolving to ::1 and 127.0.0.1 now instead.

nats-server/src/nats/server/__init__.py
caspervonb: Yeah stale, removed.

nats-server/src/nats/server/__init__.py
caspervonb: Fixed, merged "FTL" parsing/normalization into it as-well.

nats-server/tests/conftest.py
caspervonb: Done.

nats-server/src/nats/server/__init__.py
caspervonb: Done, did a typed dict of server info as-well/instead.

nats-server/src/nats/server/__init__.py
caspervonb: Was for when it was 3.8 ish, but just importing from typing now.

.github/workflows/test.yml
caspervonb: Bumped them.

.github/workflows/test.yml
caspervonb: Bumped them.

.github/workflows/test.yml
caspervonb: Bumped them.

PR #732, opened 2025-09-22
Add nats-client package

caspervonb:
This adds a standalone nats.client package

Main differences in `nats.client` compared to `nats.aio`

- Asynchronous first
- Substantially faster than nats.aio
- Proper headers, with support for multi-values.
- JetStream unaware as a feature.


nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+    MSG = "MSG"
+    HMSG = "HMSG"
+    PING = "PING"
+    PONG = "PONG"
+    INFO = "INFO"
+    ERR = "ERR"
+
+
+# Message type aliases
+MsgResult = tuple[Literal["MSG"], str, str, str | None,
```
philpennock: Was `typing.NamedTuple` considered as an option here?

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+    try:
+        parts = line.decode().split()
+        if not parts:
+            msg = "Empty control line"
+            raise ParseError(msg)
+
+        try:
+            op = Op(parts[0])
+        except ValueError as e:
+            msg = f"Unknown operation: {parts[0]}"
```
philpennock: Given that there could be control characters in that, perhaps `{parts[0]!r}` ?

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+        raise ParseError(msg)
+
+    try:
+        parts = line.decode().split()
+        if not parts:
+            msg = "Empty control line"
+            raise ParseError(msg)
+
+        try:
+            op = Op(parts[0])
```
philpennock: <https://docs.nats.io/reference/reference-protocols/nats-protocol> says:
> NATS protocol operation names are case insensitive, thus SUB foo 1␍␊ and sub foo 1␍␊ are equivalent.

So I think this needs to be: `op = Op(parts[0].upper())`

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+        status_code = status_parts[1]
+
+        # If there's a description part, extract it
+        if len(status_parts) >= MIN_STATUS_PARTS_WITH_DESC:
+            status_description = status_parts[2]
+
+    # Parse header key-value pairs
+    for line in lines[1:]:
+        if not line or ":" not in line:
+            continue
```
philpennock, line 147: This tickles my spidey-sense. `not line` I can see, as the last item in the data because the header_data including final `CRLFCRLF` blank line is included, though that should trigger an issue if it's _not_ the last one.  But silently ignoring headers missing `:` seems like a way for badness to silently slip through.

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+            parts = control_line.split(b" ")
+            op = parts[0]  # Keep as bytes
+            args = parts[1:]  # Keep as bytes
+
+        except Exception as e:
+            msg = f"Invalid control line: {e}"
+            raise ParseError(msg) from e
+
+        # Handle different operations
+        if op == b"MSG":
```
philpennock: Again, protocol messages are case-insensitive.

Similarly for other branches below.

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+            sid_bytes = args[1]
+
+            if len(args) == MIN_MSG_ARGS:
+                # No reply subject
+                reply_to_bytes = None
+                size = int(args[2])
+            else:
+                # With reply subject
+                reply_to_bytes = args[2]
+                size = int(args[3])
```
philpennock: Nothing here is checking the _max_ number of MSG args, and if the intent is to allow for future extensiion, shouldn't this be using `size = int(args[-1])` ?

Similarly for other types below.

nats-client/src/nats/client/protocol/types.py
```diff
@@ -0,0 +1,80 @@
...
+    tls_required: NotRequired[bool]  # If this is set, client must use TLS
+    tls_verify: NotRequired[
+        bool]  # If this is set, client must use TLS with valid cert
+    tls_available: NotRequired[
+        bool
+    ]  # If this is true, client can provide valid cert during TLS handshake
+    connect_urls: NotRequired[
+        list[str]]  # List of server URLs available for client to connect
+    ws_connect_urls: NotRequired[list[str]]  # List of websocket server URLs
+    ldm: NotRequired[bool]  # If set, server supports limited data mode
```
philpennock: ??
"Lame Duck Mode", not "Limited Data Mode".

nats-client/src/nats/client/connection.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @abstractmethod
+    async def write(self, data: bytes) -> None:
+        """Write data to the connection."""
+
+    @abstractmethod
+    def is_connected(self) -> bool:
+        """Check if the connection is active."""
+
+    async def readline(self) -> bytes:
```
philpennock, line 40: Why are these last two _not_ `@abstractmethod` and raising `NotImplementedError`?  Are there legitimate sub-classes which don't implement these?  If so, could we get a comment explaining so, please?

nats-client/src/nats/client/errors.py
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__(f"{status}: {description}")
+
+    @classmethod
+    def from_status(
+        cls,
+        status: str,
+        description: str,
+        *,
+        subject: str | None = None
+    ) -> StatusError:
```
philpennock: Can we get a comment re "Python 3.10 compatibility" to explain why this isn't `typing.Self` and we're accepting the broken typing of subclasses as an ease-of-coding issue, while still supporting 3.10?

nats-client/src/nats/client/message.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Headers):
+            return NotImplemented
+        return self._headers == other._headers
+
+
+@dataclass
+class Status:
+    """NATS message status information.
```
philpennock, line 114: Which generic NATS message status is using 3-digit codes _and_ allows for success codes?  The wire protocol is using `+OK`/`-ERR`.  The JS API is using 5-digit codes.  ADR-2 has 3-digit codes but specifies them as starting at 300 because they're _always_ for errors.

So I'm confused about the target usage of this class where `.is_error` might ever be false.

If this code is correct, then it needs a little more detail in the docstring.

nats-client/src/nats/client/message.py
```diff
@@ -0,0 +1,129 @@
...
+    def has_status(self) -> bool:
+        """Check if this message has a NATS status.
+
+        Returns:
+            True if the message has status information
+        """
+        return self.status is not None
+
+    @property
+    def is_error_status(self) -> bool:
```
philpennock: Similarly here: what NATS status exists which is _not_ an error and where is it specified, please?

nats-client/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[project]
+name = "nats-client"
+version = "0.0.0"
+description = "NATS client implementation in Python"
+authors = [{ name = "Casper Beyer", email = "casper@synadia.com" }]
+dependencies = []
+requires-python = ">=3.10"
+readme = "README.md"
+license = { text = "MIT" }
```
philpennock: nats.py is under Apache 2 license

nats-client/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
+## 🚀 Performance
+
+This client implementation delivers significant performance improvements over the nats.aio client, particularly for high-frequency, small message workloads.
+
+Do note tho, it is not as feature complete at this point in time.
```
philpennock: `tho` -> `though`

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+
+    Args:
+        server_info: Server information
+        no_randomize: Whether to disable randomizing the server pool
+
+    Returns:
+        List of server addresses
+    """
+    # Start with current server
+    servers = [f"{server_info.host}:{server_info.port}"]
```
philpennock: Another one which will break on IPv6, I think?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+                    msg = await parse(self._connection)
+
+                    if not msg:
+                        logger.info("Connection closed by server")
+                        break
+
+                    # Handle message based on type
+                    match msg:
+                        case ("MSG", subject, sid, reply_to, payload):
+                            logger.debug(
```
philpennock: If the goal is higher performance, have you compared to guarding this with the `isEnabledFor()` method (perhaps catched from before entering the `while True` loop) to not do the string building options for debugging if we're not going to log debug messages?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+        """Handle ERR from server."""
+        self._last_error = error
+
+        # Call error callback if set
+        if self._error_callbacks:
+            for callback in self._error_callbacks:
+                try:
+                    callback(error)
+                except Exception:
+                    logger.exception("Error in error callback")
```
philpennock: I understand that `logger.exception` will include the callback stack, so the new Exception will be indicated, but it might be worth also explicitly including the error which triggered the exception, unless we're 100% certain that the exceptions will always wrap that information.

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+            self._write_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError, RuntimeError):
+                await self._write_task
+
+        await self._connection.close()
+
+        # Only attempt to reconnect if:
+        # 1. We were not explicitly closing
+        # 2. Reconnect is enabled
+        # 3. We're not already reconnecting
```
philpennock: So if using threaded python, I think this means that after a `_force_disconnect()` then _if_ only one flow at a time triggered this, and reconnections are enabled, then on return from the function we will be reconnected and can continue as normal, but if something else was concurrently reconnecting, we don't have that guarantee?

Should there be logic to wait on concurrent reconnection before we return, for a more consistent API state transition?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+                        logger.info("Trying to reconnect to %s", server)
+
+                        # Parse server address
+                        if "://" in server:
+                            parsed_url = urlparse(server)
+                            host = parsed_url.hostname
+                            port = parsed_url.port or 4222
+                            scheme = parsed_url.scheme
+                        else:
+                            host, port_str = server.split(":")
```
philpennock: Is it intentional that we support `192.0.2.1:4222` but not `[2001:db8::1]:4222` ?  Should this be an `rsplit` with max-splits and removal of square brackets if present?  (I know, square brackets are a peculiarity of the URL format for IPv6 literals; I honestly don't know what the accepted forms are for _non-URL_ addresses here though.  I strongly suspect that rsplit/with-max is more safer here.)

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+                            port = int(port_str)
+                            scheme = "tls" if self._server_info.tls_required else "nats"
+
+                        try:
+                            # Open new connection based on server info
+                            if scheme in ("tls", "wss"):
```
philpennock: Don't we also have some newer scheme for nats-with-tls-on-connect?

This logic looks like it only supports TLS-on-connect, for `tls:` scheme, not tls-proto-upgrade (the older nats-server behavior).  Is this worth calling out as a change in semantics of `tls://` when using this new API?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1158 @@
...
+                                new_server_info.server_id,
+                                new_server_info.version
+                            )
+
+                            # Send CONNECT
+                            connect_info = ConnectInfo(
+                                verbose=False,
+                                pedantic=False,
+                                lang="python",
+                                version="0.1.0",
```
philpennock: This version doesn't seem to match that in the new `nats-client/pyproject.toml` which says `0.0.0`; can something please be done to keep these two versions in sync, and at the least not bury a client version identifier 589 lines deep in the module init file?

caspervonb: Yeah this should be using `importlib.metadata` for sure.

nats-client/src/nats/client/__init__.py
caspervonb: Will add some more coverage for ipv6.

nats-client/src/nats/client/__init__.py
caspervonb: TLS is mostly stubbed here, have a follow up PR (https://github.com/nats-io/experimental-nats.py/pull/14) just trying to keep the merge queue here reasonable.

nats-client/src/nats/client/__init__.py
caspervonb: Added `__version__` that assigns from importlib metadata.

nats-client/src/nats/client/protocol/message.py
caspervonb: Yeah it started out simple, but HMSG etc do have quite a few fields.
Removed the Op enum and added named tuples instead now.

nats-client/src/nats/client/protocol/message.py
caspervonb: .upper'ed

nats-client/src/nats/client/__init__.py
caspervonb: Unlikely to race (there is the boolean flag, and GIL applies) but added a lock for clarity.

nats-client/src/nats/client/__init__.py
caspervonb: Been reworked, added coverage for ipv6

nats-client/src/nats/client/protocol/message.py
caspervonb: Reworked.

nats-client/src/nats/client/protocol/message.py
caspervonb: Yes, reworked (altho nats-server always sends uppercase).

nats-client/src/nats/client/protocol/message.py
caspervonb: Done.

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,436 @@
...
+            except ValueError as e:
+                msg = f"Invalid payload size: {size_str}"
+                raise ParseError(msg) from e
+            return subject, sid, reply_to, size
+        case _:
+            msg = "Invalid MSG: too many arguments"
+            raise ParseError(msg)
+
+
+def parse_hmsg_args(args: list[str]) -> tuple[str, str, str, int, int]:
```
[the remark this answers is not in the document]
caspervonb: Using named tuples now.

nats-client/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+| Message Size | nats.py (python3) | nats.py (pypy3) | experimental-nats.py (python3) | experimental-nats (pypy3) | Performance Gain |
```
wallyqs: are these benchmarks publish / subscribe / requests per sec?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+        Raises:
+            RuntimeError: If the connection is closed
+            TimeoutError: If no response is received within the timeout
+            StatusError: If return_on_error=False and the response contains error status headers
+        """
+        if self._status == ClientStatus.CLOSED:
+            msg = "Connection is closed"
+            raise RuntimeError(msg)
+
+        inbox = f"_INBOX.{uuid.uuid4().hex}"
```
wallyqs: nuids?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+            logger.debug("->> SUB %s %s", subject, sid)
+
+        await self._connection.write(command)
+
+        return subscription
+
+    async def _subscribe(self, subject: str, sid: str, queue_group: str | None) -> asyncio.Queue:
+        """Create a subscription on the server and return the message queue.
+
+        This method is deprecated and maintained for backward compatibility.
```
wallyqs: _subscribe deprecated?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+    ) -> Subscription:
+        """Subscribe to a subject."""
+        if self._status == ClientStatus.CLOSED:
+            msg = "Connection is closed"
+            raise RuntimeError(msg)
+
+        sid = str(self._next_sid)
+        self._next_sid += 1
+
+        message_queue = asyncio.Queue()
```
wallyqs: queue does not have bounds?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+
+    async def _handle_msg(self, subject: str, sid: str, reply_to: str | None, payload: bytes) -> None:
+        """Handle MSG from server."""
+        if sid in self._subscriptions:
+            subscription = self._subscriptions[sid]
+            msg = Message(subject=subject, data=payload, reply_to=reply_to)
+
+            for callback in subscription._callbacks:
+                try:
+                    callback(msg)
```
wallyqs: callback runs in place in the read loop?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+            )
+
+            for callback in subscription._callbacks:
+                try:
+                    callback(msg)
+                except Exception:
+                    logger.exception("Error in subscription callback")
+
+            try:
+                await subscription.queue.put(msg)
```
wallyqs: same with `callback` dispatch, isn't this blocking the read loop? I guess no since it is unbound?

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1076 @@
...
+
+        self._subscriptions[sid] = subscription
+
+        command = encode_sub(subject, sid, queue_group)
+        if queue_group:
+            logger.debug("->> SUB %s %s %s", subject, queue_group, sid)
+        else:
+            logger.debug("->> SUB %s %s", subject, sid)
+
+        await self._connection.write(command)
```
wallyqs, line 999: subs are sync writes?

nats-client/src/nats/client/__init__.py
caspervonb: Yes, only synchronous callbacks allowed. If user wants to perform an asynchronous  task, up to them to spawn one.

nats-client/src/nats/client/message.py
caspervonb: From HMSG, there's can be status and description (e.g NoResponders).

We do the same thing in nats.rs where we consider those two separate from the user headers provided in the header block because they're parsed from different things (e.g headers might be 0 bytes, but there can still be a status).

nats-client/src/nats/client/__init__.py
caspervonb: With an unbound queue, not an issue. With slow consumer wiring it will eventually be put_nowait.

nats-client/src/nats/client/__init__.py
caspervonb: Yeah, no that's documentation help not helping 😅

nats-client/src/nats/client/__init__.py
caspervonb: Currently, PUB and HPUB are buffered.
PING, PONG, SUB are unbuffered.

nats-client/src/nats/client/__init__.py
caspervonb: Yep unbounded atm.

nats-client/src/nats/client/__init__.py
wallyqs: where is client slow consumer handled?

nats-client/src/nats/client/__init__.py
wallyqs: that means single queue can use all memory?

nats-client/src/nats/client/__init__.py
wallyqs: so if you have two subscriptions, one callback processing will cause head of line blocking on the other subscription no? unless you create a task per msg so that it doesn't?

nats-client/src/nats/client/__init__.py
caspervonb: Currently isn't.

nats-client/src/nats/client/__init__.py
caspervonb: We can't await as it introduces a bunch of overhead, will basically stall everything to a halt. If the user chooses to do heavy processing in a callback, then they are responsible for scheduling that in a way that makes sense.

Ideally, you will use async for await, callbacks are for dispatching of messages as they arrive.

nats-client/README.md
caspervonb: Publish / Subscribe.

nats-client/src/nats/client/__init__.py
caspervonb: Yes, we can follow up with that.

nats-client/src/nats/client/__init__.py
wallyqs: I don't think I understand how this is expected to be used, can you try to implement this example?

```python
import asyncio
import time
import nats

async def main():
    nc = await nats.connect("nats://localhost:4222")

    async def cb1(msg):
        print(time.time_ns(), "      [1] --->", msg)
        await asyncio.sleep(5)

    sub2 = await nc.subscribe("foo", cb=cb1)

    async def cb2(msg):
        print(time.time_ns(), "[2] --->", msg)
        await asyncio.sleep(1)

    sub2 = await nc.subscribe("bar", cb=cb2)

    async def task1():
        while True:
            await nc.publish("foo", b'1')
            await asyncio.sleep(1)

    async def task2():
        while True:
            await nc.publish("bar", b'2')
            await asyncio.sleep(0.5)

    t1 = asyncio.create_task(task1())
    t2 = asyncio.create_task(task2())

    await asyncio.sleep(120)
            
    await nc.close()

if __name__ == "__main__":
    asyncio.run(main())
```

I just see a lot of blocking happening using this model, starting point for example:

```python
import asyncio
import time
from nats.client import connect

async def main():
    nc = await connect("nats://localhost:4222")

    def cb1(msg):
        print(time.time_ns(), "     [1] --->", msg)
        # can't async await or do i/o since that blocks the event loop
        time.sleep(5)
        
    sub2 = await nc.subscribe("foo", callback=cb1)

    def cb2(msg):
        print(time.time_ns(), "[2] --->", msg)
        # e.g. do some work
        for i in range(1, 50000000):
            pass

    sub2 = await nc.subscribe("foo", callback=cb2)

    async def task1():
        while True:
            await nc.publish("foo", b'1')
            await asyncio.sleep(1)

    async def task2():
        while True:
            await nc.publish("foo", b'2')
            await asyncio.sleep(0.5)

    t1 = asyncio.create_task(task1())
    t2 = asyncio.create_task(task2())

    await asyncio.sleep(120)
            
    await nc.close()

if __name__ == "__main__":
    asyncio.run(main())
```

caspervonb: Should never do that, use async patterns for consumption.
Same as nats.js, we allow callbacks for observing etc.

I've removed it from client.subsribe, only path is to `.add_callback` on subscription. Same as other observation callbacks.

nats-client/src/nats/client/connection.py
caspervonb: Protocol now, resolved.

nats-client/src/nats/client/__init__.py
wallyqs: @caspervonb I don't think that's true since it is basic nats usage, you can do this in nats.js:

```js
import { connect, deferred, nuid } from "@nats-io/transport-node";

async function main() {
    console.log("Connecting to NATS demo server...");

    // Connect to the public demo server
    const nc = await connect({ servers: "localhost" });
    console.log(`Connected to ${nc.getServer()}`);

    const subj = `foo`;

    let messageCount = 0;
    const d = deferred();

    // Subscribe with callback function
    const sub = nc.subscribe(subj, {
        callback: (err, msg) => {
            messageCount++;
            console.log(`Received[${msg.subject} ${msg.reply}]: ${msg.string()}`);
            if (msg.reply) {
                msg.respond(`Echo: ${msg.string()}`);
            }
        }
    });

    nc.publish(subj, "Hello, World!");
    nc.publish(subj, "Second message");

    try {
        const response = await nc.request("foo", "ping", { timeout: 1000 });
        console.log(`Response: ${response.string()}`);
    } catch (err) {
        console.log(`Error: ${err.message}`);
    }

    await nc.drain();
    console.log("Done!");
}

main().catch(console.error);
```

nats-client/src/nats/client/__init__.py
```diff
@@ -0,0 +1,1071 @@
...
+
+                                if not server.startswith("[") and server.count(":") > 1:
+                                    last_colon = server.rfind(":")
+                                    try:
+                                        port_val = int(server[last_colon + 1 :])
+                                        if 0 <= port_val <= 65535:
+                                            host_part = server[:last_colon]
+                                            server = f"[{host_part}]:{port_val}"
+                                    except ValueError:
+                                        server = f"[{server}]"
```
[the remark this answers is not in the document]
caspervonb: Known server handling / discovery etc needs some rework, but punting for later as this is already plenty large enough.

nats-client/src/nats/client/protocol/message.py
```diff
@@ -0,0 +1,456 @@
...
+    if not args:
+        msg = "ERR message missing error text"
+        raise ParseError(msg)
+
+    # Join the args and decode once
+    error_bytes = b" ".join(args)
+    error_text = error_bytes.decode()
+
+    # Remove quotes if present
+    if error_text.startswith("'") and error_text.endswith("'"):
```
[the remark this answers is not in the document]
caspervonb: Known server handling / discovery etc needs some rework, but punting for later as this is already plenty large enough.

nats-client/src/nats/client/__init__.py
caspervonb: Not quite, let's take a step back here.

The key difference: **JavaScript Promises are hot** by default (calling them starts work immediately). **Python coroutines are cold** by default (calling them does nothing until you await or create_task).
If I allowed async callbacks, you'd get two problems:

- Blocking: `await msg.respond()` makes messages process sequentially which absolutely destroys throughput.
- Overhead: Creating a `Task` for every message wastes a huge amount of cycles better used elsewhere.

Synchronous callbacks do match JavaScript's pattern, avoid blocking, and only pay `Task` overhead when you actually need async work.

The `create_task()` is just Python's explicit way of doing what JavaScript does implicitly - making coroutines hot without blocking.

caspervonb: For reference (https://github.com/nats-io/nats.js/blob/ea522a8ddbc3cddac6014eadc2c5ceedeb371dd3/core/src/protocol.ts#L701-L721):

```typescript
  processMsg(msg: MsgArg, data: Uint8Array) {
    this.inMsgs++;
    this.inBytes += data.length;
    if (!this.subscriptions.sidCounter) {
      return;
    }

    const sub = this.subscriptions.get(msg.sid);
    if (!sub) {
      return;
    }
    sub.received += 1;

    if (sub.callback) {
      sub.callback(null, new MsgImpl(msg, data, this));
    }

    if (sub.max !== undefined && sub.received >= sub.max) {
      sub.unsubscribe();
    }
  }
```

And good measure, @aricart you do not await for user callbacks anywhere do you?

aricart: you cannot - if you await for user callbacks you will get slow consumered - they will block and the client with get disconnected. Callbacks are not expected to wait. If they use an iterator then the iterator buffers for them, and then can await in there.

wallyqs: of course can't block the read loop which is the main concern, thus the approach in nats-py is a task per subscription when there is an async callback then dispatch msgs to them to avoid blocking in your model you would still need a task to avoid this so there is no big difference, it's just not done by the client:

```python
import asyncio
import time
from nats.client import connect

async def main():
    nc = await connect("nats://localhost:4222")
     
    sub1 = await nc.subscribe("foo")
    sub2 = await nc.subscribe("bar")

    async def task1():
        while True:
            await nc.publish("foo", b'1')
            await asyncio.sleep(1)

    async def task2():
        while True:
            await nc.publish("bar", b'2')
            await asyncio.sleep(0.5)

    async def task3():
        async with await nc.subscribe("foo") as sub:
            while True:
                msg = await sub.next()
                print(time.time_ns(), "     [1] --->", msg)
                await asyncio.sleep(1)

    t1 = asyncio.create_task(task1())
    t2 = asyncio.create_task(task2())
    t3 = asyncio.create_task(task3())

    async with await nc.subscribe("bar") as sub:
        while True:
            msg = await sub.next()
            print(time.time_ns(), "[2] --->", msg)
            await asyncio.sleep(10)

    await asyncio.sleep(120)
            
    await nc.close()

if __name__ == "__main__":
    asyncio.run(main())

Comment (caspervonb) [reply nats-client/src/nats/client/__init__.py]: Ran some benchmarks on subscriptions just to give some context here.

| Message Size | nats-client (msgs/sec) | nats.aio (msgs/sec) | Speedup | Throughput Gain |
|--------------|------------------------|---------------------|---------|-----------------|
| **16B** | 308,999 | 21,661 | **14.3x** | +287,338 msgs/sec |
| **64B** | 286,032 | 21,455 | **13.3x** | +264,577 msgs/sec |
| **128B** | 282,930 | 21,491 | **13.2x** | +261,439 msgs/sec |
| **256B** | 271,612 | 21,148 | **12.8x** | +250,464 msgs/sec |

---

### Key Observations:

1. **nats-client is consistently 13-14x faster** across all message sizes
2. **nats.aio is stuck at ~21K msgs/sec** - incredibly slow and consistent (note: if you pass it a callback it can process about half this).
3. **nats-client maintains ~280-310K msgs/sec** - much more reasonable
4. **Message size has minimal impact** on either client (it's CPU-bound, not I/O-bound)


The overhead in nats.aio looks to be **severe** and to me it seems to be caused by the async callback overhead and context switches we discussed earlier.

Comment (wallyqs) [reply nats-client/src/nats/client/__init__.py]: this is using the sub.next_msg code path from nats-py no?

Comment (caspervonb) [reply nats-client/src/nats/client/__init__.py]: No, this is using `sub.messages`, same bench script for both clients.
Using `sub.next_msg` is about 2x slower.

Comment (wallyqs) [reply nats-client/src/nats/client/__init__.py]: have you tried the `nats/benchmark/sub_perf.py` in the repo? 
I get this for example:
```
# nats bench pub test --size 64
python3 nats/benchmark/sub_perf.py --servers nats://localhost:4222
Waiting for 100000 messages on [test]...
****************************************************************************************************
Test completed : 337487.6190604489 msgs/sec sent
Received 100000 messages (337487.6190604489 msgs/sec)
```

File: nats-client/src/nats/client/protocol/message.py (hunk 6 above)
Comment (philpennock) [reply nats-client/src/nats/client/protocol/message.py]: I think this concern still stands, but it's comparatively minor, as a philosophical point on how to best defend against future protocol expansion.

File: nats-client/src/nats/client/protocol/types.py (hunk 7 above)
Comment (philpennock) [reply nats-client/src/nats/client/protocol/types.py]: This nit still stands.

File: nats-client/src/nats/client/message.py (hunk 10 above)
Comment (philpennock) [reply nats-client/src/nats/client/message.py:114]: As far as I've ever seen, though, the _success_ scenario just doesn't include a status code.  A tcpdump confirms that on the wire, a failure includes "<space> 503 <whatever>" before the CRLF of the "NATS/1.0" line, but a success just has "NATS/1.0<CR><LF>".

ADR-2 has 3-digit codes starting at 300, for errors.
This code has logic which explicitly checks for 200 and allows it.
I think that this 200-is-not-an-error logic is bogus?

File: nats-client/pyproject.toml (hunk 12 above)
Comment (philpennock) [reply nats-client/pyproject.toml]: This issue remains, I believe?  Are we deliberately using a new license for the new modules?

File: nats-client/src/nats/client/__init__.py (hunk 16 above)
Comment (philpennock) [reply nats-client/src/nats/client/__init__.py]: I don't know why this was resolved, without comment, when the code is not changed?

File: nats-client/src/nats/client/message.py (hunk 10 above)
Comment (caspervonb) [reply nats-client/src/nats/client/message.py:114]: Yeah I default to 200 in the Rust client, started with that and now cruft, taking it out and leaving it bare for caller to determine.

File: nats-client/src/nats/client/message.py (hunk 11 above)
Comment (caspervonb) [reply nats-client/src/nats/client/message.py]: 503 etc, but this is moot so leaving it bare just exposing code and description.

File: nats-client/src/nats/client/protocol/types.py (hunk 7 above)
Comment (caspervonb) [reply nats-client/src/nats/client/protocol/types.py]: Fixed

File: nats-client/src/nats/client/__init__.py (hunk 15 above)
Comment (caspervonb) [reply nats-client/src/nats/client/__init__.py]: Doesn't register during profiling with cProfile nor have any impact on the sub bench but noted for the next optimization pass.

File: nats-client/src/nats/client/__init__.py (hunk 16 above)
Comment (caspervonb) [reply nats-client/src/nats/client/__init__.py]: Was sure I left a comment, would log ERR from the protocol logging, but sure might as-well be explicit.

File: nats-client/src/nats/client/__init__.py (hunk 26 above)
Comment (wallyqs) [reply nats-client/src/nats/client/__init__.py]: same changes just addressing async sub iterator: https://github.com/nats-io/nats.py/pull/752

```
# nats bench pub test --size 64
uv run nats/benchmark/sub_perf_messages.py --servers nats://localhost:4222
Waiting for 100000 messages on [test]...
****************************************************************************************************
Test completed : 316181.45606272004 msgs/sec sent
Received 100000 messages (316181.45606272004 msgs/sec)
```

Comment (caspervonb) [reply nats-client/src/nats/client/__init__.py]: This is just sub tho? the above benchmark is pub sub, see tools/bench.py, you can run both clients with it.

Tried running with the patch and got 8000 messages per sec (slow consumers triggered).

Comment (wallyqs) [reply nats-client/src/nats/client/__init__.py]: what are you running exactly? I get this with that script:

```
uv run tools/bench.py --client aio --msgs 1000000 --size 128 --pub --sub
Starting pub/sub benchmark with nats.aio [msgs=1,000,000, size=128 B]

Publisher results: 
Test completed: 1,000,000 messages, 128,000,000 bytes, 4.44 seconds
  Throughput: 224,990 msgs/sec, 27.46 MB/sec
  Latency: (min/avg/max/std) = 0.00/0.00/60.50/0.42 ms

Subscriber results: 
Test completed: 1,000,000 messages, 128,000,000 bytes, 4.51 seconds
  Throughput: 221,720 msgs/sec, 27.07 MB/sec
  Latency: (min/avg/max/std) = 93.17/2306.20/4603.37/1284.93 ms
```

**Comment by caspervonb** [comment]
# Performance Benchmarks

I did some further optimization and comparison.

## Test Environment

**Hardware:**
- **Model**: MacBook Pro (Mac15,11)
- **Chip**: Apple M3 Max
- **Cores**: 14 (10 performance + 4 efficiency)
- **Memory**: 36 GB
- **OS**: macOS Darwin 24.0.0

**Software:**
- **CPython**: 3.13.7
- **PyPy**: 3.10.14 (PyPy 7.3.17)
- **NATS Server**: localhost:4222

**Test Parameters:**
- Messages per test: 1,000,000
- Message sizes: 8 B, 128 B, 1 KB, 8 KB
- Timeout: 60 seconds for slow consumers
- Modes: Queue-based and Callback-based subscriptions

---

## Complete Results Table (Ordered by Subscriber Performance)

| Runtime | Client | Mode | Size | Pub Rate (msg/s) | Pub MB/s | Sub Rate (msg/s) | Sub MB/s | Received | Dropped | Drop % |
|---------|--------|------|------|------------------|----------|------------------|----------|----------|---------|--------|
| **PyPy** | nats-client | callback | 8 B | 4,691,949 | 35.80 | **4,615,995** 🥇 | 35.22 | 1,000,000 | 0 | 0% |
| PyPy | nats-client | queue | 8 B | 4,565,524 | 34.83 | **2,336,040** | 17.82 | 1,000,000 | 0 | 0% |
| **PyPy** | nats-client | callback | 128 B | 2,459,309 | 300.21 | **2,243,722** | 273.89 | 1,000,000 | 0 | 0% |
| **PyPy** | nats-client | queue | 128 B | 1,674,313 | 204.38 | **1,402,979** | 171.26 | 1,000,000 | 0 | 0% |
| PyPy | nats.aio | callback | 8 B | 2,493,561 | 19.02 | **924,494** | 7.05 | 1,000,000 | 0 | 0% |
| **CPython** | nats-client | callback | 8 B | 1,579,744 | 12.05 | **880,430** | 6.72 | 1,000,000 | 0 | 0% |
| PyPy | nats.aio | callback | 128 B | 775,115 | 94.62 | **629,760** | 76.87 | 1,000,000 | 0 | 0% |
| **PyPy** | nats-client | callback | 1 KB | 609,447 | 595.16 | **615,474** | 601.05 | 1,000,000 | 0 | 0% |
| **CPython** | nats-client | callback | 128 B | 704,567 | 86.01 | **562,402** | 68.65 | 1,000,000 | 0 | 0% |
| CPython | nats-client | queue | 8 B | 1,564,495 | 11.94 | **553,636** | 4.22 | 1,000,000 | 0 | 0% |
| **PyPy** | nats-client | queue | 1 KB | 487,882 | 476.45 | **493,128** | 481.57 | 1,000,000 | 0 | 0% |
| **CPython** | nats-client | callback | 1 KB | 440,026 | 429.71 | **430,058** | 419.98 | 1,000,000 | 0 | 0% |
| CPython | nats-client | queue | 128 B | 544,772 | 66.50 | **418,524** | 51.09 | 1,000,000 | 0 | 0% |
| **CPython** | nats-client | queue | 1 KB | 342,981 | 334.94 | **338,365** | 330.43 | 1,000,000 | 0 | 0% |
| PyPy | nats.aio | callback | 1 KB | 288,049 | 281.30 | **281,563** | 274.96 | 1,000,000 | 0 | 0% |
| CPython | nats.aio | callback | 128 B | 250,276 | 30.55 | **250,934** | 30.63 | 1,000,000 | 0 | 0% |
| CPython | nats.aio | callback | 8 B | 246,990 | 1.88 | **236,046** | 1.80 | 1,000,000 | 0 | 0% |
| **CPython** | nats.aio | callback | 1 KB | 221,796 | 216.60 | **217,994** | 212.88 | 1,000,000 | 0 | 0% |
| **CPython** | nats-client | callback | 8 KB | 166,760 | 1302.81 | **168,008** | 1312.56 | 1,000,000 | 0 | 0% |
| CPython | nats-client | queue | 8 KB | 154,709 | 1208.66 | **157,786** | 1232.70 | 1,000,000 | 0 | 0% |
| CPython | nats.aio | callback | 8 KB | 121,886 | 952.23 | **121,416** | 948.56 | 1,000,000 | 0 | 0% |
| **PyPy** | nats-client | callback | 8 KB | 101,937 | 796.38 | **102,817** | 803.26 | 1,000,000 | 0 | 0% |
| PyPy | nats-client | queue | 8 KB | 92,085 | 719.42 | **92,806** | 725.05 | 1,000,000 | 0 | 0% |
| PyPy | nats.aio | callback | 8 KB | 63,666 | 497.39 | **63,503** | 496.12 | 1,000,000 | 0 | 0% |
| CPython | nats.aio | queue | 8 B | 118,989 | 0.91 | 8,769 | 0.07 | 524,621 | 475,379 | **47.5%** ⚠️ |
| CPython | nats.aio | queue | 128 B | 104,272 | 12.73 | 8,758 | 1.07 | 524,680 | 475,320 | **47.5%** ⚠️ |
| PyPy | nats.aio | queue | 8 B | 2,627,229 | 20.04 | 8,755 | 0.07 | 524,349 | 475,651 | **47.5%** ⚠️ |
| PyPy | nats.aio | queue | 128 B | 406,140 | 49.58 | 8,756 | 1.07 | 524,660 | 475,340 | **47.5%** ⚠️ |
| CPython | nats.aio | queue | 1 KB | 72,483 | 70.78 | 2,232 | 2.18 | 133,723 | 866,277 | **86.6%** ⚠️ |
| PyPy | nats.aio | queue | 1 KB | 131,912 | 128.82 | 2,232 | 2.18 | 133,726 | 866,274 | **86.6%** ⚠️ |
| CPython | nats.aio | queue | 8 KB | 49,323 | 385.34 | 622 | 4.86 | 37,265 | 962,735 | **96.2%** ⚠️ |
| PyPy | nats.aio | queue | 8 KB | 54,782 | 427.99 | 622 | 4.86 | 37,264 | 962,736 | **96.2%** ⚠️ |

---

## Key Findings

### 1. Performance Rankings

**By Subscriber Throughput:**
1. 🥇 **PyPy + nats-client callback**: 103K - 4.6M msg/s (BEST)
2. **PyPy + nats-client queue**: 93K - 2.3M msg/s
3. **PyPy + nats.aio callback**: 64K - 924K msg/s
4. **CPython + nats-client callback**: 168K - 880K msg/s
5. **CPython + nats-client queue**: 158K - 554K msg/s
6. **CPython + nats.aio callback**: 121K - 251K msg/s
7. ⚠️ **nats.aio queue mode**: 622 - 8.8K msg/s with **47-96% message loss**

### 2. Message Loss Analysis

**nats.aio Queue Mode - Critical Issue:**
- **8 B**: 47.5% dropped (475K messages lost)
- **128 B**: 47.5% dropped (475K messages lost)
- **1 KB**: 86.6% dropped (866K messages lost)
- **8 KB**: 96.2% dropped (963K messages lost)

**All other configurations**: 0% message loss ✅

### 3. Runtime Comparison

**PyPy vs CPython (nats-client callback):**
- **8 B**: 5.2x faster with PyPy
- **128 B**: 4.0x faster with PyPy
- **1 KB**: 1.4x faster with PyPy
- **8 KB**: 0.6x (CPython faster due to PyPy memory overhead)

### 4. Implementation Comparison

**nats-client vs nats.aio (callback mode):**
- **CPython**: 1.4x - 3.7x faster
- **PyPy**: 1.6x - 5.0x faster

**Callback vs Queue (nats-client):**
- **CPython**: 1.0x - 1.6x faster
- **PyPy**: 1.1x - 2.0x faster


## Performance Comparison

### Callback Mode (Subscriber Throughput)

| Message Size | CPython nats-client | CPython nats.aio | Speedup | PyPy nats-client | PyPy nats.aio | Speedup |
|--------------|---------------------|------------------|---------|------------------|---------------|---------|
| 8 B | 880,430 msg/s | 236,046 msg/s | **3.7x** | 4,615,995 msg/s | 924,494 msg/s | **5.0x** |
| 128 B | 562,402 msg/s | 250,934 msg/s | **2.2x** | 2,243,722 msg/s | 629,760 msg/s | **3.6x** |
| 1 KB | 430,058 msg/s | 217,994 msg/s | **2.0x** | 615,474 msg/s | 281,563 msg/s | **2.2x** |
| 8 KB | 168,008 msg/s | 121,416 msg/s | **1.4x** | 102,817 msg/s | 63,503 msg/s | **1.6x** |

### Queue Mode (Subscriber Throughput)

| Message Size | CPython nats-client | CPython nats.aio | Speedup | PyPy nats-client | PyPy nats.aio | Speedup |
|--------------|---------------------|------------------|---------|------------------|---------------|---------|
| 8 B | 553,636 msg/s | 8,769 msg/s* | **63x** | 2,336,040 msg/s | 8,755 msg/s* | **267x** |
| 128 B | 418,524 msg/s | 8,758 msg/s* | **48x** | 1,402,979 msg/s | 8,756 msg/s* | **160x** |
| 1 KB | 338,365 msg/s | 2,232 msg/s* | **152x** | 493,128 msg/s | 2,232 msg/s* | **221x** |
| 8 KB | 157,786 msg/s | 622 msg/s* | **254x** | 92,806 msg/s | 622 msg/s* | **149x** |

*\* nats.aio queue mode experiences **severe message** loss (47-96%)*

### Summary

- **Callback mode**: nats-client is **1.4x - 5.0x faster** depending on runtime and message size
- **Queue mode**: nats-client is **48x - 267x faster** and delivers 100% of messages vs 4-53% for nats.aio
- **Performance gap increases** with smaller messages and PyPy runtime
- **nats.aio queue mode** is fundamentally broken under high load with catastrophic message loss

## Benchmark Commands

```bash
cd nats-client

# Individual tests
uv run python tools/bench.py --client client --messages 1000000 --size 128 [--callback]

uv run python tools/bench.py --client aio --messages 1000000 --size 128 [--callback]
```

PR #733, opened 2025-09-23
Migrate project to a uv workspace

caspervonb:


pyproject.toml
```diff
@@ -1,53 +1,16 @@
...
-
-[tool.setuptools.packages.find]
-include = ["nats*"]  # package names should match these glob patterns (["*"] by default)
-exclude = ["scripts", "tests"]  # exclude packages matching these glob patterns (empty by default)
-namespaces = false  # to disable scanning PEP 420 namespaces (true by default)
+workspace = { members = ["nats"] }
 
 [tool.mypy]
-files = ["nats"]
+files = ["nats/src"]
```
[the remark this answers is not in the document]
wallyqs: why is src necessary? can it be rolled back?

.github/workflows/check.yml
```diff
@@ -1,33 +1,33 @@
...
+              with:
+                  python-version: "3.8"
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install pipenv
-          pipenv install --dev
+            - name: Install uv
+              run: pip install uv
```
philpennock: Probably better to `uses: astral-sh/setup-uv` and get caching etc, and avoid hitting the pypi servers unnecessarily.

.github/workflows/test.yml
```diff
@@ -25,23 +25,24 @@ jobs:
 
         steps:
             - name: Check out repository
-              uses: actions/checkout@v2
+              uses: actions/checkout@v3
```
philpennock: v5

.github/workflows/test.yml
```diff
@@ -25,23 +25,24 @@ jobs:
...
 
             - name: Set up Python ${{ matrix.python-version }}
-              uses: actions/setup-python@v2
+              uses: actions/setup-python@v4
```
philpennock: v6

.github/workflows/test.yml
```diff
@@ -25,23 +25,24 @@ jobs:
...
               with:
                   python-version: ${{ matrix.python-version }}
 
-            - name: Install dependencies
+            - name: Install uv
+              run: pip install uv
```
philpennock: same here, `uses: astral-sh/setup-uv`

.github/workflows/test.yml
```diff
@@ -25,23 +25,24 @@ jobs:
...
+
+            - name: Install dependencies and project
               run: |
-                  pip install pipenv
-                  pip install certifi
-                  pipenv install --dev
-                  bash ./scripts/install_nats.sh
+                  uv sync --dev
+                  bash ./nats/scripts/install_nats.sh
```
philpennock: This should probably avoid explicitly specifying `bash`, and leave that to the `#!` line.

nats/pyproject.toml
```diff
@@ -0,0 +1,46 @@
+[build-system]
+requires = ["setuptools>=68.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nats-py"
+version = "2.11.0"
```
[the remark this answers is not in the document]
philpennock: I loosely agree with Copilot here.  It would be great if the `uv_build` backend already supported dynamic metadata, then we could switch entirely away from setuptools even as a backend.

I would amend Copilot's suggestion slightly though: we should explicitly mark the dynamic field:

```ini
[project]
dynamic = ["version"]
```

See: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata

nats/README.md
```diff
@@ -0,0 +1,253 @@
...
+An [asyncio](https://docs.python.org/3/library/asyncio.html) Python client for the [NATS messaging system](https://nats.io).
+
+[![docs](https://img.shields.io/static/v1?label=docs&message=docs&color=informational)](https://nats-io.github.io/nats.py/)
+[![pypi](https://img.shields.io/pypi/v/nats-py.svg)](https://pypi.org/project/nats-py)
+[![Versions](https://img.shields.io/pypi/pyversions/nats-py.svg)](https://pypi.org/project/nats-py)
+[![License Apache 2.0](https://img.shields.io/badge/License-Apache2-blue.svg)](https://www.apache.org/licenses/LICENSE-2.0)
+
+## Supported platforms
+
+Should be compatible with at least [Python +3.8](https://docs.python.org/3.8/library/asyncio.html).
```
philpennock: Was that a typo for `Python 3.8+` ?

nats/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Installing
+
+```bash
+pip install nats-py
+```
```
philpennock, line 18: GitHub is showing this as a new doc, not primarily a rename; if it is new, then can we take this chance to mention the optional dependencies here, with a "see below for details", so that the other install options are in the Installing section?  Personally I find it slightly frustrating when the Installing step is missing functionality folks might need, such as nkeys support.

.gitignore
```diff
@@ -3,6 +3,9 @@
 *.py[cod]
 .coverage
 
+# Lockfiles
+*.lock
```
philpennock: Wait what?  No.  The `uv.lock` file should be committed into the repo.  This provides for consistent versions, some protection against supply chain attack (not full, merely window reduction), and helps with reproducible builds.

This is why I'm going to NAK the review instead of just commenting with issues: I am firmly opposed to keeping `uv.lock` out of git.

pyproject.toml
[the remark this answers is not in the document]
philpennock: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

Generally `src/` is the way the ecosystem is moving, because getting bitten by import bugs without it makes for debugging nightmares.  Sticking with a flat layout will have us swimming against the tide.  I think the gains are small enough to not make it worth a migration on its own, but big enough that if we're migrating to a new build-system anyway then we might as well do so at this time, so I think this change is a good thing.

.github/workflows/test.yml
caspervonb: Done

.github/workflows/test.yml
caspervonb: Done

nats/README.md
caspervonb: Not from this pull request, but yeah might as-well fix it.

.gitignore
caspervonb: Question is, if we "re-export" packages like nats-jetstream, we probably don't want them pinned? Committed the lock file for now tho. We only have dev dependencies on pytest etc.

.github/workflows/test.yml
caspervonb: Bumped

.github/workflows/check.yml
caspervonb: Sure

nats/README.md
caspervonb: Basically just a copy, will revise a little bit.

nats/pyproject.toml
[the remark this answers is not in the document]
caspervonb: Inverted it, so we assign __version__ from importlib metadata instead.

philpennock (approved):

PR #734, opened 2025-09-23
Add nats-jetstream package

caspervonb:
This adds a new modern (following the newer simplified interface) JetStream package under the name `nats-jetstream` which implements the following:

- [ ] ADR-7: NATS Server Error Codes (errors need work)
- [x] ADR-10: JetStream Extended Purge
- [x] ADR-13: Pull Subscribe Internals (needs review)
- [ ] ADR-17: Ordered Consumer (not started)
- [x] ADR-22: JetStream Publish Retries on No Responders
- [x] ADR-28: JetStream RePublish
- [x] ADR-31: JetStream Direct Get
- [x] ADR-33: Metadata for Stream and Consumer
- [x] ADR-34: JetStream Consumers Multiple Filters
- [x] ADR-36: Subject Mapping Transforms in Streams
- [x] ADR-37: JetStream Simplification
- [ ] ADR-42: Priority Groups for Pull Consumers (partially implemented - missing pinned and overflow modes)
- [x] ADR-44: Consumer Pause/Resume

Depends on #732


nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,344 @@
...
+            if raise_on_error and is_error_response(data):
+                raise Error.from_response(data["error"])
+
+            is_valid, unknown_keys, missing_keys = check_response(
+                data, response_type
+            )
+            if not is_valid:
+                if missing_keys:
+                    logger.warning(
+                        "[%s] Missing keys in response: %s", request_id,
```
aricart: why is the client creating schemas for the responses and then validating them? - the server will now do that on input, so hopefully the response is correct?

caspervonb: This bit isn't quite working optimally yet, but the intent is to validate very lightly the other way around. Just looks if the structural type knows about the key rather than full on json schema.

Server validates if we SEND it wrong keys with strict setting, but this is to validate if we RECEIVE unknown keys from the server.

e.g Server sends us a response and our types don't know about those keys. If caught in development that means we're out of parity and in theory a nice little dev helper.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def metadata(self) -> Metadata:
+        """Return metadata for the JetStream message."""
+        return self._metadata
+
+    def data(self) -> bytes:
+        """Return the message body."""
+        return self._data
+
+    def headers(self) -> Headers | None:
```
caspervonb, line 189: Needs to be a property.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+
+        # Override with explicitly provided metadata if available
+        if metadata is not None:
+            self._metadata = metadata
+
+    def metadata(self) -> Metadata:
+        """Return metadata for the JetStream message."""
+        return self._metadata
+
+    def data(self) -> bytes:
```
caspervonb, line 184: Needs to be a property.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+        if reply_to:
+            parsed_metadata = Metadata.from_reply(reply_to)
+            if parsed_metadata:
+                self._metadata = parsed_metadata
+
+        # Override with explicitly provided metadata if available
+        if metadata is not None:
+            self._metadata = metadata
+
+    def metadata(self) -> Metadata:
```
caspervonb, line 179: Needs to be a property.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def data(self) -> bytes:
+        """Return the message body."""
+        return self._data
+
+    def headers(self) -> Headers | None:
+        """Return the message headers, or None if no headers exist."""
+        return self._headers
+
+    def subject(self) -> str:
```
caspervonb, line 194: Needs to be a property.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+        """Return the subject on which the message was published/received."""
+        return self._subject
+
+    def reply_to(self) -> str:
```
caspervonb: Needs to be a property.

nats-jetstream/src/nats/jetstream/__init__.py
```diff
@@ -0,0 +1,584 @@
...
+            payload,
+            headers=headers,
+            timeout=5.0,  # Use a longer timeout for publish operations
+        )
+
+        publish_ack = PublishAck.from_response(json.loads(response.data))
+
+        return publish_ack
+
+    async def stream_names(self,
```
caspervonb: Should we be consistent or follow Go?

```suggestion
    async def get_stream_names(self,
```

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+class StreamConfig:
+    """Configuration for a JetStream stream"""
+
+    # All fields are optional with defaults
+    max_age: int = 0  # Maximum age of any message in nanoseconds (0 for unlimited)
+    max_bytes: int = -1  # Maximum size in bytes (-1 for unlimited)
+    max_consumers: int = -1  # Maximum number of consumers (-1 for unlimited)
+    max_msgs: int = -1  # Maximum number of messages (-1 for unlimited)
+    num_replicas: int = 1  # Number of replicas to keep for each message
+    retention: Literal["limits", "interest",
```
caspervonb: Needs a type alias

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+                       "workqueue"] = "limits"  # How messages are retained
+    storage: Literal["file", "memory"] = "file"  # Storage backend to use
```
caspervonb: Needs a type alias

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+
+    allow_direct: bool | None = None  # Allow direct access to get messages
+    allow_msg_ttl: bool | None = None  # Enable per-message TTL using headers
+    allow_rollup_hdrs: bool | None = None  # Allow Nats-Rollup header usage
+    compression: Literal["none", "s2"] | None = None  # Compression algorithm
```
caspervonb: Needs a type alias

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    deny_delete: bool | None = None  # Restrict message deletion via API
+    deny_purge: bool | None = None  # Restrict stream purge via API
+    description: str | None = None  # Stream description
+    discard: Literal["old", "new"] | None = None  # Discard policy
```
caspervonb: Needs a type alias

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    discard_new_per_subject: bool | None = None  # Apply discard new per subject
+    duplicate_window: int | None = None  # Duplicate tracking window in ns
+    first_seq: int | None = None  # First message sequence number
+    max_msg_size: int | None = None  # Maximum message size (-1 for unlimited)
+    max_msgs_per_subject: int | None = None  # Per-subject message limit
+    metadata: dict[str, str] | None = None  # Additional metadata
+    mirror: api.StreamSource | None = None  # Mirror configuration
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    mirror_direct: bool | None = None  # Allow direct access for mirrors
+    name: str | None = None  # Stream name
+    no_ack: bool | None = None  # Disable message acknowledgment
+    placement: api.Placement | None = None  # Replica placement directives
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    republish: api.Republish | None = None  # Message republishing config
+    sealed: bool | None = None  # Prevent message deletion and updates
+    sources: list[api.StreamSource] | None = None  # Replication sources
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    first_seq: int | None = None  # First message sequence number
+    max_msg_size: int | None = None  # Maximum message size (-1 for unlimited)
+    max_msgs_per_subject: int | None = None  # Per-subject message limit
+    metadata: dict[str, str] | None = None  # Additional metadata
+    mirror: api.StreamSource | None = None  # Mirror configuration
+    mirror_direct: bool | None = None  # Allow direct access for mirrors
+    name: str | None = None  # Stream name
+    no_ack: bool | None = None  # Disable message acknowledgment
+    placement: api.Placement | None = None  # Replica placement directives
+    republish: api.Republish | None = None  # Message republishing config
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    sealed: bool | None = None  # Prevent message deletion and updates
+    sources: list[api.StreamSource] | None = None  # Replication sources
+    subject_delete_marker_ttl: int | None = None  # TTL for delete markers
+    subject_transform: api.SubjectTransform | None = None  # Subject transform
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+    subjects: list[str] | None = None  # Subjects to consume
+    template_owner: str | None = None  # Managing template name
+    consumer_limits: api.StreamConsumerLimits | None = None  # Consumer limits
```
caspervonb: Needs a refined data type serialized from structured type.

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+        """Fetch a batch of messages from the consumer without waiting."""
+        ...
+
+    async def next(self, max_wait: float = 5.0) -> Message:
+        """Fetch a single message from the consumer."""
+        ...
+
+    async def consume(
+        self,
+        callback: Callable[[Message], Awaitable[None]],
```
caspervonb: Require named parameters

```suggestion
        callback: Callable[[Message], Awaitable[None]],
        *,
```

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    @overload
+    async def fetch_nowait(
+        self,
+        *,
+        max_bytes: int | None = None,
+    ) -> MessageBatch:
+        """Fetch a batch of messages from the consumer without waiting."""
+        ...
+
+    async def next(self, max_wait: float = 5.0) -> Message:
```
caspervonb, line 471: Rename

```suggestion
    async def next(self, timeout: float = 5.0) -> Message:
```

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+        """Get cached consumer info."""
+        ...
+
+    async def get_info(self) -> ConsumerInfo:
+        """Get consumer info from the server."""
+        ...
+
+    @overload
+    async def fetch(
+        self,
```
caspervonb, line 434: Require named parameters.

```suggestion
        self,
        *,
```

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    from ..message import Message
+
+
+@dataclass
+class ConsumerInfo:
+    """Information about a consumer."""
+
+    stream_name: str
+    name: str
+    config: dict[str, Any]
```
caspervonb: Strictly type

```suggestion
    config: ConsumerConfig
```

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+        Args:
+            sequence: The sequence number of the message to delete
+        """
+        api = getattr(self._jetstream, "_api", None)
+        if api is None:
+            raise RuntimeError("JetStream does not have an API client")
+
+        await api.stream_msg_delete(self._name, seq=sequence, no_erase=False)
+
+    async def upsert_consumer(
```
caspervonb: Make private

```suggestion
    async def _upsert_consumer(
```

nats-jetstream/src/nats/jetstream/__init__.py
```diff
@@ -0,0 +1,584 @@
...
+    async def get_stream(self, name: str) -> Stream:
+        """Get a stream by name."""
+        info = await self.get_stream_info(name)
+        return Stream(self, name, info)
+
+    async def create_consumer(
+        self,
+        stream_name: str,
+        name: str,
+        durable_name: str | None = None,
```
caspervonb: Pack into kwargs

```suggestion
```

caspervonb: Pack into kwargs

```suggestion
```

nats-jetstream/src/nats/jetstream/util.py
```diff
@@ -0,0 +1,15 @@
...
+def new_inbox() -> str:
+    """Generate a new NATS inbox name.
+
+    This is an internal utility function that creates a unique inbox subject
+    by generating a UUID and formatting it in the NATS inbox format.
+
+    Returns:
+        A unique inbox subject string in the format "_INBOX.{uuid}"
+    """
+    return f"_INBOX.{uuid.uuid4().hex}"
```
wallyqs, line 15: what happened to the nuid?

nats-jetstream/examples/pull_consume_callback.py
```diff
@@ -0,0 +1,80 @@
...
+from nats.jetstream import new as new_jetstream
+
+
+async def endless_publish(nc, js):
+    """Continuously publish messages to the stream."""
+    i = 0
+    while True:
+        await asyncio.sleep(0.5)
+        if nc.status != ClientStatus.CONNECTED:
+            continue
```
philpennock: Since this is example code for people to start from, and people do Interesting Things when copy/pasting without full understanding, I think it's worth a small comment:

```python
        if nc.status != ClientStatus.CONNECTED:
            # NB: this is only safe because of the sleep above; beware busy loops
            continue
```

nats-jetstream/examples/pull_consume_callback.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            await js.publish("FOO.TEST1", f"msg {i}".encode())
+        except Exception as err:
+            print("pub error: ", err)
```
philpennock: **grumble** _(unix stdio pedant)_ : `, file=sys.stderr, flush=True)`  (and import `sys` above)

nats-jetstream/examples/pull_consume_iterator.py
```diff
@@ -0,0 +1,91 @@
...
+from nats.jetstream import new as new_jetstream
+
+
+async def endless_publish(nc, js):
+    """Continuously publish messages to the stream."""
+    i = 0
+    while True:
+        await asyncio.sleep(0.5)
+        if nc.status != ClientStatus.CONNECTED:
+            continue
```
philpennock: Similarly to other suggestion re commenting for safety when people copy/paste with less than full understanding.  (I won't repeat this one for any other instances, nor my stdio pedantry for line 21 🤐).

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://nats.io/schemas/jetstream/api/v1/account_info_response.json",
```
philpennock, line 3: Uh, are we expecting every client library for each language to hold a copy of these?  That seems ... sub-optimal.  Shouldn't there be one primary authoritative location, designed to be pulled in via git submodules or whichever other repo-linking poison folks prefer?

(I'm assuming no point reviewing the schemas in this PR so skipping past them.)

nats-jetstream/src/nats/jetstream/api/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    PublishAck,
+    Republish,
+    StreamConfig,
+    StreamConsumerLimits,
+    StreamInfo,
+    StreamSource,
+    StreamState,
+    SubjectTransform,
+    Tier,
+    LostStreamData,
```
philpennock: sort

nats-jetstream/src/nats/jetstream/api/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+__all__ = [
```
philpennock, line 27: This would be easier to review if (1) it were sorted and (2) the `from .types` above had comments on the entries to not be exported.

In fact, since we're specifying `__all__` anyway, it seems it would be safe to use `from . import types` and then `globals()[attr] = getattr(types, attr)` with `attr` being a list, letting these two sets then be in sync.  I'm okay with letting this go, but do think we should think carefully about maintainability of keeping 2 separate lists here.

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def is_error_response(data: Any) -> TypeGuard[ErrorResponse]:
+    return isinstance(data, dict) and "error" in data
+
+
+def check_response(
+    data: Any, expected_type: type[ResponseT]
+) -> tuple[bool, set[str] | None, set[str] | None]:
+    # TODO check for missing and unknown keys
```
philpennock: For clarity: are you expecting this TODO to be done before this PR moves out of draft, or for some later milestone?

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,353 @@
...
+        if not stream_name:
+            raise ValueError("stream_name is required")
+
+        consumer_config = request.get("config")
+        if not consumer_config:
+            raise ValueError("config is required")
+
+        consumer_name = consumer_config.get("name")
+        if not consumer_name:
+            raise ValueError("name is required")
```
philpennock: This would be more friendly if all 3 were collected, and then `if not (a and b and c)` used to return a ValueError listing all the missing fields at once.

(nb: this is an opinion and not a blocking issue, happy to be ignored on this)

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,353 @@
...
+        if max_msgs < -1:
+            raise ValueError(
+                "max_msgs must be -1 (unlimited) or a positive number"
+            )
+
+        # Validate mirror configuration
+        mirror = kwargs.get("mirror")
+        subjects = kwargs.get("subjects")
+        if mirror is not None and subjects:
+            raise ValueError("Cannot specify both mirror and subjects")
```
philpennock: I look at logic like this and can't help but think "Isn't this schema enforcement? If we have all these `.json` schema files, why aren't the client functions (at least for infrequent ops like stream create) just validating the config against the schema file instead of implementing policy checks manually?"

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,353 @@
...
+        )
+
+    @overload
+    async def request_json(
+        self,
+        subject: str,
+        payload: Any | None = None,
+        *,
+        response_type: type[ResponseT],
+        raise_on_error: Literal[True] = True,
```
philpennock, line 366: This, and the same in the next `@overload`, is missing the `timeout` parameter; the implementation is assignable to these function signatures so it should work, but it looks like this means that type-checkers should reject attempts to actually provide the `timeout` parameter.  Have I misunderstood?  Is this intentional?  Or is it just a forgotten parameter for the signature?

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,353 @@
...
+    async def request_json(
+        self,
+        subject: str,
+        payload: Any | None = None,
+        *,
+        response_type: type[ResponseT],
+        raise_on_error: bool = True,
+        timeout: float = 5.0,
+    ) -> ResponseT | ErrorResponse:
+        request_id = str(uuid.uuid4())[:8]
```
philpennock, line 388: This is reading 16 bytes from /dev/urandom every time, just to return the representation of the first 4 bytes.  Aside from making me twitch (which is probably amusing 😉), that's also going to be a device read on every JSON request.

The `secrets` module was introduced in Python 3.6; `secrets.token_hex(4)` will be more efficient, but still hit `/dev/urandom`.  But the only goal is a distinct ID, there's no defense against other clients knowing the ID, right?  Presumably not using an incrementing integer to avoid threading issues, but _should_ then be using a client-side cache to prevent re-use, so just rolling the dice here (but the die has a lot of sides).  I'd be tempted to use an in-process PRNG seeded once and just roll with that.

For now, just use `secrets.token_hex(4)` please and remember to benchmark to see if this ever matters with lots of small JSON requests.

philpennock (commented):
Reviewed through end of the `src/nats/jetstream/api/` folder, need to resume from the `consumer/` sibling to that when I next look at this again.  Sending what I've looked at so far though.  (Will be around on Thursday to resume).

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
piotrpio: Agreed, those schemas live and are maintained in `jsm.go`, it'd be hard to keep it up to date across many repos.

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    created: int
+    delivered: dict[str, int]
+    ack_floor: dict[str, int]
+    num_ack_pending: int
+    num_redelivered: int
+    num_waiting: int
+    num_pending: int
+    cluster: dict[str, Any] | None = None
```
philpennock, line 325: Is there a better type for the values in this dict?

nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            cluster=data.get("cluster"),
+            push_bound=data.get("push_bound"),
+        )
+
+
+class MessageBatch(Protocol):
+    """Protocol for a batch of messages retrieved from a JetStream consumer."""
+
+    def __aiter__(self) -> AsyncIterator[Message]:
+        """Return self as an async iterator."""
```
philpennock, line 390: Here and in `MessageStream` below, is this really `self`?

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+    from nats.client.message import Message as ClientMessage
+    from nats.jetstream.stream import Stream
+
+MessageCallback = Callable[[Message], Awaitable[None]]
+
+
+class MessageHandler:
+    """Handler for message consumption with callback processing."""
+
+    def __init__(self, task: asyncio.Task, message_stream: PullMessageStream):
```
philpennock: minor nit: `-> None:` for return type; I can't remember where this cropped up for me, but one of the type checkers happily switched modes to incomplete checking without this.

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+        self._task = task
+        self._message_stream = message_stream
+
+    @property
+    def is_active(self) -> bool:
+        """Check if the consume task is still active."""
+        return not self._task.done() and self._message_stream.is_active
```
philpennock: I'm not digging into this, but am hesitant when I see this: if the message stream ends and there are messages outstanding for the consumer, will this correctly report as still being active until those last messages have been consumed?  It seems to me that the stream could legitimately stop being active first?

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+                            continue
+                        case "404":  # No messages
+                            raise StopAsyncIteration
+                        case "408":  # Timeout
+                            raise StopAsyncIteration
+                        case "409":  # Message Size Exceeds MaxBytes
+                            raise StopAsyncIteration
+                        case _:
+                            self._error = Exception(
+                                f"Status {raw_msg.status.code}: {raw_msg.status.description or ''}"
```
philpennock: minor nit: can we use `or '(missing description)'` to make it clearer if this exception is ever seen that there was something expected that's missing?

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+                    jetstream=self._jetstream,
+                )
+
+                self._pending_messages -= 1
+                return js_msg
+        except (StopAsyncIteration, asyncio.TimeoutError):
+            if not self._terminated:
+                await self._subscription.unsubscribe()
+                self._terminated = True
+            raise StopAsyncIteration
```
philpennock, line 208: I'm hesitant because this seems to turn a timeout into an EOF, suppressing the timeout indication.

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+                    case "409":  # Message Size Exceeds MaxBytes - batch is complete due to size limit
+                        # Reset pending counts as the batch is done
+                        self._pending_messages = 0
+                        self._pending_bytes = 0
+                        # Immediately request more messages
+                        await self._send_request()
+                        continue
+                    case _:  # Error
+                        await self._cleanup()
+                        raise StopAsyncIteration
```
philpennock, line 366: This seems to silently swallow errors and just stop the iteration, which seems bad?

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+        subject = f"{api_prefix}.CONSUMER.MSG.NEXT.{self._consumer.stream_name}.{self._consumer.name}"
+        payload = json.dumps(request).encode()
+
+        await jetstream.client.publish(
+            subject, payload=payload, reply_to=self._subscription.subject
+        )
+
+        self._pending_messages += self._batch
+        if self._max_bytes is not None:
+            self._pending_bytes += self._max_bytes
```
philpennock, line 421: What's the rationale here?  (Perhaps I'm just being dense)

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+        """Background task that continuously ensures we have pending requests."""
+        try:
+            # Send initial request
+            await self._send_request()
+
+            while not self._terminated:
+                await asyncio.sleep(0.1)  # Check every 100ms
+
+                # Send new request if we're running low on messages
+                should_request = self._pending_messages < self._batch // 2
```
philpennock: If it takes a bit of time for the messages to come back after a request, doesn't this mean we can send up to `self._batch // 2` sequential requests for more messages?  Seems we'd want a state toggle to track "have requested more"

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -0,0 +1,625 @@
...
+
+                # Also consider bytes if we have a limit
+                if self._max_bytes is not None:
+                    should_request = should_request or self._pending_bytes < self._max_bytes // 2
```
philpennock: And if there's a delay here, we could end up significantly over max_bytes, which might be surprising profiling behavior

nats-jetstream/src/nats/jetstream/__init__.py
```diff
@@ -0,0 +1,584 @@
...
+
+@dataclass
+class APIStats:
+    """JetStream API statistics."""
+
+    total: int
+    errors: int
+
+    @classmethod
+    def from_response(cls, data: api.ApiStats) -> APIStats:
```
philpennock: Is this being done this way for performance, or avoiding a dependency upon pydantic, or something?  I have concerns about maintenance as the API changes over time and new fields getting missed.  Perhaps reflection-based tests which ensure that any field seen in the schema must appear as an attribute in the corresponding dataclass?

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+
+    domain: str = ""
+    """Domain this message was received on (empty string if not set)."""
+
+    @classmethod
+    def from_reply(cls, reply_subject: str) -> Metadata | None:
+        """Parse metadata from a JetStream reply subject.
+
+        The reply subject contains the message metadata in the form:
+        $JS.ACK.<stream>.<consumer>.<delivered>.<stream sequence>.<consumer sequence>.<timestamp>.<pending>
```
philpennock: This doesn't handle JS domains.

<https://docs.nats.io/reference/reference-protocols/nats_api_reference#reference>:
> Note that if the resources you're trying to access have a JetStream domain associated with them, then the subject prefix will be $JS.{domain}.API rather than $JS.API.

nats-jetstream/src/nats/jetstream/message.py
```diff
@@ -0,0 +1,306 @@
...
+            ValueError: If reply_to or jetstream reference is missing.
+        """
+        if not self._reply_to:
+            raise ValueError("Cannot terminate message without reply_to")
+        if not self._jetstream:
+            raise ValueError(
+                "Cannot terminate message without JetStream reference"
+            )
+
+        payload = f"+TERM {reason}".encode()
```
philpennock, line 338: I don't see reason support in the Go client.  Where might reasons come from, and should we guard against embedded `CRLF` inside the reason leading to protocol injection?

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+                self._name, seq=sequence, last_by_subj=last_by_subject
+            )
+            message = response["message"]
+
+        # Decode base64 data if present
+        data = b""  # Initialize with empty bytes by default
+        if "data" in message:
+            data = base64.b64decode(message["data"])
+
+        # Parse headers if present
```
philpennock, line 1161: There was something in `message.py` (I think) in this same dir which had header parsing as a TODO item.  Is this suitable for lifting out into a new function and calling there too?

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -0,0 +1,620 @@
...
+        headers = None
+        if "hdrs" in message:
+            try:
+                headers_bytes = base64.b64decode(message["hdrs"])
+                headers_str = headers_bytes.decode("utf-8")
+                if headers_str.startswith("NATS/1.0\r\n"):
```
philpennock: Should there be an exception raised if the headers don't start with this?

nats-jetstream/tools/fetch_schemas.py
```diff
@@ -0,0 +1,106 @@
...
+        path.write_text(json.dumps(data, indent=2) + "\n")
+    except Exception as e:
+        print(f"Error fixing JSON file {path}: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    """Main entry point."""
+    # Get the root directory of our project
+    root_dir = Path(__file__).parent.parent
```
philpennock, line 54: This is fragile to current location in the git repo.  If you're going to invoke `git` anyway for this rare maintenance operation, then `git rev-parse --show-toplevel` is your friend here.

nats-jetstream/examples/pull_consume_callback.py
caspervonb: Yeah example got stale, need new ones.

nats-jetstream/src/nats/jetstream/message.py
caspervonb: https://pkg.go.dev/github.com/nats-io/nats.go/jetstream#Msg.TermWithReason

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
caspervonb: These are pulled from there, just with a little bit of filtering unwanted files. I'd do a submodule if we had nats-io/schema or something.

nats-jetstream/src/nats/jetstream/message.py
caspervonb: Fixed

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
wallyqs: I guess there is a task to fetch them from jsm.go?

nats-jetstream/src/nats/jetstream/consumer/__init__.py
caspervonb: Yeah strongly typed all of it with dataclasses.

nats-jetstream/src/nats/jetstream/consumer/pull.py
caspervonb: Took out message handler in its entirety.. need to really think if this is a pattern we want in python, which has first class asynchronous iteration. One can always use a combinator to get a callback like effect.

nats-jetstream/src/nats/jetstream/consumer/pull.py
caspervonb: Removed the message handler, might add it back in but feeling like an anti-pattern.

nats-jetstream/src/nats/jetstream/stream.py
caspervonb: Fixed.

nats-jetstream/src/nats/jetstream/consumer/pull.py
caspervonb: I'm thinking we shouldn't swallow unexpected errors.

nats-jetstream/src/nats/jetstream/api/__init__.py
caspervonb: So am not exporting everything, because it's generated code, so only re-exporting what we actually use.

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
caspervonb: Yeah tools/fetch_schemas.py

nats-jetstream/pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[build-system]
+requires = ["setuptools>=61.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "nats-jetstream"
+version = "0.0.0"
+description = 'Python library for managing NATS servers for development and testing'
+readme = "README.md"
+requires-python = ">=3.11"
```
dennari, line 10: This should probably be >=3.13 since nats-core has that?

pyproject.toml
```diff
@@ -4,23 +4,25 @@ dev = [
...
-
-[tool.ty.src]
-exclude = ["nats/"]
+[tool.mypy]
+files = ["nats/src"]
+python_version = "3.7"
+ignore_missing_imports = true
+follow_imports = "silent"
+show_error_codes = true
+check_untyped_defs = false
```
dennari: Maybe now that `ty` is in beta it would be a better option than mypy?

pyproject.toml
```diff
@@ -4,23 +4,25 @@ dev = [
     "pytest-asyncio>=0.21.0",
     "pytest-cov>=7.0.0",
     "pytest-xdist>=3.0.0",
-    "ty>=0.0.0a6",
+    "mypy>=1.0.0",
     "ruff>=0.1.0",
-    "codespell>=2.2.0",
+    "flake8>=7.0.0",
```
dennari: Isn't flake8 obsolete with ruff?

nats-jetstream/src/nats/jetstream/api/client.py
```diff
@@ -0,0 +1,416 @@
...
+        error_code=err_code,
+        description=description,
+    )
+
+
+def is_error_response(data: Any) -> TypeGuard[ErrorResponse]:
+    return isinstance(data, dict) and "error" in data
+
+
+def check_response[ResponseT](data: Any, expected_type: type[ResponseT]) -> tuple[bool, set[str], set[str]]:
```
dennari, line 92: Should this be `ReponseT: dict`?

nats-jetstream/schemas/jetstream/api/v1/account_info_response.json
dennari: But what's the point of including the json schema files in this repository? Isn't it enough if the resulting generated Python types are committed?

nats-jetstream/tools/generate_types.py
dennari, line 1: Are you familiar with https://github.com/koxudaxi/datamodel-code-generator/? I've had really good experiences with it before. Also, I'm wondering whether in this case Pydantic would be a better fit for the generated types than `TypedDict`? It would allow us to express all the constraints too (e.g. `0 <= num_replicas <=5`) and comes with JSON serde out of the box.

Just to see how it would fare in this case I ran the following in `jetstream/api/v1`:

```bash
datamodel-codegen \
    --input definitions.json \
    --input-file-type jsonschema \
    --output-model-type pydantic_v2.BaseModel \
    --output definitions.py \
    --formatters ruff-check \
    --formatters ruff-format \
    --use-annotated \
    --use-standard-collections \
    --use-one-literal-as-default \
    --use-specialized-enum \
    --target-pydantic-version 2.11 \
    --target-python-version 3.14 \
    --type-overrides '{"GolangInt": "int", "GolangInt32": "int", "GolangInt64": "int", "GolangUint32": "int", "GolangUint64": "int", "GolangFloat32": "float", "GolangTime": "pydantic.AwareDatetime", "GolangDuration": "str"}'
```

To me the result, even if not perfect, looks really promising.

caspervonb: I've tried the various existing tools for generation, including datamodel-codegen doesn't pan out well for our json-schemas.

If we were to take a dependency, `struct` would probably be the best fit for the actual models.

PR #738, opened 2025-09-25
Add `time` field to `RawStreamMsg`

mjvankampen:
Fixes a todo that was still open.


mjvankampen:
@caspervonb should be better now. As soon as python 3.8 support is dropped the helper function can be removed.

caspervonb:
Thank you very much @mjvankampen, merged

PR #745, opened 2025-10-07
uv + nats-server package changes

wallyqs:
Follow up from a couple of the previous PRs:

- Adds `aiohttp`, `aiofiles`, `nkeys`, `fast_mail_parser` to development group
- Move `nats-server` package from `import nats.server` into `import nats.testing.server` instead


pyproject.toml
```diff
@@ -2,15 +2,19 @@
...
-    "pytest-cov>=7.0.0",
+    "pytest-cov>=4.0.0,<7.0.0",
     "pytest-xdist>=3.0.0",
     "mypy>=1.0.0",
     "yapf>=0.40.0",
     "ruff>=0.1.0",
     "isort>=5.0.0",
-    "flake8>=7.0.0",
+    "flake8>=5.0.0,<7.0.0",
+    "aiohttp",
```
caspervonb: This is a workspace manifest, these dependencies are in nats/pyproject.toml.

https://github.com/nats-io/nats.py/blob/main/nats/pyproject.toml#L33-L36

nats-server/pyproject.toml
```diff
@@ -1,40 +0,0 @@
-[build-system]
```
caspervonb, line 1: Comment on the file being removed, this is, intentionally its own package in a workspace.

pyproject.toml
```diff
@@ -2,15 +2,19 @@
 dev-dependencies = [
     "pytest>=7.0.0",
     "pytest-asyncio>=0.21.0",
-    "pytest-cov>=7.0.0",
+    "pytest-cov>=4.0.0,<7.0.0",
```
caspervonb: Why pinning this?

Jarema:
@wallyqs why we want to move it back to nats package?
It's a dev dependency for us, and very rarely for users.
In many clients we try to split those. And as javascript did recently - even split core nats from js, kv and object store.

nats/src/nats/testing/server/__init__.py
```diff
@@ -24,8 +24,12 @@
 import os
 import re
 import socket
-import tempfile
-from typing import Self
+import sys
+
+if sys.version_info >= (3, 11):
```
caspervonb: Minimum version of this on the greener side, since it's a new package.

wallyqs:
Updated to keep the workspace and instead only changed the import to `nats.testing.server`.
I think we need to add a matrix step on CI where we run tests with these deps: `aiohttp`, `aiofiles`, `nkeys` (websockets + nkeys tests)

PR #758, opened 2025-10-17
Add type-check job to workflow

caspervonb:


pyproject.toml
```diff
@@ -4,19 +4,17 @@ dev-dependencies = [
...
-python_version = "3.7"
-ignore_missing_imports = true
-follow_imports = "silent"
-show_error_codes = true
-check_untyped_defs = false
+[tool.ty.environment]
+root = ["nats-server/src"]
+
+[tool.ty.src]
+exclude = ["nats/"]
```
caspervonb, line 17: Too many errors, enabling it for the future rather than the past.

pyproject.toml
```diff
@@ -4,19 +4,17 @@ dev-dependencies = [
     "pytest-asyncio>=0.21.0",
     "pytest-cov>=7.0.0",
     "pytest-xdist>=3.0.0",
-    "mypy>=1.0.0",
+    "ty>=0.0.0a6",
     "ruff>=0.1.0",
     "codespell>=2.2.0",
 ]
 workspace = { members = ["nats", "nats-server"] }
```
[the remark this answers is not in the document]
caspervonb: Old configuration doesn't actually passs, 60 different errors in nats/src, 500 plus if we add nats/tests and nats/examples.

pyproject.toml
wallyqs: the past is the present ;)

wallyqs (approved):
LGTM

pyproject.toml
caspervonb: Yeah tried fixing them but, it'll be breaking to add/change typedefs etc.

PR #775, opened 2025-11-04
Fix EOF processing while client is connecting

caspervonb:
Re-raises #719 against main as it was merged into a v2 branch.


wallyqs (approved):
LGTM

PR #780, opened 2025-11-07
Add `consumer_limits` support to `StreamConfig`

### LINKED ISSUE #760: Missing StreamConsumerLimits props from StreamConfig
### Proposed change

Finish implementation of the inherited consumer limits in streams: https://github.com/nats-io/nats-architecture-and-design/issues/242.

### Use case

Multi-tenant scenarios when we can't trust end-users for DDOS the broker.

### Contribution

_No response_

caspervonb:
Partially resolves https://github.com/nats-io/nats.py/issues/760


philpennock (changes requested):
I concur with both AIs.  Fix that issue and I would approve.

philpennock (approved):

PR #796, opened 2025-12-17
Normalize ISO timestamp fractional seconds to 6 digits for Python <3.11

purvish14:
### Problem
When using NATS object store in Go, the ModTime field is serialized with variable microsecond precision (e.g., 5 digits like 20373). Python versions before 3.11 require exactly 3 or 6 digits for fractional seconds in datetime.fromisoformat(), causing parsing failures:
```
Invalid isoformat string: '2025-12-15T16:55:37.20373+00:00'
```

### Root Cause
The existing `_python38_iso_parsing()` workaround only truncates fractional seconds to 6 digits but doesn't pad shorter values. This fails for Go timestamps with 1, 2, 4, or 5 digit precision.

### Fix
Normalize fractional seconds to exactly 6 digits by padding with trailing zeros when needed.

### DIFF
--- a/nats-core/pyproject.toml
+++ b/nats-core/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
-requires = ["setuptools>=61.0"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.9.28,<0.10.0"]
+build-backend = "uv_build"
 
 [project]
 name = "nats-core"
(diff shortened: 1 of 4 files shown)


klepalinsoft:
I actually stumbled upon this issue today, and the fix I was about to propose as a PR is exactly the same.

```python
frac = frac[:6].ljust(6, "0")
```

krisztian-fl:
Bump on this, it is breaking durable consumers on Python <3.11.

PR #801, opened 2025-12-19
Rename from nats-client to nats-core

caspervonb:
This is the package name we have on pypi, and reflects what the other clients do.


Jarema (approved):
LGTM

PR #807, opened 2026-01-14
Avoid $JS.API.STREAM.NAMES call since stream name is known

liaoya:
If `stream` parameter is not passed, `JetStreamContext.subscribe` will call `JetStreamManager.find_stream_name_by_subject` with `$JS.API.STREAM.NAMES` to get the stream name. This is a redundant operation. In our environment, `$JS.API.STREAM.NAMES` is not allowed.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,6 +6,8 @@ on:
     pull_request:
         branches:
             - "*"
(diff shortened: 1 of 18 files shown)


caspervonb (approved):
lgtm, thank you.

PR #810, opened 2026-01-26
Add jetstream direct get timestamp parsing

### LINKED ISSUE #809: `js.get_msg(..., direct=True)` does not populate `RawStreamMsg.time`
### Proposed change

Parse "Nats-Time-Stamp" header into `RawStreamMsg.time`

In `js.get_msg(..., direct=False)`, time is populated by `RawStreamMsg.from_response()`, but `from_response()` isn't used for direct get because the time is in `response.headers["Nats-Time-Stamp"]` instead of the `response["time"]`

A way to do this would be:
- pull the RawStreamMsg._python38_iso_parsing classmethod out into a helper function
- use that helper function in JetstreamManager._lift_msg_to_raw_msg:

```diff
  @classmethod
  def _lift_msg_to_raw_msg(self, msg) -> api.RawStreamMsg:
      ...
      seq = msg.headers.get("Nats-Sequence")
      if seq:
          raw_msg.seq = int(seq)
      raw_msg.data = msg.data
      raw_msg.headers = msg.headers
+     if time_string := msg.headers.get("Nats-Time-Stamp"):
+         raw_msg.time = parse_iso_dt(time_string)

      return raw_msg
```

### Use case

Anyone who wants to use the message timestamp from a direct get currently has to parse it out manually, which is unexpected behaviour and not a great experience:

```python
from datetime import UTC, datetime

nc = nats.connect()
js = nc.jetstream()
msg = await js.get_msg(..., direct=True)
if msg.time is None and msg.headers and (dt := msg.headers.get("Nats-Time-Stamp")):
    msg.time = datetime.fromisoformat(dt).astimezone(UTC)

# parsing above requires python 3.11+
# for <3.11 need to drop ns from dt string first and use datetime.timezone.utc
```

### Contribution

Sure

oliverlambson:
Closes #809

### DIFF
--- a/nats/src/nats/js/api.py
+++ b/nats/src/nats/js/api.py
@@ -446,6 +446,7 @@ class StreamInfo(Base):
     sources: Optional[List[StreamSourceInfo]] = None
     cluster: Optional[ClusterInfo] = None
     did_create: Optional[bool] = None
+    created: Optional[datetime.datetime] = None
 
     @classmethod
     def from_response(cls, resp: Dict[str, Any]):
@@ -454,6 +455,8 @@ class StreamInfo(Base):
         cls._convert(resp, "mirror", StreamSourceInfo)
         cls._convert(resp, "sources", StreamSourceInfo)
         cls._convert(resp, "cluster", ClusterInfo)
+
+        cls._convert_utc_iso(resp, "created")
         return super().from_response(resp)
 
 
--- a/nats/tests/test_js.py
+++ b/nats/tests/test_js.py
@@ -1216,6 +1216,16 @@ class JSMTest(SingleJetStreamServerTestCase):
         await nc.close()
 
     @async_test
+    async def test_stream_info_created_is_datetime(self):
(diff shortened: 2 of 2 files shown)


nats/src/nats/js/api.py
```diff
@@ -871,3 +859,15 @@ def is_link(self) -> bool:
...
+
+
+def _parse_iso_dt(time_string: str) -> datetime.datetime:
+    if sys.version_info < (3, 11):
+        # Replace Z with UTC offset
+        time_string = time_string.replace("Z", "+00:00")
+        # Trim fractional seconds to 6 digits
+        date_part, frac_tz = time_string.split(".", 1)
+        frac, tz = frac_tz.split("+")
+        frac = frac[:6]  # keep only microseconds
```
oliverlambson: i kept this as-is, but it's worth noting the proposed changes to this logic in #796 (i.e., this may contain a pre-existing bug)

PR #829, opened 2026-02-20
Add server pool management and reconnect handler

caspervonb:


nats/src/nats/aio/client.py
```diff
@@ -1457,10 +1532,44 @@ async def _attempt_reconnect(self) -> None:
...
+                    if max_reconnect > 0:
+                        eligible = [s for s in self._server_pool if s.reconnects <= max_reconnect]
+                    else:
+                        eligible = list(self._server_pool)
+
+                    if len(eligible) == 0:
+                        raise errors.NoServersError
+
+                    server_snapshot = [Server(uri=s.uri, reconnects=s.reconnects) for s in eligible]
+                    selected, callback_delay = self._reconnect_to_server_handler(server_snapshot, self._server_info)
```
wallyqs: makes sense to pass a copy of `_server_info` here instead?

wallyqs: maybe need to catch the error thrown in the handler, don't think it will be caught right now and will cause the task to exit

caspervonb: Sure, done.

caspervonb: Spot on, skips and calls error cb

PR #832, opened 2026-02-24
Add ordered consumers

caspervonb:


nats-jetstream/src/nats/jetstream/consumer/__init__.py
```diff
@@ -383,9 +383,59 @@ def from_response(cls, data: api.ConsumerInfo, *, strict: bool = False) -> Consu
...
+    filter_subjects: list[str] | None = None
+    """Filter the stream by multiple subjects."""
+
+    deliver_policy: DeliverPolicy = "all"
+    """The point in the stream from which to receive messages."""
+
+    opt_start_seq: int | None = None
+    """Start sequence used with the DeliverByStartSequence deliver policy."""
+
+    opt_start_time: str | None = None
```
caspervonb:
```suggestion
    opt_start_time: datetime | None = None
```

caspervonb:
PTAL @claude

caspervonb:
Protocol issues were fixed up in main @claude

nats-jetstream/src/nats/jetstream/consumer/ordered.py
```diff
@@ -0,0 +1,513 @@
...
+        consumer_config = oc._get_consumer_config()
+        consumer = await stream.create_or_update_consumer(consumer_config)
+        oc._current_consumer = consumer  # type: ignore[assignment]
+
+        return oc
+
+    @property
+    def name(self) -> str:
+        if self._current_consumer is None:
+            return f"{self._name_prefix}_0"
```
philpennock, line 85: Why is this safe? Is it just ephemeral if no name is set, so we can use "whatever"?

nats-jetstream/src/nats/jetstream/consumer/ordered.py
```diff
@@ -0,0 +1,513 @@
...
+            # No messages consumed yet - use user's deliver policy
+            opt_start_seq = self._cfg.opt_start_seq
+            deliver_policy = self._cfg.deliver_policy
+            opt_start_time = self._cfg.opt_start_time
+
+            # For certain policies, opt_start_seq should not be set
+            if deliver_policy in ("last_per_subject", "last", "new", "all"):
+                opt_start_seq = None
+            elif deliver_policy == "by_start_time":
+                opt_start_seq = None
```
philpennock, line 305: Why is line 304 not simply another entry in the set, if we're taking the same action, or should we be doing something different in this branch?

PR #833, opened 2026-02-24
Add consumer pinning

caspervonb:


caspervonb:
Addressed in 0b7936b:
- `PullConsumer.get_info()` now refreshes from the server via `self._stream.get_consumer_info(self._info.name)`.
- `ConsumerUnpinResponse.error` corrected to `NotRequired[Error]`.
- Format issues resolved by `ruff format`.

PTAL @claude

caspervonb:
Amended — scope of review fixes on this PR is now:
- `PullConsumer.get_info()` refreshes from the server.
- Three unused `consumer` variable assignments removed from the new priority-group tests.

The `ConsumerUnpinResponse.error` observation is a real pre-existing defect at the schema layer (`consumer_unpin_response.json` inlines `error_response` as a property instead of using `oneOf` like `consumer_delete_response.json`). Fixing that and regenerating `types.py` produces drift against main's checked-in version that's not about this PR, so it belongs in a separate schema-cleanup PR. Will file a follow-up issue.

PTAL @claude

PR #854, opened 2026-04-03
Protect client callbacks from exceptions

### LINKED ISSUE #204: Client stops trying to connect if `disconnected_cb` raises exception
If the callback passed to `disconnected_cb` raises an exception, then the NATS client will never reconnect to the cluster.

While this behavior could be classified as "user error" (that callback should not throw!), it would be nice to handle this gracefully: 
 * Print an error message so the user knows the callback has raised an exception (currently no message) 
 * Keep trying to reconnect (currently, no attempt to reconnect)

Attaching a simple script that reproduces the behavior: [callback_exception.py.txt](https://github.com/nats-io/nats.py/files/6942266/callback_exception.py.txt)

See instructions at the top of the file.

Expected behavior:

```
📤 Message published
📥 Message received
📤 Message published
📥 Message received
📤 Message published
📥 Message received
>>>> Server crash-restart
nats: encountered error
NoneType: None
🔴 Disconnected!
🟢 Reconnected!
📤 Message published
📥 Message received
📤 Message published
📥 Message received

```

Actual behavior (with `DISCONNECT_CALLBACK_EXCEPTION=True`):

```
📤 Message published
📥 Message received
📤 Message published
📥 Message received
📤 Message published
📥 Message received
>>>> Server crash-restart
nats: encountered error
NoneType: None
🔴 Disconnected!
📤 Message published
📤 Message published
📤 Message published
📤 Message published
```

Notice no errors, no attempts to reconnect, and no message delivered forever after

### LINKED ISSUE #287: nc.flush times out after keyboard interrupt

caspervonb:
Fixes #204
Fixes #287


caspervonb:
PTAL @claude

caspervonb:
> Note: Python 3.7 CancelledError

On Python 3.7, asyncio.CancelledError inherits from Exception (changed to BaseException in 3.8). In subscription.py, the outer except asyncio.CancelledError is checked before except Exception, so the message handler's CancelledError is handled correctly. However, if the error_cb itself raises a CancelledError on Python 3.7, it will be caught and logged at line 323 instead of propagating. This is an unlikely edge case and the existing comment in the previous review notes this — just confirming it hasn't changed.

3.7 is EOL so, we don't support it so that's fine IMO.

caspervonb:
PTAL @claude

wallyqs:
@claude can you fix conflicts

wallyqs:
rebase and fix conflicts

caspervonb:
@claude please review again

caspervonb:
@claude please review again

nats/tests/test_client.py
```diff
@@ -1169,6 +1169,50 @@ async def err_cb(e):
...
+        nc = NATS()
+
+        async def bad_error_cb(e):
+            raise RuntimeError("broken error callback")
+
+        await nc.connect(error_cb=bad_error_cb)
+
+        # Trigger an error callback via a permissions violation error message.
+        # This should not raise even though the error_cb throws.
+        await nc._process_err('Permissions Violation for Subscription to "test"')
```
renerocksai: I think this doesn't reach the branch it's aiming for. The parser lowercases the error text before dispatching:

```python
# nats/src/nats/protocol/parser.py:137
emsg = err_msg[0].decode().lower()
await self.nc._process_err(emsg)
```

and `PERMISSIONS_ERR` is `"permissions violation"` (lowercase). Calling `_process_err` directly with the server's original casing bypasses that normalisation, so `PERMISSIONS_ERR in m` is `False`, the permissions branch is skipped, and control falls through to `asyncio.create_task(self._close(...))` — which the test never awaits.

I ran both spellings against the pre-fix code to check:

| `err_msg` passed to `_process_err` | `error_cb` invocations | `_process_err` raises? |
|---|---|---|
| `'Permissions Violation for Subscription to "test"'` | 0 | no |
| `'permissions violation for subscription to "test"'` | 1 | yes |

So as written I believe this passes with the `try/except` reverted. Lowercasing the string should fix it, and asserting that the callback was actually invoked would keep it honest.

Separately: the name mentions reconnect, but the permissions branch `return`s before any reconnect happens — something like `test_error_callback_exception_does_not_break_process_err` might describe it more accurately.

nats/tests/test_client.py
```diff
@@ -1270,6 +1314,34 @@ async def receiver_cb(msg):
...
+        async def bad_error_cb(e):
+            raise RuntimeError("bad error callback")
+
+        await nc.connect(error_cb=bad_error_cb)
+        await nc.subscribe("test", cb=bad_handler)
+        await nc.publish("test", b"hello")
+        await nc.flush()
+        await asyncio.sleep(0.1)
+
+        # Subscription should still be alive (task not crashed)
```
renerocksai: A few tightenings for this test:

1. **Nothing asserts what the comment claims.** I checked against the pre-fix code and the test still completes cleanly, so I think it passes with the `try/except` in `subscription.py` reverted. What actually happens without the fix:

   ```
   handler invocations  : 1      (not 2)
   _wait_for_msgs done  : True
   task exception       : RuntimeError('bad error callback')
   ```

   The subscription really is dead, but nothing here notices — `publish`/`flush`/`close` all succeed either way. Having `bad_handler` count its invocations and asserting `2` would pin it; capturing the subscription and asserting `sub._wait_for_msgs_task.done()` is `False` would work too.

2. **`await asyncio.sleep(0.1)`** here and on line 1341 is timing-based. If the handler is counting invocations anyway, an `asyncio.Event` plus `asyncio.wait_for` removes both sleeps and the CI flake risk — line 909 in this file already uses that idiom.

3. Optional: `assertLogs("nats.aio.client", level="ERROR")` would pin *swallowed and logged* rather than just *didn't raise*. I don't see `assertLogs` used anywhere else in this suite though, so purely take-or-leave.

nats/src/nats/aio/client.py
```diff
@@ -799,13 +802,22 @@ async def _close(self, status: int, do_cbs: bool = True) -> None:
...
+                try:
+                    await self._error_cb(e)
+                except Exception:
+                    _logger.error("nats: error in error callback", exc_info=True)
 
         if do_cbs:
             if self._disconnected_cb is not None:
-                await self._disconnected_cb()
+                try:
+                    await self._disconnected_cb()
```
renerocksai: Optional, but I think this is the main idiomatic point in the PR: this same four-line block is repeated 21 times across `client.py` (15 of them for `_error_cb`), plus once in `subscription.py`. A small helper would collapse each site to a single line:

```python
async def _run_callback(self, cb, *args, name: str) -> None:
    try:
        await cb(*args)
    except Exception:
        _logger.exception("nats: error in %s callback", name)
```

so this block becomes:

```python
if do_cbs:
    if self._disconnected_cb is not None:
        await self._run_callback(self._disconnected_cb, name="disconnected")
    if self._closed_cb is not None:
        await self._run_callback(self._closed_cb, name="closed")
```

Beyond diff size, the reason I'd suggest it is that it turns the invariant from conventional into enforceable — a new callback site can't be added unprotected. That's not hypothetical here: `_lame_duck_mode_cb` was added while this PR was open and needed a follow-up commit to get the same treatment, and the two `_error_cb` calls in `nats/src/nats/js/client.py` are still uncovered. With a helper those would be one-liners.

Small thing while you're in here: inside an `except` block, `_logger.exception(...)` is the canonical spelling for `_logger.error(..., exc_info=True)` — identical behaviour, just less to repeat 21 times.

renerocksai (commented):

A few things I think are worth a look:

1. `test_error_callback_exception_does_not_break_reconnect` doesn't appear to reach the branch it targets — the parser lowercases the error text before dispatch, so the capitalised string never matches `PERMISSIONS_ERR` and `error_cb` is never invoked. Details inline.
2. `test_error_callback_exception_does_not_break_subscription` has no assertions, and I believe it passes with the fix reverted. Details inline.
3. Two `_error_cb` call sites in `nats/src/nats/js/client.py` look like they're still unprotected: `_conn._error_cb(ecs)` at line 837 (ordered-consumer sequence mismatch) and `_conn._error_cb(err)` at line 888 in `recreate_consumer`. The second seems the more awkward of the two — `recreate_consumer` is launched as a bare `asyncio.create_task(...)` at line 879 with no reference retained, so a raising `error_cb` there would surface only as a "Task exception was never retrieved" warning at GC time, with the consumer silently never recreated. That reads like the same failure mode as #287. Entirely fair to call out of scope if you're deliberately keeping this PR to `aio/`, but thought it worth mentioning.

nats/tests/test_client.py
caspervonb: Fixed in 8aa751fb: the test now lowercases the error text, asserts the error callback ran once, and is renamed test_error_callback_exception_does_not_break_process_err.

nats/tests/test_client.py
caspervonb: Done in 8aa751fb: the handler counts invocations and the test asserts 2, waits on an asyncio.Event instead of sleeping, checks the subscription task is still alive, and uses assertLogs to pin the logged error.

nats/src/nats/aio/client.py
caspervonb: Done in 8aa751fb: added _run_callback/_run_error_callback helpers using _logger.exception and routed every callback site in client.py, subscription.py and js/client.py through them.

caspervonb:
PTAL @claude

PR #855, opened 2026-04-03
Disable WebSocket max message size

caspervonb:
Supersedes https://github.com/nats-io/nats.py/pull/470
Protocol enforces message sizes (max payload, tedious to deal with two limits).


wallyqs (approved):
LGTM

PR #869, opened 2026-04-08
Add lame duck mode handling for graceful reconnection

### LINKED ISSUE #868: missing handling of lame duck mode: causes client failures during rolling restarts
## Observed behavior

nats.py does not handle lame duck mode. When a NATS server enters lame duck mode (e.g. during a rolling restart), it sends `INFO {"ldm": true}` to connected clients. However, `_process_info` only inspects `connect_urls` for cluster discovery — the `ldm` field lands in `self._server_info` and is never acted on.

As a result, clients sit idle until the server force-closes the connection, producing `unexpected EOF` and `Connection refused` errors instead of proactively migrating to a healthy node.

In a 3-node cluster on k8s (official helm chart), every rolling restart cycle produces ~3 minutes of errors across all Python clients, even though 2 of 3 nodes are healthy the entire time.

What happens today:

1. Server receives SIGTERM, enters lame duck mode
2. Server sends `INFO {"ldm": true}` to connected clients
3. nats.py ignores the `ldm` field
4. Server waits out `lame_duck_grace_period`, then starts closing connections
5. Client gets unexpected EOF, error callback fires
6. Client shuffles server pool, may try the dying server first — `ConnectionRefusedError`
7. Eventually reconnects to a healthy node

Logs from prod:

```
18:37:48 [ERROR] nats:error error=nats: unexpected EOF
18:37:48 [WARN]  nats:disconnected
18:37:48 [ERROR] nats:error error=[Errno 111] Connection refused
18:37:48 [INFO]  nats:reconnected
```

Each client is unavailable during the reconnect cycle (can't publish or receive), and this happens 3 times across the rolling restart.

## Expected behavior

When the server sends `INFO {"ldm": true}`, the client should proactively reconnect to a healthy node before the server closes the connection, avoiding errors entirely.

The Go client ([nats.go](https://github.com/nats-io/nats.go)) already supports this via `LameDuckModeHandler` and `ForceReconnect()`:

```go
nc, _ := nats.Connect(servers,
    nats.LameDuckModeHandler(func(nc *nats.Conn) {
        nc.ForceReconnect()
    }),
)
```

Proposed changes for nats.py:

1. **Detect LDM in `_process_info`** — check for `ldm: true` in async INFO messages (not on initial connect). The data is already there, just needs a conditional.
2. **Auto-reconnect by default** — when LDM is detected and `allow_reconnect` is true, automatically trigger reconnection. The Go client requires users to wire this up themselves, but auto-reconnecting is the right default for Python — most users won't know to set up a handler, and "do nothing" is never the right response to LDM.
3. **Add `lame_duck_mode_handler` callback** — for users who want custom behavior, following the existing callback pattern (`error_cb`, `disconnected_cb`, etc.). When set, it replaces the auto-reconnect.
4. **Add `force_reconnect()` method** — public method to proactively trigger reconnection, useful from the LDM handler but also generally useful on its own.
5. **Skip the LDM server during reconnection** — mark the current server when LDM is detected so `_select_next_server` doesn't try reconnecting to the server that's shutting down.

```python
nc = await nats.connect(servers)  # auto-reconnects on LDM by default

# custom handler if needed
nc = await nats.connect(servers, lame_duck_mode_handler=my_handler)

# force reconnect from anywhere
await nc.force_reconnect()
```

## Server and client version

- **Server**: nats-server v2.12.4 (helm chart `nats/nats` v2.12.4, image `nats:2.12.4-alpine`, 3-node cluster on k8s)
- **Client**: nats-py >=2.14.0, <3.0.0

## Host environment

Kubernetes (Linux containers), but this is not environment-specific — lame duck mode is ignored on all platforms.

## Steps to reproduce

1. Start a multi-node NATS cluster
2. Connect a nats.py client to one of the nodes
3. Send SIGTERM to the connected node (or trigger a rolling restart)
4. Observe that the client does not react to the `INFO {"ldm": true}` message
5. Client eventually gets disconnected with `unexpected EOF` and must reconnect reactively

oliverlambson:
Closes #868

When a NATS server enters lame duck mode during a rolling restart, it sends `INFO {"ldm": true}` to connected clients. Previously nats.py ignored this field, causing clients to wait until the server forcibly closed the connection before reconnecting -- producing unnecessary errors and downtime.

## Changes

- Detect `ldm: true` in `_process_info` and invoke `lame_duck_mode_cb` callback (matching Go client behavior -- no auto-reconnect, user decides)
- Add `lame_duck_mode_cb` callback option on `connect()` for custom LDM handling
- Add `force_reconnect()` public method for proactive reconnection (can be called from the callback)
- Add `NATSD.send_signal()` test helper

## API

```python
nc = NATS()

async def handle_lame_duck():
    await nc.force_reconnect()

# Custom handler for lame duck mode
await nc.connect(
    servers,
    lame_duck_mode_cb=handle_lame_duck,  # new arg
)

# Force reconnect from anywhere
await nc.force_reconnect()  # new public method
```

## Tests

- [x] `test_lame_duck_callback` -- callback fires on SIGUSR2, no auto-reconnect
- [x] `test_lame_duck_callback_with_force_reconnect` -- callback calls `force_reconnect()`, client moves to other server
- [x] `test_force_reconnect` -- public method works standalone
- [x] Existing test suite passes


oliverlambson:
Tested this against a rolling restart of a 3-node NATS cluster in production. LDM detection is working -- every disconnect was preceded by a lame duck mode callback, and clients proactively reconnected before the server closed the connection.

One thing I noticed: there are still occasional `[Errno 111] Connection refused` errors when the reconnect pool rotation lands on a server that's also shutting down or hasn't fully started yet. The client recovers immediately on the next attempt, so no user impact, but it's not completely zero-error.

I checked the nats.go implementation to see if it handles this differently -- it doesn't. `selectNextServer()` just rotates the pool without any LDM-aware filtering, so the same race window exists there too. This PR is at parity with the Go client.

Before this change, every pod restart produced unexpected EOF + connection refused errors for all connected clients. Now, clients detect LDM and migrate proactively -- the occasional connection refused during pool rotation is the only remaining noise (since it's parity with nats.go, i think that's acceptable).

caspervonb:
Nice! will review shortly.

nats/src/nats/aio/client.py
```diff
@@ -2001,6 +2019,12 @@ async def _process_info(self, info: Dict[str, Any], initial_connection: bool = F
                 if not initial_connection and connect_urls and self._discovered_server_cb:
                     await self._discovered_server_cb()
 
+        if not initial_connection and info.get("ldm", False):
+            if self._lame_duck_mode_cb is not None:
+                await self._lame_duck_mode_cb()
```
caspervonb, line 2024: As far as I can tell, Go doesn't do force_reconnect, leaves that to the user to do in the callback we should probably do the same.

caspervonb (changes requested):

caspervonb:
PTAL @claude

nats/src/nats/aio/client.py
oliverlambson: my mistake, pr (& description) updated

caspervonb (approved):
LGTM, will merge shortly just going to give @wallyqs a chance to object.

wallyqs:
let's cleanup the PR_DESCRIPTION.md‎ from the branch before merging 🙏

oliverlambson:
> let's cleanup the PR_DESCRIPTION.md‎ from the branch before merging 🙏 

whoops, fixed

oliverlambson:
@caspervonb rebased on to latest main fyi

PR #872, opened 2026-04-13
Add close() and context manager to Consumer protocol

caspervonb:
## Summary
- Adds `close()`, `__aenter__()`, and `__aexit__()` to the `Consumer` Protocol so consumers can be used as async context managers through the protocol type.

## Test plan
- [ ] Verify `uv run ruff check` passes
- [ ] Verify existing tests pass (`uv run pytest nats-jetstream/tests`)


philpennock (approved):
But please first reply to Claude

caspervonb:
PTAL @claude

PR #878, opened 2026-04-16
Overhaul README to reflect new package layout

caspervonb:


README.md
```diff
@@ -1,246 +1,139 @@
-# NATS.py - Python Client and Tools
+# NATS - Python Clients and Tools
```
wallyqs, line 1: put these parameters to the original names, not needed to rename them
```
| `connect_timeout` | `timeout` | 
| `reconnect_time_wait` | `reconnect_time_wait` | Same name, now `float` instead of `int` |
| `max_reconnect_attempts` | `reconnect_max_attempts` | Renamed; default changed: 60 → 10 |
| `dont_randomize` | `no_randomize` | Renamed |
```

wallyqs (changes requested):

README.md
wallyqs: same with these options, look for the original ones

```
    nkey="SUAGIEYF...",         # was nkeys_seed
    jwt=Path("file.creds"),     # was user_credentials
```

wallyqs:
is `msg.respond()` removed in the nats-core package?

caspervonb:
No msg.respond isn't in at this time.

PR #879, opened 2026-04-17
Add connect to nats.client exports

caspervonb:


nats-core/src/nats/client/__init__.py
```diff
@@ -1663,6 +1663,7 @@ async def connect(
 
 
 __all__ = [
+    "connect",
```
wallyqs, line 1666: @claude can you add an example of code that connects to local nats-server using this exported name

caspervonb: Out of scope for this pull request, thanks claude.

wallyqs:
plz rebase

renerocksai (approved):
LGTM

aricart (approved):

philpennock (approved):

PR #882, opened 2026-04-17
Strip whitespace from nkeys seed input

### LINKED ISSUE #417: If file given with nkeys_seed parameter contains a newline, the  nkeys.from_seed() fails with nkeys: invalid seed
When using the `nk` utility as described in https://docs.nats.io/running-a-nats-service/configuration/securing_nats/auth_intro/nkey_auth#generating-nkeys-and-configuring-the-server  it seems ok to cut/paste the seed into a file and load it with the `nkeys_seed` parameter of `nats.py`. However, most editors (and shell redirect) will add a newline at the end of the seed, and that will make `nkeys.from_seed()` in https://github.com/nats-io/nats.py/blob/main/nats/aio/client.py#L515 fail with `nkeys: invalid seed ` . This could be more robust if the file reading included a stripping of whitespace at end of line before decoding and turning it into a bytearray. Seems like the tests will not cover it either since those include seed-strings directly and not read from file. 

In order for nets.py to work with user nkeys generated  with the `nk` utility the file could be generated like this:

`printf "%s"  $(nk -gen user) > file`

And to get the pubkey for using in the server config:

`nk -pubout -inkey file`

Perhaps the file loading code behind the nkeys_seed parameter could sanitize the file before load ?

caspervonb:
Fixes https://github.com/nats-io/nats.py/issues/417

### DIFF
--- a/nats/src/nats/aio/client.py
+++ b/nats/src/nats/aio/client.py
@@ -692,15 +692,12 @@ class Client:
         import nkeys
 
         def _get_nkeys_seed() -> nkeys.KeyPair:
-            import os
-
             if self._nkeys_seed_str:
-                seed = bytearray(self._nkeys_seed_str.encode())
+                seed = bytearray(self._nkeys_seed_str.strip().encode())
             else:
                 creds = self._nkeys_seed
                 with open(creds, "rb") as f:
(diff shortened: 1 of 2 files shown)


caspervonb:
PTAL @claude

wallyqs (approved):

PR #886, opened 2026-04-17
Use base64url without padding for nonce signatures

caspervonb:
ADR-14 and Go (base64.RawURLEncoding) specify URL-safe base64 without padding for the nonce signature sent in CONNECT.

The nkey and JWT signature handlers were emitting standard base64 (+/ with = padding), which strict servers reject.

### DIFF
--- a/nats-core/src/nats/client/__init__.py
+++ b/nats-core/src/nats/client/__init__.py
@@ -1336,7 +1336,7 @@ def _setup_nkey_auth(
     def signature_handler(nonce: str) -> bytes:
         kp = nkeys.from_seed(seed_bytes)
         sig = kp.sign(nonce.encode())
-        return base64.b64encode(sig)
+        return base64.urlsafe_b64encode(sig).rstrip(b"=")
 
     return public_key_handler, signature_handler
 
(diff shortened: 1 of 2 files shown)


nats-core/src/nats/client/__init__.py
```diff
@@ -1336,7 +1336,7 @@ def public_key_handler() -> str:
     def signature_handler(nonce: str) -> bytes:
         kp = nkeys.from_seed(seed_bytes)
         sig = kp.sign(nonce.encode())
-        return base64.b64encode(sig)
+        return base64.urlsafe_b64encode(sig).rstrip(b"=")
```
wallyqs, line 1339: this an issue in the nats-py client as well?

caspervonb: Yes — same bug at `nats/src/nats/aio/client.py:613, 633, 715`. Will fix in a separate PR.

caspervonb:
Test helpers `nkey_handlers()` and `jwt_handlers()` updated to use base64url in 3930dfb.

PTAL @claude

nats-core/src/nats/client/__init__.py
caspervonb: ADR uses the term "base64url", but implementation is base64.RawURLEncoding (meaning stripped of padding).

wallyqs, line 1339: @claude can you add test coverage for this change

wallyqs (approved):

PR #891, opened 2026-04-17
Allow user and password to be callables for credential refresh

### LINKED ISSUE #714: Support auth callback for auth callout with in-house JWT systems
### Proposed change

Please see PR: https://github.com/nats-io/nats.go/pull/1713 

and the backstory in this this issue: https://github.com/nats-io/nats.go/issues/1694

In short, we are using the NATS 2.10+ Auth Callout feature; where we have our own IAM JWT tokens that expire every few minutes. The user obtains an IAM JWT, talks to nats using `unused` as the username but the JWT as the `password`, and the nats authCallout implementation verifies the IAM JWT from the password and eventually emits (after checking our in house permissions DB) a nats JWT, which has its own expiry schedule (ours currently is set to 24h). 

However, when that nats JWT expires (every 24h) the connection is interrupted. In this case, we need to present an updated IAM JWT, as the old one is expired (every 5 minutes). The golang client added that callback that fetches a new `username, password` pair before reconnecting in that PR linked above. We would kindly request the same callback.

I believe the `_user_credentials_callback` should go here-ish: https://github.com/nats-io/nats.py/blob/109817f756df00fdac116a6deacfd9cff732406b/nats/aio/client.py#L306-L309

And I believe you should call it right after this ish: https://github.com/nats-io/nats.py/blob/109817f756df00fdac116a6deacfd9cff732406b/nats/aio/client.py#L1490-L1491

`ish` means I've scanned this code but do not have any deep expertise with this python lib. 

Note: I originally tried to abuse the `disconnected_cb` in the golang client (the python client does have this) and just setting the equivelent of`self._user_credentials` in that callback, but it wasnt thread safe; hence their addition of that CB. I did not investigate whether the python lib is thread safe, and maybe the disconnected_cb can be (ab)used for this. 

### Use case

Nats 2.10 auth callout. 

### Contribution

_No response_

caspervonb:
Mirror the existing \`token\` handling for \`user\` and \`password\`: both now accept either a \`str\` or a zero-arg callable that returns the current value. The callable is invoked every time a CONNECT frame is built, so callers can rotate credentials on reconnect — typical use case is NATS 2.10 auth callout, where the JWT carried as the password expires on a short cycle.

Closes #714.


caspervonb:
@claude please review again

tcarpenter-eqix:
Hi - could we please get this re-engaged? We have internal users waiting to adopt our auth callout (built on the nats auth callout feature) in python, this would be huge for us. (I raised #714 and also https://github.com/nats-io/nats.go/pull/1713). I'm happy to help test , just let us know, but it seems like this is a pretty straitforward change

nats/src/nats/aio/client.py
```diff
@@ -1756,8 +1757,14 @@ def _connect_command(self) -> bytes:
             # In case there is no password, then consider handle
             # sending a token instead.
             elif self.options["user"] is not None and self.options["password"] is not None:
-                options["user"] = self.options["user"]
-                options["pass"] = self.options["password"]
+                user = self.options["user"]
+                if callable(user):
```
renerocksai, line 1764: Tiny polish: with this addition, the `if callable(x): x = x()` dance appears four times in `_connect_command` (token twice, user, password). Consider a small helper, e.g.

```python
def _resolve(value):
    return value() if callable(value) else value
```

or the expression form `user = user() if callable(user) else user`. Totally fine to leave as-is since it deliberately mirrors the existing token style.

nats/src/nats/aio/client.py
```diff
@@ -110,6 +110,7 @@
 JWTCallback = Callable[[], Union[bytearray, bytes]]
 SignatureCallback = Callable[[str], bytes]
 TokenCallback = Callable[[], str]
+CredentialCallback = Callable[[], str]
```
renerocksai, line 116: Since the target use case is short-cycle JWT rotation, I think users may be tempted to fetch fresh credentials over the network inside this callback — but it's invoked synchronously in `_connect_command`, on the event loop, during both connect and reconnect, so any blocking I/O there would stall the loop mid-reconnect. This is the same constraint the existing `TokenCallback`/`JWTCallback` already have, so nothing new — but consider a short comment or doc note that the callable must not block (i.e. return a cached/pre-fetched value).

renerocksai (approved):

Two small polish suggestions — hedged inline, take or leave.

caspervonb:
Review comments addressed, PTAL @claude

PR #892, opened 2026-04-17
Type KV Entry.operation as Literal

### LINKED ISSUE #562: KV's `Entry.operation` should have a literal type
### Proposed change

Minor thing, but it would be nice if in kv.py in the `Entry` dataclass, we instead used literals here for the `operation` field:

```python
    @dataclass
    class Entry:
        """
        An entry from a KeyValue store in JetStream.
        """
        bucket: str
        key: str
        value: Optional[bytes]
        revision: Optional[int]
        delta: Optional[int]
        created: Optional[int]
        operation: Optional[str]
```

https://github.com/nats-io/nats.py/blob/main/nats/js/kv.py#L67-L78

```python
operation: Optional[Literal[KV_DEL, KV_PURGE, ...]]
# or
operation: Optional[Literal['DEL', 'PURGE', ...]]
```

Could also be an enumerated type but I think using a literal is fine here.

### Use case

It makes development easier if you're using an editor that autocompletes using type hints. It also makes it a little more self-documenting. 

### Contribution

Sure

caspervonb:
Narrows \`Entry.operation\` from \`Optional[str]\` to \`Optional[Literal[\"DEL\", \"PURGE\"]]\` — the only two values the field can ever hold — so editors and type checkers can autocomplete against it.

Closes #562.


caspervonb:
@claude please review again

nats/src/nats/js/kv.py
```diff
@@ -93,7 +95,7 @@ class Entry:
         revision: Optional[int]
         delta: Optional[int]
         created: Optional[int]
-        operation: Optional[str]
+        operation: Optional[Literal["DEL", "PURGE"]]
```
renerocksai, line 98: Two things about this narrowing, one of which I think matters:

**1. Nothing enforces the literal at the only site that sets it.** The single place `operation` gets a non-`None` value is the watch callback:

```python
529:  op = msg.header.get(KV_OP)   # raw wire string, no membership check
...
561:  operation=op,
```

So any `KV-Operation` value other than `DEL`/`PURGE` lands in `Entry.operation` and quietly violates the declared type. It doesn't get caught statically either — `async def watch_updates(msg)` (line 520) is unannotated, so `msg.header.get(...)` is `Any` and the assignment at 561 type-checks vacuously.

What makes this stand out is that the sibling branch immediately below already does exactly this validation: the `Nats-Marker-Reason` path (531–543) whitelists `MaxAge`/`Purge`/`Remove` and deliberately skips unknown reasons rather than emitting an ambiguous entry. The `KV_OP` branch is now the odd one out. Mirroring it would be small:

```python
op = msg.header.get(KV_OP)
if op not in (KV_DEL, KV_PURGE):
    op = None
```

That is a behaviour change in an otherwise typing-only PR, so it may well be out of scope on purpose — but as it stands the annotation promises something the code doesn't guarantee.

**2. The constants aren't tied to the new type.** `KV_DEL = "DEL"` / `KV_PURGE = "PURGE"` (lines 34–35) still infer plain `str`, so the narrowing doesn't reach the code that consumes them: `op = KV_PURGE` / `op = KV_DEL` (533/535) infer `str`, and `update.operation == KV_DEL` (367) gets no exhaustiveness benefit. A named alias would close the loop and avoid restating the literal in two places:

```python
if TYPE_CHECKING:
    from typing import Literal

    KVOperation = Literal["DEL", "PURGE"]

    from nats.js import JetStreamContext

KV_DEL: KVOperation = "DEL"
KV_PURGE: KVOperation = "PURGE"

# ...
    operation: Optional[KVOperation]
```

(`KV_DEL: Final = "DEL"` is the alternative if you want per-constant precision — `Literal["DEL"]` rather than the union. Both are annotation-only, so PEP 563 keeps them free at runtime on 3.7.)

Aside, outside this PR's diff: `KeyDeletedError.__init__(self, entry=None, op=None)` in `js/errors.py` carries the same value into a public `.op` attribute and stays untyped — the natural companion to this change if you feel like it.

renerocksai (commented):

1. The narrowed `Literal` isn't enforced at the one site that actually populates `operation` — the raw `KV-Operation` header is passed straight through, and because the watch callback is untyped no checker catches it. Details inline.

The rest is a small tightening around the `KV_DEL`/`KV_PURGE` constants — hedged inline, take or leave.

nats/src/nats/js/kv.py
caspervonb: Done in 4eaba080 (merged with the branch tip in 215640df): the watch callback now drops any KV-Operation value other than DEL/PURGE, mirroring the marker-reason branch, and a KVOperation alias under TYPE_CHECKING now types KV_DEL, KV_PURGE and Entry.operation. Left KeyDeletedError.op alone as it is outside this PR.

caspervonb:
PTAL @claude

PR #893, opened 2026-04-17
Handle non-binary WebSocket frame types

### LINKED ISSUE #673: Encountered error on websocket
### Observed behavior

Hi, in my company we are using NATS on several services. The client connects via websocket. We have noticed that randomly this error happens:

```bash
nats: encountered error
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/nats/aio/client.py", line 2117, in _read_loop
    await self._ps.parse(b)
  File "/usr/local/lib/python3.8/dist-packages/nats/protocol/parser.py", line 95, in parse
    self.buf.extend(data)
TypeError: can't extend bytearray with int
```

After this error I have to restart the service because the client doesn't reconnect with the server.

I have investigated a bit and I found this other issue:
- https://github.com/nats-io/nats.py/issues/361

where it seems the same problem but should be fixed and the commit merged time ago.

Someone can kindly help me to understand the problem and trying to find a workaround to handle it?? Thank you in advance!

### Expected behavior

I expected a reconnection of the client ideally or a way to handle it

### Server and client version

nats-server: v2.10.22
nats-py: 2.9.0

### Host environment

_No response_

### Steps to reproduce

It happens randomly, I was not able to reproduce it by myself, sorry

caspervonb:
The WebSocket transport was returning \`msg.data\` from \`aiohttp.WebSocketResponse.receive()\` for every frame type other than \`CLOSED\`. For \`CLOSE\`, \`msg.data\` is an \`int\` (the close code), which caused the NATS parser to crash with \`TypeError: can't extend bytearray with int\`, killing the read loop with no reconnect.

- \`BINARY\` → return \`msg.data\` (the bytes).
- \`ERROR\` → wrap the underlying aiohttp exception in \`ConnectionError\` (via \`raise ... from msg.data\`) so it lands in the read loop's \`OSError\` branch and reconnects.
- \`CLOSE\` / \`CLOSING\` / \`CLOSED\` / \`TEXT\` / anything else → return \`b""\` so the read loop treats it as EOF and reconnects.

Closes #673.

### DIFF
--- a/nats/src/nats/aio/transport.py
+++ b/nats/src/nats/aio/transport.py
@@ -238,10 +238,16 @@ class WebSocketTransport(Transport):
 
     async def readline(self):
         data = await self._ws.receive()
-        if data.type == aiohttp.WSMsgType.CLOSED:
-            # if the connection terminated abruptly, return empty binary data to raise unexpected EOF
-            return b""
-        return data.data
+        if data.type == aiohttp.WSMsgType.BINARY:
+            return data.data
+        if data.type == aiohttp.WSMsgType.ERROR:
+            # msg.data is the underlying aiohttp exception. Wrap it in a
+            # ConnectionError so the read loop's OSError handler triggers a
+            # reconnect instead of its generic-Exception log-and-break branch.
+            raise ConnectionError("nats: websocket error") from data.data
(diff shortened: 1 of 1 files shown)


wallyqs (approved):
LGTM

PR #896, opened 2026-04-21
Enforce server max_payload on publish

caspervonb:
Reject payloads larger than the server-advertised max_payload before writing them to the wire. Without this check, oversized payloads cause a server-side disconnect mid-stream.


caspervonb:
Superseded by #888, which also accounts for HPUB header bytes.

PR #899, opened 2026-04-21
Avoid hang when KeyWatcher.stop() runs on a full queue

### LINKED ISSUE #898: `KeyWatcher.stop()` blocks indefinitely when internal queue is full.
### Observed behavior

We have observed this issue on a system running over a longer period of time with considerable NATS KV bucket activity.

Our application uses `KeyValue.KeyWatcher.updates()` to wait for a change of a KV entry used as a lock. In case of lock contention, lock acquisition (`KeyValue.create()`) may fail and the application loops back to the `KeyValue.KeyWatcher.updates()` call multiple times. When lock acquisition is eventually successful, the application proceeds to free up the watcher's resources by calling `KeyValue.KeyWatcher.stop()`, at which point we have observed that our application may enter a deadlock on `await self._updates.put(KeyValue.KeyWatcher.STOP_ITER)` because queue `KeyValue.KeyWatcher._updates` is full (`maxsize=256`) at this moment.

The code location of the call from our application into `nats-py` can be found here: https://gitlab.com/genomedx/labautomation/maestro/-/blob/0.1.12/src/maestro/nats/nats.py#L303

More context is available here: https://gitlab.com/genomedx/labautomation/maestro/-/work_items/561

We are looking into application workarounds to not be dependent on a fix in `nats-py`.

### Expected behavior

A call to `KeyValue.KeyWatcher.stop()` should never be at risk of blocking indefinitely.

### Server and client version

Server version: `2.12.1`
Client version: `2.14.0` (`nats-py`)

### Host environment

_No response_

### Steps to reproduce

_No response_

caspervonb:
Fixes #898. KeyWatcher.stop() awaited put on a bounded queue; when the consumer wasn't draining, the put blocked indefinitely. Push the sentinel with put_nowait, evicting one queued update on QueueFull.

### DIFF
--- a/nats/src/nats/js/kv.py
+++ b/nats/src/nats/js/kv.py
@@ -370,7 +370,15 @@ class KeyValue:
             stop will stop this watcher.
             """
             await self._sub.unsubscribe()
-            await self._updates.put(KeyValue.KeyWatcher.STOP_ITER)
+            while True:
+                try:
+                    self._updates.put_nowait(KeyValue.KeyWatcher.STOP_ITER)
+                    return
+                except asyncio.QueueFull:
+                    try:
+                        self._updates.get_nowait()
+                    except asyncio.QueueEmpty:
+                        pass
 
         async def updates(self, timeout=5.0):
             """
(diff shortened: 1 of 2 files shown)


caspervonb:
@claude please review again

caspervonb:
@claude please review again

Jarema (approved):
LGTM!

PR #901, opened 2026-04-21
Add nats.aio to nats.client migration guide

caspervonb:
Adds nats-core/MIGRATION.md with side-by-side examples covering connect, pub/sub, request/reply, auth, TLS, and lifecycle. Calls out the surface still only in nats.aio (JetStream, KV, WebSocket, multi-URL, LDM) so users can plan migrations realistically.

### DIFF
--- a/nats-core/MIGRATION.md
+++ b/nats-core/MIGRATION.md
@@ -1,0 +1,334 @@
+# Migrating from `nats.aio` to `nats.client`
+
+`nats.client` is the modern NATS client. It targets Python 3.13+, uses native async syntax, and models subscriptions as async iterators instead of callbacks.
+
+This guide covers the core publish/subscribe/request surface. JetStream, KV, Object Store, and Micro are not part of `nats.client` and still live in `nats.aio` / `nats.js` / `nats.micro` — applications that depend on those should continue using `nats-py` or combine both clients.
+
+## Install
+
+```bash
+pip install nats-core
+```
+
+Both packages expose the `nats` namespace; they can be installed side by side during a migration.
+
+## Connecting
+
+### Legacy
+
+```python
+import nats
+
+nc = await nats.connect(
+    servers=["nats://localhost:4222"],
+    name="my-app",
+    user="alice",
+    password="secret",
+    error_cb=on_error,
+    reconnected_cb=on_reconnect,
+    disconnected_cb=on_disconnect,
+)
+```
+
+### Modern
+
+```python
+from nats.client import connect
+
+client = await connect(
+    "nats://localhost:4222",
+    user="alice",
+    password="secret",
+)
+client.add_error_callback(on_error)
+client.add_reconnected_callback(on_reconnect)
+client.add_disconnected_callback(on_disconnect)
+```
+
+Differences:
+
+- `connect()` takes a single `url` positional argument instead of a `servers=[...]` list.
+- Event callbacks are registered via `add_*_callback` after construction rather than as constructor keyword arguments.
+- Returns a `Client` instance usable as an `async with` context manager.
+
+## Publishing
+
+### Legacy
+
+```python
+await nc.publish("greet.alice", b"hello")
+await nc.publish("greet.alice", b"hello", reply="inbox.1", headers={"k": "v"})
+await nc.flush()
+```
+
+### Modern
+
+```python
+await client.publish("greet.alice", b"hello")
+await client.publish("greet.alice", b"hello", reply="inbox.1", headers={"k": "v"})
+await client.flush()
+```
+
+The signature is the same in practice. `reply` and `headers` are keyword-only in the modern API.
+
+## Subscribing
+
+The biggest shape change: subscriptions are async iterators; there is no `cb=` parameter.
+
+### Legacy (callback)
+
+```python
+async def handler(msg):
+    print(msg.subject, msg.data)
+    await msg.respond(b"ok")
+
+sub = await nc.subscribe("greet.*", cb=handler)
+```
+
+### Modern (async iterator)
+
+```python
+subscription = await client.subscribe("greet.*")
+
+async for message in subscription:
+    print(message.subject, message.data)
+    if message.reply:
+        await client.publish(message.reply, b"ok")
+```
+
+Notes:
+
+- Drive the iterator in a task if you want the handler to run concurrently with other work.
+- `subscription.next(timeout=...)` is available for a single-message pull.
+- Replying to a request is an explicit `client.publish(message.reply, ...)`; there is no `message.respond()`.
+
+## Queue groups
+
+```python
+# Legacy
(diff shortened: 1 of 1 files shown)


nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+import nats
+
+nc = await nats.connect(
+    servers=["nats://localhost:4222"],
+    name="my-app",
+    user="alice",
+    password="secret",
+    error_cb=on_error,
+    reconnected_cb=on_reconnect,
+    disconnected_cb=on_disconnect,
```
wallyqs, line 26: @claude think lots of other fields are different here, double check the differences

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+await client.publish("greet.alice", b"hello")
+await client.publish("greet.alice", b"hello", reply="inbox.1", headers={"k": "v"})
+await client.flush()
+```
+
+The signature is the same in practice. `reply` and `headers` are keyword-only in the modern API.
+
+## Subscribing
+
+The biggest shape change: subscriptions are async iterators; there is no `cb=` parameter.
```
wallyqs: 🚩 need to fix this one

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+
+### NKey
+
+```python
+# Legacy — path to seed file
+nc = await nats.connect("nats://...", nkeys_seed="/path/to/user.nk")
+
+# Modern — pass seed as a Path
+from pathlib import Path
+client = await connect("nats://...", nkey=Path("/path/to/user.nk"))
```
wallyqs: @claude not necessary to change, can you look into making backwards compatible here

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+nc = await nats.connect(
+    "nats://...",
+    user_credentials="/path/to/user.creds",
+)
+
+# Modern
+from pathlib import Path
+client = await connect(
+    "nats://...",
+    jwt=Path("/path/to/user.creds"),
```
wallyqs: same here, needs to be user_credentials
@claude can you look into a fix to be backwards compatible here and closer to the Go client

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+| Legacy | Modern |
+|---|---|
+| `nats.errors.Error`, `TimeoutError`, `NoServersError`, `NoRespondersError`, ... | `nats.client.errors.StatusError`, `NoRespondersError`, `SlowConsumerError`, `MaxPayloadError` |
+| `await nc.publish(...)` → `nats.errors.OutboundBufferLimitError` on back-pressure | `await client.publish(...)` → `MaxPayloadError` when a payload exceeds server `max_payload` |
+
+`nats.client` does not expose a shared base `Error`; catch `Exception` if you need a generic handler until that gap is closed.
+
+## Message delivery semantics
+
+- Legacy callbacks run inside the read loop. A slow handler slows all incoming traffic on the connection.
```
wallyqs: 🚩 need to fix this one, can't ship with this as is

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,334 @@
...
+- Modern subscriptions deliver to an internal queue (default 65,536 messages / 64 MiB). When the queue fills, a `SlowConsumerError` is reported via `add_error_callback` and subsequent messages are dropped until the consumer catches up.
+
+Tune per-subscription limits:
+
+```python
+subscription = await client.subscribe(
+    "firehose",
+    max_pending_messages=1_000_000,
```
wallyqs: change to the pending limit ones which follow the Go client
@claude make a change to use the pending_* variants from the nats-py client on the subscription for backwards compatibility

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,141 @@
...
+
+| Legacy | Modern |
+|---|---|
+| `connect_timeout` | `timeout` |
+| `max_reconnect_attempts` | `reconnect_max_attempts` |
+| `dont_randomize` | `no_randomize` |
+
+New reconnect tuning: `reconnect_time_wait_max`, `reconnect_jitter`, `reconnect_timeout`.
+
+Gone (no equivalent yet): `name`, `pedantic`, `verbose`, `pending_size`, `flush_timeout`, `flusher_queue_size`, `drain_timeout`, `ws_connection_headers`, `reconnect_to_server_handler`, `signature_cb` / `user_jwt_cb` (pass handler tuples to `nkey=` / `jwt=` instead).
```
caspervonb: Defect, `name`, `pedantic` and `verbose` will be exposed.

nats-core/MIGRATION.md
```diff
@@ -0,0 +1,151 @@
...
+
+```python
+# Before
+await nats.connect(servers=["nats://a:4222", "nats://b:4222"])
+
+# After
+await connect("nats://a:4222")
+```
+
+Multi-URL seed lists aren't supported yet — pick one server; discovered cluster members will still be used for reconnect.
```
caspervonb, line 17: Just overlooked, should be fixed.

caspervonb: Pull request open.

nats-core/MIGRATION.md
caspervonb: Pull request open.

nats-core/MIGRATION.md
caspervonb: What's the actual motivation here? Want to make sure I'm solving the right thing before pushing back.

A few reasons I've been leaning iterator-only:

Perf's not the driver — nats-core is already ~280× faster than legacy on the iterator path, so throughput isn't the win.

The bigger issue is `Subscription` would have to implement `__aiter__`/`__anext__` and also support callbacks, and the two can't coexist cleanly. Legacy handles this by raising if you iterate a callback subscription, which means the type claims to be an async iterable but isn't. I'd rather not ship that.

Python also has first-class async iteration, which Go doesn't — so the nats.go rationale doesn't really carry over. If it's about migration ergonomics, the user-space wrapper is basically nothing:

```python
async def dispatch(sub, cb):
    async for msg in sub:
        await cb(msg)
asyncio.create_task(dispatch(sub, handler))
```

We could even ship that as `sub.consume(handler)` returning a Task if the ergonomics matter.

Is there a specific case you're hitting where that falls short? Happy to look at it — just want to understand what we'd actually be buying.

wallyqs: 1) need to show me how to repro that number of 280x faster or which angle you are seeing, so far have not seen why not feasible too boost that in the nats-py client, 2) supporting callbacks and iterate is not a big issue,  and yes ergonomics matter here

aricart (approved):
LGTM

PR #905, opened 2026-04-21
Default reconnect_max_attempts to 60

caspervonb:
Match nats.go's default so short cluster events don't close the client after ~20s. Addresses audit finding #7 in `.claude/audit/02-reconnect.md`.


aricart (approved):
LGTM - default changed? typically we have said 10

caspervonb:
> LGTM - default changed? typically we have said 10

Mirror's go Default so just went with that.

PR #907, opened 2026-04-22
Mirror initial-connect TLS flow on reconnect

caspervonb:
The reconnect path forced TLS-first regardless of `tls_handshake_first` and ignored `tls_required` from INFO, so upgrade-mode servers failed the handshake and credentials could leak in plaintext if a server started requiring TLS mid-session. Covers audit finding 07-tls.md #2.


caspervonb:
PTAL @claude

caspervonb:
Last looks @claude

aricart (approved):
LGTM

PR #908, opened 2026-04-22
Verify CONNECT acceptance on reconnect

caspervonb:
Reconnect wrote CONNECT and flipped to CONNECTED without waiting for PONG, so a server -ERR (e.g. rotated token) was silently ignored and subsequent publishes dropped into a dead socket.


caspervonb:
PTAL @claude

aricart (approved):

PR #910, opened 2026-04-23
Add force_reconnect for operator-initiated failover

caspervonb:
Mirrors nats.go's `Conn.ForceReconnect`: closes the current connection, triggers the standard reconnect loop, and kicks the backoff sleep if a reconnect is already in progress. Raises on closed or when `allow_reconnect=False`.


wallyqs (approved):

caspervonb:
PTAL @claude

PR #913, opened 2026-04-30
Default missing port in _setup_server_pool list path; preserve scheme/userinfo on rewrite

renerocksai:
## Summary

Fixes two related URL-parsing bugs in `nats/aio/client.py` that surface when a caller passes a server URL without an explicit port.

### 1) `_setup_server_pool` list path skips the port default

The single-string path routes through `_parse_server_uri`, which defaults a missing port to `4222` ([`client.py:1424`](https://github.com/nats-io/nats.py/blob/main/nats/src/nats/aio/client.py#L1424)). The list path didn't — it called `urlparse(server)` per entry and appended `Srv(uri)` directly, so an entry like `tls://connect.ngs.global` produced `uri.port == None` and the asyncio TCP connect landed on port `0`, raising `EADDRNOTAVAIL`.

This shows up in real deployments because external sources routinely emit port-less URLs:

- **`nats context add`** writes the URL the user supplied verbatim. The Go `nats` CLI papers over the missing port internally, so context files frequently contain entries like `tls://connect.ngs.global`.
- **Synadia Cloud's "Get Connected"** snippet copies a URL without a port. A user pasting that into `nats.connect(servers=[...])` hits the bug.

### 2) `_parse_server_uri` port-default rewrite loses scheme + userinfo

When `uri.port` was `None`, the rewrite did:

```python
uri = urlparse(f"nats://{uri.hostname}:4222")
```

This silently downgraded `tls://host` to `nats://host` (TLS lost), dropped any userinfo (`nats://TOKEN@host` → token gone), and stripped IPv6 brackets (since `uri.hostname` is bracket-less).

### Fix

- `_parse_server_uri`: track the post-scheme-prepend URL as `normalized` and append `:4222` to *that* on port-default. A trailing `":"` (e.g. `"host:"`) is stripped first so we don't produce `"host::4222"`. Original scheme, userinfo, and IPv6 brackets all round-trip.
- `_setup_server_pool` list path: now routes each entry through the (fixed) `_parse_server_uri` for symmetry — no more inline duplication of the port-default logic.

### Tests

- New `test_connect_string_preserves_scheme_and_userinfo_on_port_default` — `tls://` scheme preservation, `TOKEN@host` and `user:pass@host` round-trip, IPv6 brackets.
- New `test_connect_list_defaults_missing_port` — list path: ngs-style `tls://` URL, mixed entries (defaulted + explicit + `tls`), userinfo round-trip, IPv6, `ws`/`wss` carve-out.
- One assertion in the existing `test_connect_syntax_sugar` updated: input `"@demo.nats.io:"` now yields `uri.username == ""` (matching the no-trailing-colon `"@demo.nats.io"` assertion already at the same test). The previous `None` was a side-effect of the buggy rewrite dropping userinfo entirely.

## Test plan

- [x] `uv run ruff format --check`
- [x] `uv run ruff check`
- [x] `uv run codespell`
- [x] `uv build --package nats-py`
- [x] `uv run pytest nats/tests/test_client.py::ClientTest::test_connect_syntax_sugar nats/tests/test_client.py::ClientTest::test_connect_list_defaults_missing_port nats/tests/test_client.py::ClientTest::test_connect_string_preserves_scheme_and_userinfo_on_port_default` — all pass
- [x] Python 3.8 compat preserved (only f-strings + str slicing + `urlparse`)
- [ ] Full `pytest ./nats/tests` matrix on CI (clean container, no port-4222 squatters)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


renerocksai:
Thanks for the review @synadia-claude-reviewer — both items addressed in 3861d12.

- **`_parse_server_uri` line 1424**: switched from string-append on `normalized` to `uri._replace(netloc=...).geturl()`, so the rewrite no longer relies on the absence of a path component. Same trailing-`":"` handling, same observable behaviour.
- **Comment trim**: dropped the historical "Previously this..." context from `_parse_server_uri`, the list-path call site, and the two new tests. The remaining inline comments document timeless rationale only (the trailing-`":"` edge case, sub-block intent in the test).

Verified locally: `ruff format --check`, `ruff check`, `codespell` all clean; the three URL-parsing tests all pass.

Note on the failing CI runs at the time of the previous push: jobs `nats (3.8/3.9/3.10/3.11)` failed at `./nats/scripts/install_nats.sh` with curl exit 56 (network failure pulling the nats-server binary from `binaries.nats.dev` — all 4 hit it at the same wall-clock instant, signature of a transient CDN flake). Jobs `nats (3.12/3.13)` got past install and failed in `test_js.py::KVTest::test_kv_basic`, which is unrelated to URL parsing — none of the URL-parsing tests were red on either path.

caspervonb:
PTAL again @claude

caspervonb (approved):
lgtm

PR #916, opened 2026-05-01
Add README for nats-key-value

caspervonb:
Package was missing a README; this matches the structure of nats-core and nats-server.


aricart (approved):
LGTM

PR #918, opened 2026-05-01
Fix legacy KV, object store, compression tests on nats-server 2.14

caspervonb:
Tests assumed servers >2.12 would omit allow_direct/max_age/mirror_direct/compression from StreamConfig. 2.14.0 ships them with concrete defaults instead, matching 2.9-2.12, so the version-gated branches were dead.


aricart (approved):
LGTM

PR #920, opened 2026-05-06
Add support for consumer reset

caspervonb:
Implements ADR-60's `$JS.API.CONSUMER.RESET` endpoint added in nats-server 2.14: `Stream.reset_consumer(name, seq=None)` and `Consumer.reset(seq=None)`. Empty/0 resumes redelivery from one above the ack floor; a non-zero `seq` sets the ack floor below it so the next delivered message has a stream sequence `>= seq`.

Ordered-consumer behavior is unchanged: an external reset surfaces as a delivery-sequence gap, which the existing recovery path already handles by recreating the consumer (per ADR-60 §Consequences).

### DIFF
--- a/nats-jetstream/src/nats/jetstream/consumer/pull.py
+++ b/nats-jetstream/src/nats/jetstream/consumer/pull.py
@@ -521,6 +521,20 @@ class PullConsumer(Consumer):
         # Refresh info from server
         return self._info
 
+    async def reset(self, seq: int | None = None) -> int:
+        """Reset this consumer's delivery state (ADR-60).
+
+        See :meth:`Stream.reset_consumer` for the full semantics. ``seq=None``
+        (or ``0``) resumes redelivery from one above the consumer's ack
+        floor; a non-zero ``seq`` sets the ack floor to one below it so the
+        next delivered message has a stream sequence ``>= seq``.
+
+        Returns:
+            The stream sequence the next delivered message will be at or
+            above.
+        """
+        return await self._stream.reset_consumer(self._info.name, seq)
+
     async def close(self) -> None:
         """Close the consumer.
 
--- a/nats-jetstream/src/nats/jetstream/consumer/pull.py
+++ b/nats-jetstream/src/nats/jetstream/consumer/pull.py
@@ -529,11 +529,26 @@ class PullConsumer(Consumer):
         floor; a non-zero ``seq`` sets the ack floor to one below it so the
         next delivered message has a stream sequence ``>= seq``.
 
+        Cached :attr:`info` is refreshed from the server's response.
+
         Returns:
             The stream sequence the next delivered message will be at or
             above.
         """
-        return await self._stream.reset_consumer(self._info.name, seq)
+        api = getattr(self._stream._jetstream, "_api", None)
+        if api is None:
+            raise RuntimeError("JetStream does not have an API client")
+
+        if seq is None:
+            response = await api.consumer_reset(self._stream._name, self._info.name)
+        else:
+            response = await api.consumer_reset(self._stream._name, self._info.name, seq=seq)
+
+        reset_seq = response.pop("reset_seq")
+        # Drop response-envelope discriminator before parsing into ConsumerInfo.
+        response.pop("type", None)
+        self._info = ConsumerInfo.from_response(response, strict=self._stream._jetstream.strict)
+        return reset_seq
 
     async def close(self) -> None:
         """Close the consumer.
--- a/nats-jetstream/src/nats/jetstream/consumer/pull.py
+++ b/nats-jetstream/src/nats/jetstream/consumer/pull.py
@@ -16,6 +16,7 @@ from typing import (
 from nats.jetstream.consumer import (
     Consumer,
     ConsumerInfo,
+    ConsumerReset,
     MessageBatch,
     MessageStream,
 )
@@ -521,7 +522,7 @@ class PullConsumer(Consumer):
         # Refresh info from server
         return self._info
 
-    async def reset(self, seq: int | None = None) -> int:
+    async def reset(self, seq: int | None = None) -> ConsumerReset:
         """Reset this consumer's delivery state (ADR-60).
 
         See :meth:`Stream.reset_consumer` for the full semantics. ``seq=None``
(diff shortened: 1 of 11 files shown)


caspervonb:
PTAL @claude

nats-jetstream/src/nats/jetstream/consumer/pull.py
```diff
@@ -521,6 +521,20 @@ async def get_info(self) -> ConsumerInfo:
         # Refresh info from server
         return self._info
 
+    async def reset(self, seq: int | None = None) -> int:
```
renerocksai: `reset` lives only on `PullConsumer`, but `Stream.create_consumer` / `get_consumer` / `update_consumer` / `create_or_update_consumer` / `ordered_consumer` are all typed to return the `Consumer` Protocol (`consumer/__init__.py`), which doesn't declare `reset`. The new tests work at runtime because the concrete return is a `PullConsumer`, but a user with a strict type-checker calling `await consumer.reset(seq=3)` on the result of `create_consumer` will get an error.

Suggest adding `async def reset(self, seq: int | None = None) -> int: ...` to the `Consumer` Protocol. `OrderedConsumer` can either delegate to its existing recreation path or raise `NotImplementedError` — the PR description already explains why ordered consumers don't need a server-side reset, so a documented `NotImplementedError` would be defensible.

nats-jetstream/src/nats/jetstream/stream.py
```diff
@@ -1516,6 +1516,45 @@ async def resume_consumer(self, consumer_name: str) -> None:
...
+        """
+        api = getattr(self._jetstream, "_api", None)
+        if api is None:
+            raise RuntimeError("JetStream does not have an API client")
+
+        if seq is None:
+            response = await api.consumer_reset(self._name, consumer_name)
+        else:
+            response = await api.consumer_reset(self._name, consumer_name, seq=seq)
+        return response["reset_seq"]
```
renerocksai: The server returns a full consumer-info-shaped payload alongside `reset_seq`, and ADR-60 explicitly calls this out as something clients should be able to cache ("this response can then also be kept by clients if they need to keep a cached consumer response"). Right now both `Stream.reset_consumer` here and `PullConsumer.reset` discard everything except `reset_seq`, and `PullConsumer._info` is never refreshed — so after a reset, `consumer.info.num_pending` / `delivered` / `ack_floor` are stale by design.

Two reasonable shapes:
- Update `self._info` from the response inside `PullConsumer.reset` (cheap, no extra round-trip), or
- Surface `ConsumerInfo` from `Consumer.reset` so callers don't need a follow-up `get_consumer_info`.

There's precedent for throwing the response away (`pause_consumer` returns `None`), but `reset_seq` plus the post-reset state is much richer than pause and the server is going out of its way to send it.

nats-jetstream/tests/test_consumer.py
```diff
@@ -1082,3 +1082,90 @@ async def test_consumer_info_timestamp(jetstream: JetStream):
...
+        await msg.ack()
+
+    info = await stream.get_consumer_info("reset_floor")
+    floor = info.ack_floor["stream_seq"]
+
+    reset_seq = await consumer.reset()
+    assert reset_seq == floor + 1
+
+    msg = await consumer.next(max_wait=1.0)
+    assert msg.metadata.sequence.stream >= reset_seq
```
renerocksai: A few small test tightenings — all hedged, take or leave:

1. **This `>=` assertion**: I think it should be `== reset_seq` (i.e. `floor + 1`). The setup is deterministic — 5 contiguous publishes, no filter, two fetched+acked, then empty-payload reset which preserves the ack floor per ADR-60 — so the next delivered message should land exactly at `floor + 1`. `>=` would let a regression that skipped messages slip through silently. Worth tightening if it holds in CI.

2. **`test_stream_reset_consumer` (line 1137)**: only asserts `reset_seq == 2`, which verifies the API plumbing but not the server-side semantics. Consider adding a `next()` after the reset and asserting `msg.metadata.sequence.stream == 2` to close the loop on what `reset_seq` is supposed to mean.

3. **`test_consumer_reset_below_start_seq_rejected` (line 1154)**: optional — tighten the `pytest.raises(JetStreamError)` with `match=` on a substring of the expected error message so a future regression that returns the wrong error doesn't pass silently. ADR-60 doesn't pin the exact code, so a substring match is probably the right level of strictness.

renerocksai (commented):
Please bear in mind, I am fairly new to the nats-py client and might be wrong in my assessments.

Overall this looks idiomatic and faithful to ADR-60 — the empty-payload / `seq=0` / `seq>0` semantics in the docstrings match the spec, error mapping mirrors `consumer_pause` / `consumer_info`, and the `Unpack[ConsumerResetRequest]` typed-kwargs surface is consistent with the rest of the API client. No async hazards (no missed `await`, no blocking I/O on the loop).

Two things I'd treat as should-fix:

1. `reset` should live on the `Consumer` Protocol, not just on `PullConsumer` — otherwise the return type of `Stream.create_consumer` (`Consumer`) doesn't actually expose the new method to type-checkers. Details inline.
2. The full reset response is discarded; ADR-60 explicitly intends it to refresh cached consumer state. Either update `PullConsumer._info` from the response or surface `ConsumerInfo` from `Consumer.reset`. Details inline.

The rest are small test tightenings — hedged inline, take or leave.

caspervonb:
> reset should live on the Consumer Protocol, not just on PullConsumer — otherwise the return type of Stream.create_consumer (Consumer) doesn't actually expose the new method to type-checkers. Details inline.

Might need to circle back here later, in that reset should be on something like a ResettableConsumer protocol, landing it on the top level for now can tighten the types later if needed.

renerocksai (approved):
See my comment. Since I can't be the judge, I'll approve.

Jarema (changes requested):
I think we are missing:

1. we return int instead of Consumer Info like other clients
2. we do not map 10204 error?

caspervonb:
@Jarema thanks — addressed both in `Address review comments`:

1. New `ConsumerReset` dataclass (`info: ConsumerInfo`, `reset_seq: int`) returned by `Stream.reset_consumer` and `Consumer.reset`. Mirrors Go's `*ConsumerResetResponse` and Rust's `ConsumerResetResponse` shape (composition rather than embedding to keep the dataclass clean — `result.info.delivered`). On the consumer path `_info` is still refreshed as a side effect, so `consumer.info is result.info` after the call.
2. `CONSUMER_INVALID_RESET = 10204` on `ErrorCode`, new `ConsumerInvalidResetError` class, mapped in `Client.consumer_reset`. The existing `test_consumer_reset_below_start_seq_rejected` is now tightened to `pytest.raises(ConsumerInvalidResetError)` and passes — confirms 10204 is what the server returns for that scenario.

Both `ConsumerReset` and `ConsumerInvalidResetError` are exported from `nats.jetstream`.

aricart (approved):

Jarema (approved):
LGTM

PR #921, opened 2026-05-07
Add sourcing-consumer config

caspervonb:
Implements ADR-60's pre-created push-durable consumer for stream sourcing/mirroring, added in nats-server 2.14: a `StreamSource.consumer = StreamConsumerSource(name, deliver_subject)` field plus the `flow_control` ack policy variant that the server requires for these consumers. Also picks up the related error codes (10213–10222) on the `ErrorCode` constants.

Test pre-creates a push consumer with `ack_policy="flow_control"` on UP, points DOWN at it via `StreamConsumerSource`, publishes to UP, and verifies the message lands on DOWN — i.e. the same end-to-end shape as the Go client's coverage.


renerocksai (approved):
Please bear in mind, I am fairly new to nats-py and might be wrong in my assessments. My review focused on idiomatic / good-quality async Python code; I haven't evaluated NATS-specific design choices.

From that perspective: LGTM!

nats-jetstream/src/nats/jetstream/api/types.py
```diff
@@ -21,9 +21,22 @@ class ExternalStreamSource(TypedDict):
     """The delivery subject to use for the push consumer"""
 
 
+class StreamConsumerSource(TypedDict):
+    """Pre-created push-durable consumer configuration for stream sourcing/mirroring (ADR-60)"""
+
+    name: str
+    """Consumer name"""
+
+    deliver_subject: NotRequired[str]
```
Jarema: both fields are required. Otherwise you will get `10215`

Jarema (changes requested):

aricart (approved):
LGTM

Jarema (approved):
LGTM!

PR #922, opened 2026-05-07
Add atomic batch publish wire support

caspervonb:
Implements ADR-50's atomic batch publish wire surface for nats-server 2.14:

- `StreamConfig.allow_atomic` and `allow_batched` (and the equivalents on `StreamCreateRequest` / `StreamUpdateRequest`).
- `PublishAck.batch_id` / `batch_size`, populated on the commit ack of an atomic batch.
- `nats/jetstream/headers.py` exposes `Nats-Batch-Id`, `-Sequence`, `-Commit` plus value constants `"1"` / `"eob"`, and `Nats-Required-Api-Level`.

End-to-end test: stream with `allow_atomic=True`, publish three messages with `Nats-Batch-Id` (the first two via raw `client.publish`, the third via `jetstream.publish` with `Nats-Batch-Commit: 1`), then assert the commit ack carries the batch id and a batch size of 3 and the stream ends up with three persisted messages.

No high-level batch publish API in this PR — only the wire-level building blocks (matches the Rust client's PR 1581 scope). A future PR can wrap these in an ergonomic batch publisher.

### DIFF
--- a/nats-jetstream/src/nats/jetstream/__init__.py
+++ b/nats-jetstream/src/nats/jetstream/__init__.py
@@ -208,6 +208,10 @@ class PublishAck:
     sequence: int | None = None
     domain: str | None = None
     duplicate: bool = False
+    batch_id: str | None = None
+    """Set on the final ack of an atomic batch publish (ADR-50): id of the committed batch."""
+    batch_size: int | None = None
+    """Set on the final ack of an atomic batch publish (ADR-50): number of messages persisted in the committed batch."""
 
     @classmethod
     def from_response(cls, data: api.PublishAck, *, strict: bool = False) -> PublishAck:
@@ -215,6 +219,8 @@ class PublishAck:
         sequence = data.pop("seq", None)
         domain = data.pop("domain", None)
         duplicate = data.pop("duplicate", False)
+        batch_id = data.pop("batch", None)
+        batch_size = data.pop("count", None)
 
         # Check for unconsumed fields
         if strict and data:
@@ -225,6 +231,8 @@ class PublishAck:
             sequence=sequence,
             domain=domain,
             duplicate=duplicate,
+            batch_id=batch_id,
+            batch_size=batch_size,
         )
 
 
--- a/nats-jetstream/src/nats/jetstream/headers.py
+++ b/nats-jetstream/src/nats/jetstream/headers.py
@@ -2,26 +2,37 @@
 
 from __future__ import annotations
 
+from typing import Final
+
 # Atomic Batch Publishing (ADR-50)
-NATS_BATCH_ID = "Nats-Batch-Id"
+NATS_BATCH_ID: Final[str] = "Nats-Batch-Id"
 """Atomic batch publish: batch id (max 64 chars)."""
 
-NATS_BATCH_SEQUENCE = "Nats-Batch-Sequence"
+NATS_BATCH_SEQUENCE: Final[str] = "Nats-Batch-Sequence"
 """Atomic batch publish: per-message sequence within the batch."""
 
-NATS_BATCH_COMMIT = "Nats-Batch-Commit"
+NATS_BATCH_COMMIT: Final[str] = "Nats-Batch-Commit"
 """Atomic batch publish: commit marker. ``"1"`` to commit and store the
 final message; ``"eob"`` to commit without storing it (end-of-batch)."""
 
-NATS_BATCH_COMMIT_FINAL = "1"
+NATS_BATCH_COMMIT_FINAL: Final[str] = "1"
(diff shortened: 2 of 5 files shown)


nats-jetstream/tests/test_stream.py
```diff
@@ -686,3 +686,76 @@ async def test_stream_state_timestamps(jetstream: JetStream):
...
+        name="BATCH_PUB",
+        subjects=["bp.>"],
+        allow_atomic=True,
+    )
+
+    batch_id = "batch-1"
+
+    # Pre-commit messages are fire-and-forget: the server stores them but
+    # holds the ack until the commit message arrives. Use the underlying
+    # NATS publish (no request/reply) for these.
```
renerocksai: Two readings of this one — I'm honestly not sure which is right and would appreciate a sanity-check from someone closer to the protocol.

**Maybe I'm misreading the ADR**, in which case please ignore the rest of this comment.

**But maybe** ADR-50's "Client Design" section is saying the initial batch message *should* be a request, not a fire-and-forget publish:

> "A batch will be started by adding the `Nats-Batch-Id:uuid` and `Nats-Batch-Sequence:1` headers **using a request**, the server will reply with an error or zero byte message."

If that reading is right, the comment here ("Pre-commit messages are fire-and-forget … the server holds the ack until the commit message arrives") would slightly mislead future readers — per the ADR the initial message gets an immediate zero-byte ack (or an error if the feature isn't supported on this stream/server), and that handshake is also what lets a client detect feature support before committing to the rest of the batch.

If that reading isn't right, then everything's fine — but it might still be worth a one-liner explaining *why* raw publish is used here (e.g. "we deliberately skip the documented initial-request handshake for test simplicity") so the next reader doesn't have to re-derive it from the ADR.

Either way, low priority; the test as written does pass against a 2.14+ server.

nats-jetstream/src/nats/jetstream/headers.py
```diff
@@ -0,0 +1,27 @@
+"""JetStream protocol header names and value constants."""
+
+from __future__ import annotations
+
+# Atomic Batch Publishing (ADR-50)
+NATS_BATCH_ID = "Nats-Batch-Id"
```
renerocksai: Two small consistency nits for this new module:

1. **`Final[str]` annotations on the constants.** Codebase precedent for protocol-level constants is `nats-core/src/nats/client/protocol/message.py:45-52` (`CRLF: Final[bytes]`, `MAX_CONTROL_LINE: Final[int]`, etc.). e.g.:
   ```python
   from typing import Final

   NATS_BATCH_ID: Final[str] = "Nats-Batch-Id"
   ```
2. **An `__all__`.** Sibling `nats-jetstream/src/nats/jetstream/message.py:342` defines one. Adding `__all__` here makes the public surface explicit and avoids `annotations` (from the `from __future__` import) leaking on `import *`.

Both are small polish; take or leave.

nats-jetstream/src/nats/jetstream/__init__.py
```diff
@@ -208,13 +208,19 @@ class PublishAck:
     sequence: int | None = None
     domain: str | None = None
     duplicate: bool = False
+    batch_id: str | None = None
```
renerocksai, line 213: Tangential to the new field here: these wire building blocks (`PublishAck.batch_id`, `StreamConfig.allow_atomic`, etc.) are exposed via `nats.jetstream`, but the matching header name constants live only at `nats.jetstream.headers`. Users of these features have to know the submodule layout to import the header names.

Consider either re-exporting the most commonly used constants from this module's `__all__`, or adding `headers` itself to `__all__` so `from nats.jetstream import headers` is a documented entry-point. Hedge: this is a deliberate API-surface decision and you may prefer to keep the top-level minimal — feel free to drop.

renerocksai (commented):
Please bear in mind, I am fairly new to nats-py and might be wrong in my assessments. The review focused on idiomatic, good-quality async Python code; I haven't evaluated NATS-specific design choices.

Just a few small things, all hedged inline:

1. A test comment that, *if* I'm reading ADR-50 correctly, may not quite match what the ADR specifies for the initial batch message. I might well be misreading the ADR — details inline.

The rest are small consistency tightenings (`Final[str]`, `__all__`, public surface) — take or leave.

nats-jetstream/src/nats/jetstream/__init__.py
caspervonb: Keeping the headers module accessible only via the full `nats.jetstream.headers` path. The top-level surface stays minimal and discoverable; protocol-level constants are a deliberate submodule.

renerocksai (approved):
Fixes LGTM incl. the bonuses!

aricart (approved):

PR #923, opened 2026-05-07
Add message schedule headers

caspervonb:
Implements ADR-51's message schedules wire surface for nats-server 2.14:

- `StreamConfig.allow_msg_schedules` on the TypedDict and dataclass.
- Header module gains the request-side `Nats-Schedule`, `-Target`, `-Source`, `-TTL`, `-Time-Zone`, `-Rollup` plus the server-set `Nats-Scheduler` / `Nats-Schedule-Next`. Cron shorthand constants (`@yearly` … `@hourly`) and the `Nats-Schedule-Rollup` value constant (`"sub"`) ride along.
- Schedule-related error codes (10186–10192, 10203, 10212, 10223) on `ErrorCode`.

End-to-end test: publish a scheduled message via the header constants, refetch it from the stream, and verify the headers round-trip. A second test asserts the server returns `SCHEDULE_TARGET_INVALID` (10190) when the schedule is set without a target.

No high-level schedule API in this PR — only the wire-level building blocks.

Conflicts with #922 only on `nats/jetstream/headers.py`: that PR creates the module with batch-publish constants, this one creates it with schedule constants. At merge time the resolution is just a union of the constants.


nats-jetstream/tests/test_stream.py
```diff
@@ -686,3 +686,71 @@ async def test_stream_state_timestamps(jetstream: JetStream):
...
+    ack = await jetstream.publish(
+        "sched.every",
+        b"",
+        headers={NATS_SCHEDULE: "@every 5s", NATS_SCHEDULE_TARGET: "target.every"},
+    )
+
+    msg = await stream.get_message(ack.sequence)
+    assert msg.headers is not None
+    assert msg.headers.get(NATS_SCHEDULE) == "@every 5s"
+    assert msg.headers.get(NATS_SCHEDULE_TARGET) == "target.every"
```
renerocksai, line 749: Consider also asserting `msg.subject == "sched.every"` (and maybe `msg.data == b""`) here. The current assertions only check the two headers, so if `get_message` ever returned a different message in the stream (e.g. due to a future bug in sequence handling), the test would still pass as long as the wrong record happened to carry those header values. A subject/body check makes the round-trip a bit more explicit at zero extra cost. Minor — feel free to skip.

renerocksai (commented):

One small polish suggestion on the round-trip test, hedged inline — take or leave.

nats-jetstream/src/nats/jetstream/headers.py
```diff
@@ -0,0 +1,59 @@
+"""JetStream protocol header names and value constants."""
+
+from __future__ import annotations
```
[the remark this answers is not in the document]
caspervonb: Keeping `from __future__ import annotations` for consistency — every other module under `nats/jetstream/` has it, and once the `Final[str]` annotations from #922 land here it becomes load-bearing.

renerocksai (approved):
Fixes LGTM! Consistency-reasoning for keeping `from __future__ import annotations` is reasonable!

aricart (commented):
I am not seeing the ability to set the payload from a specific subject to the message dispatched by the schedule

nats-jetstream/src/nats/jetstream/headers.py
```diff
@@ -0,0 +1,59 @@
...
+# server will produce messages on a recurring or one-off basis.
+NATS_SCHEDULE = "Nats-Schedule"
+"""Schedule expression. Cron form (6 fields), an ``@every <duration>`` /
+``@at <RFC3339>`` shorthand, or one of the predefined constants below.
+Requires :class:`StreamConfig.allow_msg_schedules`."""
+
+NATS_SCHEDULE_TARGET = "Nats-Schedule-Target"
+"""Subject the schedule will publish to when it fires."""
+
+NATS_SCHEDULE_SOURCE = "Nats-Schedule-Source"
```
aricart: found it

aricart (approved):

Issue #924, opened 2026-05-08
Header values with non-ASCII bytes silently drop the entire headers dict (#491 still unfixed in 2.14)


NixBiks:
### Observed behavior

`Client._process_headers` ([client.py:1782](https://github.com/nats-io/nats.py/blob/v2.14.0/nats/aio/client.py#L1782)) parses NATS message headers with Python's `email.parser.BytesParser` (compat32 policy, the default when `fast_mail_parser` is not installed). When **any header value contains a non-ASCII byte**, the email parser returns an `email.header.Header` object instead of a plain `str` for that value. The dict-comp on the same line then blows up:

```
File "nats/aio/client.py", line 1782, in _process_headers
  parsed_hdr = {k.strip(): v.strip() for k, v in self._hdr_parser.parsebytes(raw_headers).items()}
                           ^^^^^^^
AttributeError: 'Header' object has no attribute 'strip'
```

The exception is swallowed at line 1796-1798:

```python
except Exception as e:
    await self._error_cb(e)
    return hdr
```

…so the consumer is silently delivered a message with `headers = None` (or `{}` after caller normalization). No retry, no DLQ signal, no backpressure — just lost headers.

### Why this matters

Whatever caller-side validation depends on a header field then breaks. In our case (subject `pipeline.feed_item.matched`, NATS 2.10 broker, JetStream pull consumer, nats-py 2.14.0), 100% of traffic on a particular subject DLQ'd on `KeyError: 'event_id'` because every message had empty headers. Cause: vendor-supplied identifier values in those messages contained Nordic characters (å, ä, ö, ø). The bug is silent at the nats-py layer — only visible because our application code raised on missing keys.

### Reproduction

Minimal repro using the email parser directly:

```python
from email.parser import BytesParser
p = BytesParser()
parsed = p.parsebytes(b"Nats-Msg-Id: ABC\xc2\xa3DEF\r\n\r\n")  # £ in UTF-8
for k, v in parsed.items():
    print(k, type(v).__name__, repr(v))
# Nats-Msg-Id Header <email.header.Header object at 0x...>
```

End-to-end repro via the CLI (from rickdgeerling in #491, which is closed but unfixed):

```
nats request --server localhost:4222 'some-subject' 'some payload' --header 'Test: ø'
```

### Why #491 was closed prematurely

[...]

PR #925, opened 2026-05-08
Replace email.parser path in _process_headers with byte-level parser (#491)

### LINKED ISSUE #491: hdr[k.strip()] = v.strip() AttributeError: 'Header' object has no attribute 'strip'
### What version were you using?

In` nats-py==2.2.` sometimes this error occurs

### What environment was the server running in?

I have seen in linux and iMac with using python3.8

### Is this defect reproducible?

No

### Given the capability you are leveraging, describe your expectation?

-

### Given the expectation, what is the defect you are observing?

-

NixBiks:
Closes #491. Reported again in #924 with confirmed reproduction in 2.14.0.

## Problem

NATS message headers are HTTP-flavoured but not emails — no RFC 2047 encoded-words, no header folding, no charset decoding. The default code path in `Client._process_headers` uses `email.parser.BytesParser` (compat32 policy), which on certain non-ASCII header values returns `email.header.Header` objects instead of plain `str`. The dict-comprehension `{k.strip(): v.strip() for ...}` then raises:

```
AttributeError: 'Header' object has no attribute 'strip'
```

The exception is swallowed by `_default_error_callback` and the consumer is delivered a message with `headers = None` (downstream code coerces to `{}`). Silent data loss — no retry, no DLQ signal, no backpressure.

Minimal repro:

```python
from email.parser import BytesParser
BytesParser().parsebytes(b"Test: \xc2\xa3\r\n").items()
# → [('Test', <email.header.Header object>)]
```

End-to-end repro:

```sh
nats request --server localhost:4222 'some-subject' 'some payload' --header 'Test: ø'
```

## Fix

Add a static `_parse_header_lines(raw: bytes) -> Dict[str, str]` byte-level parser. Splits on `\r\n`, partitions each line on `": "`, decodes ASCII keys and UTF-8 values. Used at both call sites (the heartbeat-with-headers branch at the inline-status path, and the standard header block).

The `fast_mail_parser` opt-in path is left unchanged so users who installed it for the whitespace-key-pruning behaviour relied on by `nats/tests/test_js.py:806` keep working. (Note: that path still has its own correctness gap — `fast_mail_parser` Latin-1-decodes UTF-8 bytes, so e.g. `Svågertorp` comes back as `SvÃ¥gertorp` — but that's a separate concern; this PR is scoped to fixing the default path.) The `BytesParser` import and `_hdr_parser` instance attribute are dropped since nothing else uses them.

## Performance

Benchmarked on a 10-key, 460-byte header block (typical JetStream pull-consumer delivery), 100k iterations on Python 3.13:

| Parser | per-call | relative |
| -- | -- | -- |
| **byte-level (this PR)** | **1.97 µs** | **1×** |
| fast_mail_parser | 3.34 µs | 1.7× slower |
| email.parser (status quo default) | 13.54 µs | 6.9× slower |

So the default path becomes both correct and ~7× faster. fast_mail_parser stays the opt-in fast path, but for most workloads the new default is competitive enough that the dependency adds little value.

## Tests

7 unit tests in `nats/tests/test_client.py::ProcessHeadersTest` covering all branches:

- `test_non_ascii_value_does_not_drop_dict` — the regression case (`£` in a header value)
- `test_standard_block` — typical `event_id: ... \r\n Nats-Msg-Id: ... \r\n` block
- `test_inline_status_only` — `NATS/1.0 404\r\n\r\n`
- `test_inline_status_with_description` — `NATS/1.0 503 No Responders\r\n\r\n`
- `test_heartbeat_with_status_and_headers` — `NATS/1.0 100 Idle Heartbeat\r\nNats-Last-Consumer: ...\r\n`
- `test_value_with_multiple_colons` — guards against `partition(b": ")` greedy splits
- `test_empty_returns_none`

All 7 pass locally. Existing test suite unaffected (the `fast_mail_parser` opt-in branch retains its current behaviour).

## Production validation

Running this exact parser as a downstream monkey-patch in our company's NATS pipeline ([FinancialNewsSystems/soil#502](https://github.com/FinancialNewsSystems/soil/pull/502)) for several hours of staging traffic. Eliminated 100% of header-parse failures (~7800 in 4 h → 0).


NixBiks:
@wallyqs apologies — I accidentally closed this PR. The cause: I'd cloned my fork with `--depth 1` (shallow) earlier in the session, then ran `git reset --soft HEAD~1` to drop an empty retrigger commit. In a shallow clone, that landed on a rootless commit, and the subsequent `git push --force-with-lease` wrote an orphan commit to the branch — which GitHub treats as a force-push that recreates history, permanently locking the PR closed (the API now refuses `reopen`: *"state cannot be changed. The branch was force-pushed or recreated."*).

The fork branch is restored to your rebased commit `2a3fb53` so the patch + tests are intact. Happy to open a new PR pointing at this same branch, with full credit to your rebase, if you're willing to take it as a fresh review. Sorry for the churn.

PR #927, opened 2026-05-08
Examples for Documentation

scottf:


caspervonb (commented):
Solid foundation, and great that this fits a cross-language parity effort with Java (already merged) and Rust (`nats-io/nats.rs#1579`). I read the Java and Rust counterparts side-by-side; a few patterns translate cleanly and a few don't. The biggest one: the Python files port Java's threaded-callback structure, but Python's asyncio model has different ownership semantics than Java's `Dispatcher` and Rust's `tokio::spawn`. A literal port from Java works in Java but is unsafe / non-idiomatic in Python.

**Themes (details inline below):**

1. **`while True: try sub.next(timeout=1) except TimeoutError: break` should be `async for msg in sub:`.** This is a port of Java's `while ((m = sub.nextMessage(1000)) != null)`. Rust's matching pattern is `while let Some(msg) = sub.next().await` — that's the direct equivalent of Python's `async for`. Exception-as-control-flow misrepresents the timeout signal as termination, and breaks the natural pairing of `basics_subscribe.py` with `basics_publish.py` (the subscriber dies after 1s if you start it first).

2. **`asyncio.create_task(handler())` without holding a reference is unsafe.** Tokio owns spawned tasks; Python's loop only weak-refs them. CPython docs say plainly: *"Save a reference to the result of this function, to avoid a task disappearing mid-execution."* The fix is `asyncio.TaskGroup` (Python 3.11+, available since the workspace requires 3.13). Bonus: it gives structured cancellation and removes the need for the trailing `await asyncio.sleep(0.5)` waits.

3. **Trailing `await asyncio.sleep(N)` to let messages arrive is brittle.** Same pattern as Java's `Thread.sleep(N)` and Rust's `tokio::time::sleep`, but in Python it doesn't actually drain anything — it just hopes timing works out. Use an explicit `Event`, message count, or `TaskGroup` completion. `subjects_monitoring.py` already does this right and is the pattern the others should mirror.

4. **README's Rust reference resolves to a real directory but the doc-examples don't exist there yet** (#1579 is open, not merged). Either link the PR by number or wait until it lands before adding the relative path.

**Minor:**
- CI is red on lint (7 import-sort) + format (3 files). `uv run ruff check --fix && uv run ruff format` clears both.
- `request_reply_headers.py` has a small header-copy bug (inline below).

Nothing here blocks the parity intent — the examples *work* on a happy-path local run. But since these ship to readers as the canonical Python idiom, worth investing in the asyncio-native shape now.

examples/docs/README.md
```diff
@@ -0,0 +1,62 @@
+# NATS Documentation Examples
+
+These are the Python equivalents of the docs examples published in the NATS
+documentation. They mirror the [Rust](../../../nats.rs/async-nats/examples)
```
caspervonb, line 4: The relative path resolves to `nats.rs/async-nats/examples/`, which exists but doesn't have the `natsIoDoc` series — those land in `nats-io/nats.rs#1579` (open, by the same author). Either link the Rust PR explicitly until it merges, or drop the Rust reference for now and add it back when #1579 lands. As written, a contributor cloning the workspace and following the link finds generic examples (`pub.rs`, `sub.rs`, `kv.rs`) and gets confused.

examples/docs/basics_subscribe.py
```diff
@@ -0,0 +1,26 @@
...
+    # Subscribe to 'weather.updates' synchronously
+    sub = await nc.subscribe("weather.updates")
+
+    # Process messages
+    while True:
+        try:
+            msg = await sub.next(timeout=1)
+            print(f"Received: {msg.data.decode()}")
+        except TimeoutError:
+            break
```
caspervonb, line 19: This is a port of Java's `while ((m = sub.nextMessage(1000)) != null)`, but Rust uses `while let Some(msg) = sub.next().await` — and the Python equivalent of *that* is `async for msg in sub:`, not exception-as-control-flow on `TimeoutError`. `TimeoutError` here means "no message in the last second", not "the publisher is done" — so on a slow runner or if the user starts this subscriber before `basics_publish.py` (the natural order), it exits silently before any message arrives.

Suggested shape:

```python
sub = await nc.subscribe("weather.updates")
print("Listening on weather.updates (Ctrl-C to exit)")
async for msg in sub:
    print(f"Received: {msg.data.decode()}")
```

If you want it self-contained (publish in the same file then receive once), add an explicit count + break.

examples/docs/queue_groups_basic.py
```diff
@@ -0,0 +1,36 @@
...
+        async for msg in sub:
+            print(f"Worker {name} Received: {msg.data.decode()}")
+
+    sub_a = await nc.subscribe("orders.new", queue="new-orders-queue")
+    sub_b = await nc.subscribe("orders.new", queue="new-orders-queue")
+    sub_c = await nc.subscribe("orders.new", queue="new-orders-queue")
+
+    asyncio.create_task(worker(sub_a, "A"))
+    asyncio.create_task(worker(sub_b, "B"))
+    asyncio.create_task(worker(sub_c, "C"))
```
caspervonb, line 20: Fire-and-forget `asyncio.create_task` is unsafe in Python (unlike `tokio::spawn` in Rust): the event loop only holds a weak reference to the task, so it can be garbage-collected mid-handler. CPython docs: *"Save a reference to the result of this function, to avoid a task disappearing mid-execution."*

`asyncio.TaskGroup` (3.11+, available here) solves this and removes the need for the trailing `await asyncio.sleep(1)` on line 31:

```python
async with asyncio.TaskGroup() as tg:
    tg.create_task(worker(sub_a, "A"))
    tg.create_task(worker(sub_b, "B"))
    tg.create_task(worker(sub_c, "C"))
    await nc.flush(timeout=1)
    for i in range(1, 11):
        await nc.publish("orders.new", f"Order {i}".encode())
    # explicit termination — workers loop forever otherwise
    await asyncio.sleep(0.2)
    for sub in (sub_a, sub_b, sub_c):
        await sub.unsubscribe()
```

Applies to all the other examples that fan out handlers via `create_task` (queue_groups_mixed_subscribers, subjects_multi_wildcard, subjects_single_wildcard, queue_groups_request_reply, request_reply_*).

examples/docs/queue_groups_basic.py
```diff
@@ -0,0 +1,36 @@
...
+    # flush() waits for the server to acknowledge all pending commands.
+    # This ensures all subscriptions are at the server before publish starts.
+    await nc.flush(timeout=1)
+
+    for i in range(1, 11):
+        await nc.publish("orders.new", f"Order {i}".encode())
+    # NATS-DOC-END
+
+    # give time for all the messages to be received
+    await asyncio.sleep(1)
```
caspervonb, line 31: `await asyncio.sleep(1)` here doesn't actually drain handler tasks — it just hopes the 10 messages arrive in 1 second. Java's `Thread.sleep(1000)` worked because the dispatcher's worker threads kept running independently; Python's tasks are cooperative on the same loop and `sleep` yields to *all* of them, so it's effectively a race. Replacing the create_tasks with a `TaskGroup` (see comment above) makes this unnecessary.

examples/docs/queue_groups_dynamic_scaling.py
```diff
@@ -0,0 +1,47 @@
...
+    async def start(self, nc, subject, queue):
+        self.subscription = await nc.subscribe(subject, queue=queue)
+        self.task = asyncio.create_task(self._run())
+
+    async def _run(self):
+        async for msg in self.subscription:
+            print(f"Worker {self.id} processing: {msg.data.decode()}")
+
+    async def stop(self):
+        await self.subscription.unsubscribe()
```
caspervonb, line 22: `stop()` calls `unsubscribe()` but never `cancel()`s `self.task` or awaits it. The handler task is held in `self.task` (good) but then never used (less good) — it'll exit naturally when `_run`'s `async for` ends (the unsubscribe terminates the iterator), which works but isn't structured. A more honest shape:

```python
async def stop(self):
    await self.subscription.unsubscribe()
    if self.task:
        await self.task
```

Also note this example as a whole — scale up 5 workers, then immediately scale them down with no publish in between — produces no observable output. The Java mirror has the same gap. Worth adding a publish loop between the two phases so readers see the load-balancing happen.

examples/docs/request_reply_headers.py
```diff
@@ -0,0 +1,44 @@
...
+    sub = await nc.subscribe("service")
+
+    async def service():
+        async for msg in sub:
+            response_headers: dict[str, str | list[str]] = {}
+            if msg.headers:
+                request_id = msg.headers.get("X-Request-ID") or ""
+                response_headers["X-Response-ID"] = request_id
+                for key, values in msg.headers.items():
+                    response_headers[key] = values
```
caspervonb, line 21: Small bug: setting `response_headers["X-Response-ID"] = request_id` and *then* iterating `for key, values in msg.headers.items(): response_headers[key] = values` will overwrite `X-Response-ID` if the request also carried one (or copy the empty `""` from line 18 if it didn't). Either copy first then override, or skip `X-Request-ID` in the loop:

```python
request_id = msg.headers.get("X-Request-ID") or ""
for key, values in msg.headers.items():
    response_headers[key] = values
response_headers["X-Response-ID"] = request_id
```

examples/docs/subjects_monitoring.py
```diff
@@ -0,0 +1,39 @@
...
+    asyncio.create_task(monitor())
+    await nc.flush()
+
+    # Publish a message to various subjects
+    await nc.publish("hello", b"Hello NATS!")
+    await nc.publish("event.new", b"click")
+    await nc.publish("weather.north.fr", "Temperature: 11°C".encode())
+
+    print("Waiting for messages...")
+    await received.wait()
```
caspervonb, line 32: FYI — this example is the one that gets the synchronization right: explicit `asyncio.Event`, monitor signals it after N messages, main awaits the event. It mirrors Java's `CountDownLatch(3)` cleanly and avoids the trailing `await asyncio.sleep(...)` race. The other examples should mirror this shape (or `TaskGroup`) instead of sleeping at the end.

examples/docs/request_reply_basic.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # NATS-DOC-START
+    # Set up a service
+    sub = await nc.subscribe("time")
+
+    async def time_service():
+        async for msg in sub:
+            now = datetime.now(timezone.utc).isoformat()
+            if msg.reply:
+                await nc.publish(msg.reply, now.encode())
```
caspervonb, line 19: Same fire-and-forget hazard as queue_groups_basic — the request itself happens to synchronize (the requester only completes once the service replies), so this works in practice, but the pattern as written would still drop the task on a slow first request. Either wrap in `TaskGroup` or use `asyncio.ensure_future` and hold the handle. Same applies to `request_reply_calculator.py:27`, `request_reply_headers.py:25`, `request_reply_multiple_responders.py:18-19`, `queue_groups_request_reply.py:21`.

PR #928, opened 2026-05-08
Replace email.parser path in _process_headers with byte-level parser

### LINKED ISSUE #491: hdr[k.strip()] = v.strip() AttributeError: 'Header' object has no attribute 'strip'

NixBiks:
Reopens #925 (which I accidentally locked closed via a `git reset --soft HEAD~1` on a shallow clone — the `reopen` API refuses on force-pushed/recreated branches; details and apology in [#925's last comment](https://github.com/nats-io/nats.py/pull/925#issuecomment-4409300808)).

Same patch as #925, on the same branch, rebased on current main by @wallyqs (commit `2a3fb53`). Closes #491.

## Problem

NATS message headers are HTTP-flavoured but not emails — no RFC 2047 encoded-words, no header folding, no charset decoding. The default code path in `Client._process_headers` uses `email.parser.BytesParser` (compat32 policy), which on certain non-ASCII header values returns `email.header.Header` objects instead of plain `str`. The dict-comprehension `{k.strip(): v.strip() for ...}` then raises:

```
AttributeError: 'Header' object has no attribute 'strip'
```

The exception is swallowed by `_default_error_callback` and the consumer is delivered a message with `headers = None`. Silent data loss — no retry, no DLQ signal, no backpressure.

Minimal repro:

```python
from email.parser import BytesParser
BytesParser().parsebytes(b"Test: \xc2\xa3\r\n").items()
# → [('Test', <email.header.Header object>)]
```

End-to-end repro (from #491 follow-up):

```sh
nats request --server localhost:4222 'some-subject' 'some payload' --header 'Test: ø'
```

## Fix

Add a static `_parse_header_lines(raw: bytes) -> Dict[str, str]` byte-level parser. Splits on `\r\n`, partitions each line on `": "`, decodes ASCII keys (skipping malformed names rather than emitting U+FFFD-laced unreachable keys) and UTF-8 values. Drops keys whose names contain whitespace, matching the existing `fast_mail_parser` post-pass behaviour relied on by `nats/tests/test_js.py:806`.

Used at both call sites (the heartbeat-with-headers branch and the standard header block). The `fast_mail_parser` opt-in path is left unchanged so users who installed it keep working — though note `fast_mail_parser` Latin-1-decodes UTF-8 bytes (e.g. `Svågertorp` → `SvÃ¥gertorp`), so the byte-level parser is more correct for non-ASCII traffic; phasing out the opt-in branch is a follow-up. The `BytesParser` import and `_hdr_parser` instance attribute are dropped — nothing else in the class uses them.

## Performance

Benchmarked on a 10-key, 460-byte header block (typical JetStream pull-consumer delivery), 100k iterations on Python 3.13:

| Parser | per-call | relative |
| -- | -- | -- |
| **byte-level (this PR)** | **1.97 µs** | **1×** |
| fast_mail_parser | 3.34 µs | 1.7× slower |
| email.parser (status quo default) | 13.54 µs | 6.9× slower |

Default path becomes both correct and ~7× faster. fast_mail_parser stays the opt-in fast path, but the dependency adds little value once the default is competitive.

## Tests

8 unit tests in `nats/tests/test_client.py::ProcessHeadersTest` covering all branches:

- `test_non_ascii_value_does_not_drop_dict` — the regression case (`£` in a header value)
- `test_non_ascii_in_name_is_skipped_not_replaced` — bot-review fix; malformed names skip rather than U+FFFD-pollute
- `test_standard_block` — typical multi-key block
- `test_inline_status_only` — `NATS/1.0 404\r\n\r\n`
- `test_inline_status_with_description` — `NATS/1.0 503 No Responders\r\n\r\n`
- `test_heartbeat_with_status_and_headers` — `NATS/1.0 100 Idle Heartbeat\r\n…`
- `test_value_with_multiple_colons` — guards against `partition(b": ")` greedy splits
- `test_empty_returns_none`

Plus `nats/tests/test_js.py::SubscribeTest::test_fetch_headers` was passing on #925's last green run and is unchanged in this PR (relies on the new parser dropping whitespace-in-name keys).

All 8 new tests + the existing suite pass locally. CI on the previous (closed) PR was green except for one timing-flake in `test_subscribe_custom_limits` (sub1's `pending_bytes_limit=15` racing on a slow CI runner — unrelated; passed on rerun).

### DIFF
--- a/nats/src/nats/aio/client.py
+++ b/nats/src/nats/aio/client.py
@@ -25,7 +25,6 @@ import string
 import time
 from collections import UserString
 from dataclasses import dataclass
-from email.parser import BytesParser
 from io import BytesIO
 from pathlib import Path
 from random import shuffle
@@ -303,7 +302,6 @@ class Client:
         self._flush_queue: Optional[asyncio.Queue[asyncio.Future[Any]]] = None
         self._flusher_task: Optional[asyncio.Task] = None
         self._flush_timeout: Optional[float] = 0
-        self._hdr_parser: BytesParser = BytesParser()
 
         # New style request/response
         self._resp_map: Dict[str, asyncio.Future] = {}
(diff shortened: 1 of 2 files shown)


caspervonb:
All three review comments addressed in the current head — `partition(b":")`, `c.isspace()`, and a docstring note on the `errors="replace"` U+FFFD trade-off. PTAL @claude

caspervonb (approved):
This does differ bit between nats-core's parser (which is more strict) but that's probably fine.

lgtm

PR #929, opened 2026-05-10
Add stream persist mode wire support

caspervonb:
Wire-level support for ADR-56 `persist_mode` on `StreamConfig`. Surfaced as a Literal so callers can opt into `async` persistence without an enum import; the server omits the field for the default mode so the round-trip test only asserts the `async` path.


Jarema (approved):
LGTM!

PR #934, opened 2026-05-14
Fix PullSubscription.fetch hang due to orphan lingering request

### LINKED ISSUE #933: PullSubscription.fetch hangs due to orphan pull request
### Observed behavior

There's a race condition in the code where sometimes a call to `PullSubscription.fetch()` will hang for the full timeout duration even when there are messages available.

### Expected behavior

`PullSubscription.fetch` should return immediately once at least one message is received with no further messages available, and should not wait for the timeout.

### Server and client version

nats-server: v2.12.6
nats-py: 2.14.0

### Host environment

_No response_

### Steps to reproduce

```python
"""
MRE: nats-py fetch(batch=N) stalls for ~timeout seconds when a previous
fetch left an orphaned pull request on the server.

Background
----------
When fetch(timeout=T) times out with no messages, _fetch_n may send a
second "lingering" pull request just before raising TimeoutError.  This
happens when the server's 408 REQUEST_TIMEOUT (sent at expires=T-100µs)
arrives at the client before Python's asyncio timer fires at T — a window
of ~100µs.  That second lingering request is never cancelled and remains
on the server as an orphan.

When the next fetch() call is made, the server now has two outstanding
pull requests.  NATS routes incoming messages to the *oldest* outstanding
request — the orphan.  The current fetch()'s no-wait probe receives the
message delivery, but _fetch_n then sits in a loop waiting for the
remaining batch=N-1 messages that will never arrive, only exiting when the
probe's own expires fires (~T seconds later).

Reproduction
------------
A short-timeout fetch is used to trigger orphan creation reliably (the
100µs race is consistently won by the server on localhost).  The following
fetch then demonstrates the stall.

Expected: fetch() returns within ~1 second of a message being published.
Actual:   fetch() stalls for ~timeout seconds (here ~5 s) before returning.

Tested with: nats-py 2.14.0, nats-server v2.12.6
Run against a plain `nats-server -js`.
"""
import asyncio
import time

import nats

NATS_URL = "nats://localhost:4222"
STREAM   = "mre-fetch-stall"
SUBJECT  = "mre.fetch.stall"


async def run_once(js, sub, iteration: int) -> float:
    # Step 1: fetch with a short timeout on an empty stream.
    # The server's 408 (sent at expires = 100ms - 100µs) arrives at the client
    # before Python's asyncio timer fires, causing _fetch_n to send a second
    # lingering pull request that is immediately abandoned — leaving an orphan
    # on the server.
    try:
        await sub.fetch(100, timeout=0.1)
    except Exception:
        pass  # expected: TimeoutError or FetchTimeoutError

    # Step 2: start a fresh fetch (5 s timeout) then publish one message.
    # The server routes the message to the orphaned pull request (oldest).
    # The current fetch's no-wait probe has no messages, so it waits out its
    # full 5-second expiry before returning the one message it already has.
    fetch_task = asyncio.create_task(sub.fetch(100, timeout=5.0))
    await asyncio.sleep(0.05)           # let the probe reach the server
    await js.publish(SUBJECT, b"hello")
    t0   = time.monotonic()
    msgs = await fetch_task
    elapsed = time.monotonic() - t0

    for msg in msgs:
        await msg.ack()

    flag = "  <-- BUG (expected < 1s)" if elapsed > 1.0 else ""
    print(f"  iter {iteration:2d}: fetch() returned {len(msgs)} msg in {elapsed:.3f}s{flag}")
    return elapsed


async def main():
    nc  = await nats.connect(NATS_URL)
    js  = nc.jetstream()

    try:
        await js.delete_stream(STREAM)
    except Exception:
        pass
    await js.add_stream(name=STREAM, subjects=[SUBJECT])
    sub = await js.pull_subscribe(SUBJECT, "mre-consumer", stream=STREAM)

    results = []
    for i in range(5):
        results.append(await run_once(js, sub, i))

    slow = sum(1 for r in results if r > 1.0)
    print(f"\n{slow}/{len(results)} iterations stalled (> 1 s after publish)")

    await js.delete_stream(STREAM)
    await nc.close()


asyncio.run(main())
```
output:
```
  iter  0: fetch() returned 1 msg in 4.952s  <-- BUG (expected < 1s)
  iter  1: fetch() returned 1 msg in 4.951s  <-- BUG (expected < 1s)
  iter  2: fetch() returned 1 msg in 4.952s  <-- BUG (expected < 1s)
  iter  3: fetch() returned 1 msg in 4.952s  <-- BUG (expected < 1s)
  iter  4: fetch() returned 1 msg in 4.949s  <-- BUG (expected < 1s)

5/5 iterations stalled (> 1 s after publish)
```

<sup>(Disclosure: This was fully written by Claude Code. Figuring this issue out was a major PITA, so I just set claude at it, and this is what it produced)</sup>

phemmer:
This fixes an issue in the nats-py code where a race condition occurs that leaves an orphaned lingering request on the server, causing `fetch` to hang until the timeout expires.

fixes #933

### DIFF
--- a/nats/src/nats/js/client.py
+++ b/nats/src/nats/js/client.py
@@ -1203,9 +1203,17 @@ class JetStreamContext(JetStreamManager):
 
             # First request: Use no_wait to synchronously get as many available
             # based on the batch size until server sends 'No Messages' status msg.
+            # Omit `expires` when the drain step already found messages: NATS
+            # server ignores no_wait when expires is present, treating the probe
+            # as a lingering pull and blocking for the full expires duration.
+            # Without expires the server honors no_wait immediately, so Phase 3
+            # returns quickly with any additional server-side messages or a 404,
+            # and the existing `if len(msgs) > 0` guard returns the collected
+            # messages without delay. When the drain step found nothing expires
+            # is still included to preserve the intended behaviour.
             next_req = {}
             next_req["batch"] = needed
-            if expires:
+            if expires and not msgs:
                 next_req["expires"] = expires
             if heartbeat:
                 next_req["idle_heartbeat"] = int(heartbeat * 1_000_000_000)  # to nanoseconds
--- a/nats/src/nats/js/client.py
+++ b/nats/src/nats/js/client.py
@@ -1265,9 +1265,29 @@ class JetStreamContext(JetStreamManager):
 
             # Second request: lingering request that will block until new messages
             # are made available and delivered to the client.
+            #
+            # Use the *remaining* deadline as the request's expires rather than
+            # the original full timeout.  The original expires was computed at
+            # the very start of fetch() and may be nearly exhausted by the time
+            # we reach this point (e.g. when the server's 408 for the no-wait
+            # probe arrives just before the asyncio timer fires).  Sending a
+            # lingering request with the full original expires in that situation
+            # creates an orphaned pull request that survives on the server long
+            # after the client has timed out, capturing the next published
+            # message and causing the subsequent fetch() to stall for the full
+            # timeout window.
+            deadline = JetStreamContext._time_until(timeout, start_time)
+            if deadline is not None and deadline <= 0:
+                raise asyncio.TimeoutError
+
             next_req = {}
             next_req["batch"] = needed
-            if expires:
+            if deadline is not None:
+                remaining_expires = int(deadline * 1_000_000_000) - 100_000
+                if remaining_expires <= 0:
+                    raise asyncio.TimeoutError
+                next_req["expires"] = remaining_expires
+            elif expires:
                 next_req["expires"] = expires
             if heartbeat:
                 next_req["idle_heartbeat"] = int(heartbeat * 1_000_000_000)  # to nanoseconds
@@ -1335,6 +1355,13 @@ class JetStreamContext(JetStreamManager):
                     if JetStreamContext._is_heartbeat(status):
                         got_any_response = True
                         continue
+                    if status in (
+                        api.StatusCode.NO_MESSAGES,
+                        api.StatusCode.REQUEST_TIMEOUT,
+                    ):
+                        # No more messages will be delivered on this pull
+                        # request; return what we have.
+                        break
                     if JetStreamContext._is_processable_msg(status, msg):
                         needed -= 1
                         msgs.append(msg)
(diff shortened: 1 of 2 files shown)


phemmer:
Added another fix. Completely separate issue, but related, so throwing it in this same PR. ab04fa6d4371ead8b0a4d30fb9745336b0d41cb7
Basically if there are messages in `_pending_queue` at the start of `fetch()`, it would call the server to ask for more, and then hang for the full timeout.

Also one thing that might be worth discussing: I preserved the original behavior in that if messages are sitting in `_pending_queue` when `fetch()` is called, it still calls the server to request more messages. I don't know that this is the ideal behavior. I would think that if messages are available in the local buffer, that we would want to return them immediately without having to perform a network call.

nats/src/nats/js/client.py
```diff
@@ -1265,9 +1273,29 @@ async def _fetch_n(
...
+            # we reach this point (e.g. when the server's 408 for the no-wait
+            # probe arrives just before the asyncio timer fires).  Sending a
+            # lingering request with the full original expires in that situation
+            # creates an orphaned pull request that survives on the server long
+            # after the client has timed out, capturing the next published
+            # message and causing the subsequent fetch() to stall for the full
+            # timeout window.
+            deadline = JetStreamContext._time_until(timeout, start_time)
+            if deadline is not None and deadline <= 0:
+                raise asyncio.TimeoutError
```
phemmer, line 1289: This might also be worth discussing. This and line 1296 below raise `asyncio.TimeoutError`. This is consistent with line 1234 above. However I question whether this is the right behavior. There's also [`FetchTimeoutError`](https://github.com/nats-io/nats.py/blob/ab04fa6d4371ead8b0a4d30fb9745336b0d41cb7/nats/src/nats/js/errors.py#L149). So the current code is consistent, but is it consistently wrong? Should it be `FetchTimeoutError`? And fix line 1234 as well?

caspervonb (approved):

caspervonb:
Merged, looks good. Thank you very much!

nats/src/nats/js/client.py
superlevure: FYI we just upgraded to 2.15.0 in production and we're hitting many `asyncio.TimeoutError` that are uncaught because we expect `nats.errors.TimeoutError` instead. We have to revert to `2.14.0` because of this since we can't update all our applications to catch this new Exception.

superlevure: Tentative fix: https://github.com/nats-io/nats.py/pull/1008

PR #937, opened 2026-05-15
Backport sourcing-consumer config for ADR-60

caspervonb:
ADR-60 lets clients pre-create a push-durable consumer for sourcing/mirroring from workqueue or interest streams. This adds `StreamSource.consumer` (`StreamConsumerSource(name, deliver_subject)`) and `AckPolicy.FLOW_CONTROL`, which the server requires on that consumer — without it the source is accepted but fails at runtime with `stream source consumer requires flow control ack policy` (10217).

Includes an end-to-end test sourcing from a workqueue stream (skipped on servers older than 2.14).


nats/src/nats/js/api.py
```diff
@@ -163,6 +166,25 @@ class PubAck(Base):
...
+            resp["batch_size"] = resp.pop("count")
+        return super().from_response(resp)
+
+    def as_dict(self) -> Dict[str, object]:
+        result = super().as_dict()
+        if "batch_id" in result:
+            result["batch"] = result.pop("batch_id")
+        if "batch_size" in result:
+            result["count"] = result.pop("batch_size")
+        return result
```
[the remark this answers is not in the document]
caspervonb: The batch-publish fields were split out into #938 in a later rebase — this PR's current diff is pure ADR-60 (nats-jetstream only).

caspervonb:
@claude please review again

caspervonb:
PTAL @claude

caspervonb:
PTAL @claude — the branch has moved since the last review: merged `main`, added `AckPolicy.FLOW_CONTROL` (the server requires it on the pre-created sourcing consumer, otherwise the source fails at runtime with 10217), and an end-to-end test sourcing from a workqueue stream.

nats/src/nats/js/api.py
```diff
@@ -586,6 +602,10 @@ class AckPolicy(str, Enum):
     NONE = "none"
     ALL = "all"
     EXPLICIT = "explicit"
+    # Required on the pre-created consumer used for sourcing/mirroring from a
+    # workqueue or interest stream (ADR-60). The sourcing stream, not a
+    # client, drives acknowledgements.
+    FLOW_CONTROL = "flow_control"
```
renerocksai, line 608: Now that `FLOW_CONTROL` is constructible from Python, it might be worth a look at the one place in the legacy client that branches on ack policy — `nats/src/nats/js/client.py:497`:

```python
if cb and (not manual_ack) and (config.ack_policy is not api.AckPolicy.NONE):
    cb = self._auto_ack_callback(cb)
```

As written, a callback push-subscription bound to an ADR-60 sourcing consumer picks up the auto-ack wrapper and sends a per-message ack reply, whereas the ADR has the sourcing server drive acknowledgement through flow-control messages carrying `Nats-Last-Stream` / `Nats-Last-Consumer` (*"the flow control messages will ack messages rather than the current ack reply format"*).

I think this needs `FLOW_CONTROL` sitting alongside `NONE` in that guard, but I'd hedge two ways: it's a narrow path (you'd have to deliberately `subscribe_bind` to a sourcing consumer rather than let the server manage it), and I can't tell from static reading whether the server simply ignores an unexpected ack on such a consumer. Very possibly a non-issue in practice — just flagging it since this PR is what makes the fourth variant reachable.

nats/tests/test_js.py
```diff
@@ -5501,6 +5501,110 @@ def test_opt_start_time_non_utc_timezone_preserved_on_parse(self):
...
+        consumer = info.config.sources[0].consumer
+        assert isinstance(consumer, nats.js.api.StreamConsumerSource)
+        assert consumer.name == "C"
+        assert consumer.deliver_subject == "deliver.up"
+
+        for i in range(3):
+            await js.publish("up", f"msg-{i}".encode())
+
+        for _ in range(50):
+            info = await js.stream_info("DOWN")
```
renerocksai: Cosmetic: `info` gets rebound here from the `add_stream` result to the `stream_info` result, and the earlier binding is still what `consumer = info.config.sources[0].consumer` reads from on line 5589. A separate name for the polled value — `down_info` or similar — would keep the two apart:

```python
for _ in range(50):
    down_info = await js.stream_info("DOWN")
    if down_info.state.messages == 3:
        break
    await asyncio.sleep(0.1)
assert down_info.state.messages == 3
assert down_info.sources[0].error is None
```

renerocksai (approved):

Nothing blocking came up. Two small tightenings — hedged inline, take or leave.

nats/src/nats/js/api.py
caspervonb: Good catch — checked it against 2.14 and it is real, though not in the way either of us guessed. Binding a callback subscription to a `flow_control` consumer and letting the auto-ack wrapper run:

```
received=3 ack_floor=0 delivered=3 num_ack_pending=3 stream_msgs=3
explicit ack_sync -> MsgAlreadyAckdError
```

So on your hedge: the server does simply ignore the acks — `ack_floor` stays 0, `num_ack_pending` stays 3, and the workqueue messages are not removed. No corruption. But the client side is not harmless: it publishes a useless ack per message and sets `msg._ackd`, so a caller that later acks properly gets `MsgAlreadyAckdError` on a message the server never considered acked.

Fixed in 1f2a0847 by putting `FLOW_CONTROL` alongside `NONE` in that guard. Same scenario afterwards no longer pre-acks, and an explicit ack now reaches the server (`NoRespondersError`, since there is no ack responder for such a consumer) rather than being swallowed locally.

I also switched the guard from `is not` to `not in` while there: legacy does not coerce `ack_policy` to the enum, so `consumer_info` hands back a plain `"flow_control"`/`"none"` string, and an identity check would miss it. 46 subscribe/ack tests still pass.

nats/tests/test_js.py
caspervonb: Fixed in 1f2a0847 — the polled value is `down_info` now, so the `add_stream` result the consumer assertions read from stays intact.

PR #938, opened 2026-05-15
Add atomic batch publish headers and PubAck fields

caspervonb:
Backport of ADR-50 atomic batch publish completion fields (PR #922 in nats-jetstream) so legacy users of `allow_atomic`/`allow_batched` streams can read the commit ack.


nats/src/nats/js/api.py
```diff
@@ -163,6 +166,25 @@ class PubAck(Base):
...
+    batch_size: Optional[int] = None
+
+    @classmethod
+    def from_response(cls, resp: Dict[str, Any]) -> PubAck:
+        # Server uses ``batch``/``count`` for atomic batch publish (ADR-50).
+        if "batch" in resp and "batch_id" not in resp:
+            resp["batch_id"] = resp.pop("batch")
+        if "count" in resp and "batch_size" not in resp:
+            resp["batch_size"] = resp.pop("count")
+        return super().from_response(resp)
```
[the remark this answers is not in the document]
caspervonb: Keeping the existing pop()-mutating pattern from the base class for consistency — every other from_response helper does the same. Worth revisiting as a base-class-wide change rather than a one-off here.

caspervonb:
@claude please review again

nats/tests/test_js.py
```diff
@@ -4991,6 +4991,42 @@ def test_opt_start_time_non_utc_timezone_preserved_on_parse(self):
         )
 
 
+class PubAckBatchTest(unittest.TestCase):
```
renerocksai, line 5504: Two small suggestions on the tests, both take-or-leave:

1. I think these could fold into the existing `APIDataClassTest` (test_js.py:34) rather than a new class here. That class is already the home for pure dataclass `from_response`/`as_dict` tests in exactly this shape (`test_api_stats_level_present`, `test_stream_config_subject_delete_marker_ttl_as_dict`), and it sits with the other no-server tests at the top of the file — whereas line 4994 lands in the middle of the `SingleJetStreamServerTestCase` block. Folding them in would also pick up that class's one-line-docstring-per-test convention for free.

2. `test_pub_ack_as_dict_maps_batch_fields` (line 5016) only exercises the populated path. The `if "batch_id" in result` guards in `as_dict` are there precisely because `Base.as_dict()` skips `None` fields, but nothing currently pins that behaviour — consider asserting that a batch-less `PubAck(stream="TEST", seq=1).as_dict()` contains neither `batch` nor `count`. A round-trip assertion would cover both directions in one line:

```python
assert nats.js.api.PubAck.from_response(ack.as_dict()) == ack
```

renerocksai (approved):

Nothing material came up. Two small test tightenings — hedged inline, take or leave.

nats/tests/test_js.py
caspervonb: Added both the absent-field and round-trip assertions. I kept the dedicated pure unittest class because the current branch has no `APIDataClassTest`; this class also uses no server fixture.

PR #939, opened 2026-05-15
Add message schedule header constants

caspervonb:
Backport of nats-jetstream PR #923. Adds ADR-51 schedule header constants to the legacy JetStream module so existing nats-py users can set scheduled-publish headers (`Nats-Schedule`, `Nats-Schedule-Target`, etc.) without hard-coding strings. The `allow_msg_schedules` StreamConfig flag was already present.


nats/src/nats/js/api.py
```diff
@@ -34,9 +34,24 @@ class Header(str, Enum):
...
+    SCHEDULER = "Nats-Scheduler"
     STATUS = "Status"
 
 
+# Predefined schedule expressions for use as a Header.SCHEDULE value.
+SCHEDULE_YEARLY = "@yearly"
+SCHEDULE_MONTHLY = "@monthly"
+SCHEDULE_WEEKLY = "@weekly"
+SCHEDULE_DAILY = "@daily"
+SCHEDULE_HOURLY = "@hourly"
```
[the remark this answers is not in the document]
caspervonb: Deferring SCHEDULE_ROLLUP_SUB for now — keeping the legacy backport minimal. The constant lives in nats-jetstream for callers who need named parity.

caspervonb:
@claude please review again

nats/src/nats/js/api.py
```diff
@@ -34,9 +34,24 @@ class Header(str, Enum):
...
+    SCHEDULE_SOURCE = "Nats-Schedule-Source"
+    SCHEDULE_TARGET = "Nats-Schedule-Target"
+    SCHEDULE_TIME_ZONE = "Nats-Schedule-Time-Zone"
+    SCHEDULE_TTL = "Nats-Schedule-TTL"
+    SCHEDULER = "Nats-Scheduler"
     STATUS = "Status"
 
 
+# Predefined schedule expressions for use as a Header.SCHEDULE value.
+SCHEDULE_YEARLY = "@yearly"
```
renerocksai, line 52: Small idiomatic nit, take or leave: every other closed set of string values in this module is a `(str, Enum)` — `Header`, `StatusCode`, `RetentionPolicy`, `StorageType`, `DiscardPolicy`, `StoreCompression`, `PersistMode`, `AckPolicy`, `DeliverPolicy`, `ReplayPolicy`. The only flat module constants in `api.py` are `DEFAULT_PREFIX` / `INBOX_PREFIX`, and those are one-offs rather than a family. I think these five would read better grouped the same way:

```python
class SchedulePreset(str, Enum):
    YEARLY = "@yearly"
    MONTHLY = "@monthly"
    WEEKLY = "@weekly"
    DAILY = "@daily"
    HOURLY = "@hourly"
```

It also disambiguates at the call site: flat `api.SCHEDULE_DAILY` lives in the same namespace as `Header.SCHEDULE_TTL` and so reads like a header *name*, where it's actually a header *value*. `SchedulePreset.DAILY` can't be misread that way.

Two caveats on my own suggestion: `kv.py` does use flat value constants (`KV_OP`, `KV_DEL`, `KV_PURGE`, `MSG_ROLLUP_SUBJECT`), so flat isn't unprecedented in the package — just not in this module. And copying the nats-jetstream `Final[str]` shape isn't really available here anyway, since `nats-py` declares `requires-python = ">=3.7"` and `typing.Final` is 3.8+, so an enum is the version-safe way to get the same grouping.

nats/tests/test_js.py
```diff
@@ -4776,6 +4776,27 @@ def test_cluster_info_from_response_without_leader_since(self):
         assert info.traffic_acc is None
 
 
+class ScheduleHeadersTest(unittest.TestCase):
+    """Unit tests for ADR-51 message schedule header constants."""
+
+    def test_schedule_header_values(self):
+        assert nats.js.api.Header.SCHEDULE.value == "Nats-Schedule"
```
renerocksai: Minor: these assert `.value`, but every consumer uses the member directly as a header key — `kv.py:351`, `object_store.py:337`, `js/client.py:205` — and `aio/client.py` then calls `k.strip()` / `key.encode()` on it. Dropping `.value` (`assert nats.js.api.Header.SCHEDULE == "Nats-Schedule"`) pins the wire value *and* the `str`-mixin contract those call sites depend on, for the same line count.

renerocksai (approved):

Nothing material came up. Two small tightenings — hedged inline, take or leave.

nats/src/nats/js/api.py
caspervonb: Keeping these as module-level constants to mirror the public shape in nats.jetstream.headers; introducing a legacy-only enum would create an avoidable API divergence. I added that rationale next to the constants.

PR #940, opened 2026-05-15
Add counter stream wire support

caspervonb:
Backport of ADR-49 counter stream wire support from the modern nats-jetstream package (PR #926).


nats/src/nats/js/api.py
```diff
@@ -164,6 +165,10 @@ class PubAck(Base):
     domain: Optional[str] = None
     duplicate: Optional[bool] = None
 
+    # Current value of the counter on counter-enabled streams (ADR-49).
+    # String to preserve precision beyond uint64.
+    val: Optional[str] = None
```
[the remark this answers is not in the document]
caspervonb: Leaving 'val' as-is — legacy stays close to the wire JSON; nats-jetstream is the place for the renamed 'value'.

caspervonb:
@claude please review again

nats/src/nats/js/api.py
renerocksai, line 171: Small nit on the comment rather than the code: "String to preserve precision beyond uint64" is the reason the *wire* format is a string (JS `Number` tops out at 2^53, the server uses `big.Int`), but Python's `int` is arbitrary precision — so a reader landing here may wonder why the field isn't just `int`. I think the Python-side reason worth stating is fidelity: hand back exactly what the server sent and let the caller decide with `int(ack.val)`.

Maybe something like:

```python
# Current value of the counter on counter-enabled streams (ADR-49).
# Kept as the raw string the server sent; callers can convert with int(ack.val).
```

nats/src/nats/js/api.py
```diff
@@ -29,6 +29,7 @@ class Header(str, Enum):
     EXPECTED_LAST_SEQUENCE = "Nats-Expected-Last-Sequence"
     EXPECTED_LAST_SUBJECT_SEQUENCE = "Nats-Expected-Last-Subject-Sequence"
     EXPECTED_STREAM = "Nats-Expected-Stream"
+    INCR = "Nats-Incr"
```
renerocksai, line 33: Consider a short comment on the accepted format here. Reading ADR-49 literally the value must match `^[+-]\d+$`, so the sign looks mandatory and `headers={Header.INCR: "1"}` would be rejected — which isn't obvious from the constant alone, unlike `MSG_ID` or `EXPECTED_STREAM` which are self-describing. One line noting `+N` / `-N` would save someone a round trip.

Hedging on this one: the ADR also says "A value of `0` is valid", which doesn't match its own regex, so I may well be over-reading it. Equally fine to document the format wherever a publish-side helper eventually lands instead.

nats/tests/test_js.py
```diff
@@ -4991,6 +4991,28 @@ def test_opt_start_time_non_utc_timezone_preserved_on_parse(self):
         )
 
 
+class CounterStreamTest(unittest.TestCase):
```
renerocksai, line 5483: Two small tightenings on the new tests:

**1. Coverage is asymmetric on the direction users actually hit.** `StreamConfig` gets `as_dict()` coverage (lines 4997 and 5002) but not `from_response()`, so nothing exercises reading `allow_msg_counter` back off a `stream_info` response — which is how a user would check whether a stream is a counter stream. Since the deliverable here is wire plumbing in both directions, I think that direction deserves an assertion too:

```python
def test_stream_config_from_response_with_allow_msg_counter(self):
    config = nats.js.api.StreamConfig.from_response(
        {"name": "c", "subjects": ["c"], "allow_msg_counter": True}
    )
    assert config.allow_msg_counter is True
```

**2. The `"42"` on line 5008 doesn't show what the field is for.** The field comment justifies `str` by precision beyond uint64, but this value would read identically if the type were `int`. A value past 2^64 (`"18446744073709551616"`) would document that intent at a glance. Purely cosmetic — it changes nothing behaviorally, since the `== "42"` comparison already pins the type.

renerocksai (approved):

Nothing material came up — just a few small tightenings, hedged inline, take or leave.

nats/src/nats/js/api.py
caspervonb: Documented the decimal delta with accepted examples (`"5"` and `"-2"`). The modern counter-stream tests confirm an unsigned positive value is accepted, so I avoided implying that a leading `+` is mandatory.

PR #941, opened 2026-05-15
Add consumer reset wire support

caspervonb:
Backport ADR-60's `$JS.API.CONSUMER.RESET` (PR #920 in the modern jetstream package) so legacy users can drop pending state on a workqueue/interest consumer without recreating it. Exposed as `JetStreamManager.reset_consumer`; surfaces error code 10204 as `ConsumerInvalidResetError`.


caspervonb:
@claude please review again

nats/src/nats/js/manager.py
```diff
@@ -321,6 +321,70 @@ async def resume_consumer(
         # Resume by pausing until a time in the past (epoch)
         return await self.pause_consumer(stream, consumer, "1970-01-01T00:00:00Z", timeout)
 
+    async def reset_consumer(
```
renerocksai, line 324: The docstring covers `Args:` / `Returns:` / `Note:` but not `Raises:`, and this is the only method that raises the `ConsumerInvalidResetError` the PR introduces — as it stands `errors.py` is the only place a user can discover that it exists. The modern `Stream.reset_consumer` documents it, so it might be worth mirroring:

```
        Raises:
            ConsumerInvalidResetError: If the requested reset violates the
                consumer's DeliverPolicy (e.g. ``seq`` below ``opt_start_seq``
                on a bounded policy).
```

Hedged, though: the legacy package doesn't use `Raises:` anywhere today, so this would be introducing the convention rather than following it.

nats/tests/test_js.py
```diff
@@ -2048,6 +2048,150 @@ async def test_consumer_pause_with_immediate_expiry(self):
...
+            ack_policy="explicit",
+        )
+
+        sub = await js.pull_subscribe_bind(consumer_name, "RESETTEST")
+        msgs = await sub.fetch(3, timeout=2)
+        assert len(msgs) == 3
+
+        # Delete one of the in-flight messages from the stream so it cannot
+        # be redelivered; reset should drop it from ack pending.
+        delivered = await jsm.consumer_info("RESETTEST", consumer_name)
```
renerocksai: Two small readability things here:

1. `delivered` holds a `ConsumerInfo`, and `reset.info.delivered` is asserted a few lines below — the name reads as though the two are the same concept. `info` would avoid the collision.

2. The comment on the two lines above ("Delete one of the in-flight messages…") sits on top of the `consumer_info` fetch but describes the `delete_msg` call two statements down. Moving it next to what it explains would read better.

And one on the setup more broadly: both the WORK_QUEUE retention and the `delete_msg` step are commented as being what makes the assertions meaningful, but if I'm reading ADR-60 right, pending and redelivered are *always* cleared on reset — so `num_ack_pending == 0` / `num_redelivered == 0` would hold without either step. I think either the assertions could be pushed a bit to actually cover the deletion (something on `ack_floor.stream_seq` or `num_pending`), or the comments trimmed so they don't claim more than the test checks.

nats/tests/test_js.py
```diff
@@ -2048,6 +2048,150 @@ async def test_consumer_pause_with_immediate_expiry(self):
...
+            "num_ack_pending": 0,
+            "num_redelivered": 0,
+            "num_waiting": 0,
+            "num_pending": 0,
+        }
+
+        with self.assertRaises(KeyError):
+            nats.js.api.ConsumerReset.from_response(resp)
+
+    def test_consumer_invalid_reset_error_from_error_dict(self):
```
renerocksai: A few idiomatic tightenings in this one:

- The module already does `from nats.js.errors import *` at the top, so the function-local `from nats.js import errors as js_errors` isn't needed.
- The `err` dict is built only to be unpacked field-by-field three lines later — passing the literals straight in is shorter and just as clear.
- The `try:` / `except:` wraps a `raise` the test performs itself, so the except clause is guaranteed to run and the structure isn't asserting anything the assertions couldn't make on a plain instance. The sibling test above uses `self.assertRaises`, which is the other consistent option.

Something like:

```python
    def test_consumer_invalid_reset_error(self):
        """ConsumerInvalidResetError carries the API error fields for err_code 10204."""
        exc = ConsumerInvalidResetError(code=400, description="consumer reset is invalid", err_code=10204)

        assert exc.code == 400
        assert exc.err_code == 10204
        assert exc.description == "consumer reset is invalid"
        assert isinstance(exc, BadRequestError)
        assert isinstance(exc, APIError)
```

Also a naming nit — `..._from_error_dict` suggests the value goes through `APIError.from_error`, which it doesn't.

renerocksai (approved):

Nothing material came up. A few small tightenings — hedged inline, take or leave.

PR #943, opened 2026-05-18
Add nats.js to nats.jetstream migration guide

caspervonb:
Mirrors `nats-core/MIGRATION.md` and `nats-key-value/MIGRATION.md` for the JetStream surface: only the differences from legacy `nats.js`, so users can translate code piece by piece instead of relearning the whole API.


aricart (approved):
LGTM

PR #949, opened 2026-05-27
Expose verbose and pedantic on connect()

caspervonb:
These CONNECT fields were hard-coded with no user override; expose them as `connect()` kwargs so users can opt into server +OK acks or strict checks. The values are also threaded through the reconnect path so they survive a CONNECT rebuild.


caspervonb:
PTAL @claude — addressed the +OK protocol handling (parser, read loop, both handshake paths), dropped the protocol kwarg, merged main, and added real end-to-end verbose/pedantic tests. Requesting a final review.

nats-core/src/nats/client/__init__.py
```diff
@@ -1847,7 +1867,10 @@ async def connect(
     await connection.write(encode_ping())
 
     try:
-        response = await asyncio.wait_for(parse(connection), timeout=timeout)
+        while True:
+            response = await asyncio.wait_for(parse(connection), timeout=timeout)
+            if not isinstance(response, Ok):
+                break
 
         if isinstance(response, Err):
```
renerocksai, line 1927: **Asymmetric handshake completion checks vs. the reconnect path**

There are two places in the client that do effectively the same thing after sending `CONNECT + PING`: this one (initial connect) and the reconnect path inside `_force_disconnect` (around lines 949–974). Both now drain stray `+OK` lines with a `while True` loop — symmetric and good.

The asymmetry is in what each path does **after** the loop exits.

The reconnect path checks three things and bails out if any look wrong:
1. `response is None` → server closed the socket mid-handshake → raise
2. `response` is `Err` → server rejected us → raise
3. `response` is anything other than `Pong` → unexpected → raise

The initial-connect path here only checks `Err`. If the server cleanly closes mid-handshake (`parse()` returns `None`), or sends back something unexpected, this code falls through silently and proceeds to construct a `Client` with status `CONNECTED` — even though the handshake never completed. The first publish/subscribe on that client would then fail with a confusing downstream error instead of a clear "connection closed during handshake" raised here at `connect()` time.

This isn't something the PR introduced — the gap pre-existed. But the PR touches these exact lines, and the symmetric (more defensive) version is right there in the same file as a model. Closing it would be ~4 lines copy-pasted from the reconnect path, e.g.:

```python
if response is None:
    await connection.close()
    msg = "Connection closed before PONG received"
    raise ConnectionError(msg)

if isinstance(response, Err):
    # ... existing handling ...

if not isinstance(response, Pong):
    await connection.close()
    msg = f"Unexpected response to PING: {type(response).__name__}"
    raise ConnectionError(msg)
```

I'm genuinely unsure whether this should block the PR or become a separate follow-up issue/PR — happy either way. Flagging it because we're already in this code.

renerocksai (commented):

One small observation inline — about the initial-connect handshake having less defensive completion checks than the reconnect handshake (pre-existing asymmetry, but the PR touches that code). I'm not sure whether this should block this PR or become a follow-up issue/PR — flagging it so you can decide.

nats-core/src/nats/client/__init__.py
caspervonb: Already extracted most of the common things into establish_connection, so gonna punt to a follow up with extracting the handshake as-well so both paths do the same thing.

nats-core/tests/test_client.py
```diff
@@ -1707,6 +1713,206 @@ async def handle(reader: asyncio.StreamReader, writer: asyncio.StreamWriter) ->
...
+        subject = f"test.verbose.request.{uuid.uuid4()}"
+
+        async def responder():
+            subscription = await client.subscribe(subject)
+            await client.flush()
+            message = await subscription.next(timeout=1.0)
+            assert message.reply is not None
+            await client.publish(message.reply, b"pong")
+
+        task = asyncio.create_task(responder())
```
renerocksai, line 1865: Small consistency nit (take or leave): the responder subscribes *inside* the task (line 1859), then the main coroutine flushes (1866) and immediately issues `client.request` (1868). The file's other two responder tests subscribe *before* launching the task — `test_request_reply_concurrent` (subscribe → `create_task`) and the JWT request test (subscribe → `flush` → `request`).

Putting the subscribe inside the task leaves no happens-before between the responder's `SUB` and the request's `PUB`, so there's a theoretical SUB-before-PUB ordering race. In practice the `flush()` PING/PONG round-trip almost certainly lets the SUB reach the wire first, so I doubt this actually flakes — it's mostly about matching the convention already used elsewhere in the file. Consider subscribing + flushing before `create_task(responder())` and having the responder only consume.

renerocksai (approved):

Just one small test tightening — hedged inline, take or leave.

PR #951, opened 2026-05-27
Name write-buffer cap and flush-interval constants

caspervonb:
Replace magic numbers in `Client.__init__` with named `Final` constants so the write-buffer caps and flush interval are self-documenting.


caspervonb:
PTAL @claude — addressed both review notes: dropped the redundant `1 *` arithmetic and prefixed the constants with `_` to keep them internal rather than exporting them.

piotrpio (approved):
LGTM!

PR #954, opened 2026-05-27
Make Headers a case-insensitive MutableMapping

caspervonb:
Per ADR-21, NATS header names follow HTTP-style case-insensitive semantics. `nats.client.message.Headers` now inherits `collections.abc.MutableMapping[str, str]` with case-insensitive lookup/containment/delete while preserving the original casing on iteration. Multi-value access is retained via `get_all` / `append` / new `multi_items`.


caspervonb:
PTAL @claude — addressed all three: reject empty value lists in `__init__` (ValueError), removed the `items()` override in favor of the inherited `ItemsView`, and dropped `delete()` so deletes go through `del`/`pop` with consistent `KeyError` semantics. Tests updated accordingly.

nats-core/tests/test_message.py
```diff
@@ -13,41 +15,55 @@ def test_headers_init():
...
+
     # Invalid header values
     with pytest.raises(TypeError):
         Headers({"key1": 123})  # type: ignore[dict-item]
 
     with pytest.raises(ValueError):
         Headers({"key1": ["value1", 123]})  # type: ignore[list-item]
 
+    # An empty value list would leave the mapping in an invalid state
+    with pytest.raises(ValueError):
```
renerocksai: Minor test-rigor nit (take or leave): `__init__` raises `ValueError` for two distinct reasons — a non-string list item (line 42, `"All items in header value list must be strings"`) and an empty list (here, `"Header value list for ... must not be empty"`). Both are asserted with a bare `pytest.raises(ValueError)`, so either test would still pass if the wrong branch fired. Consider pinning each with `match=` so they verify the intended branch:

```python
with pytest.raises(ValueError, match="must not be empty"):
    Headers({"key1": []})
```

(and similarly `match="must be strings"` on the line-42 case.)

renerocksai (approved):

One small test-rigor nit — hedged inline, take or leave.

FYI (non-blocking, not a code issue): the PR description cites ADR-21, but ADR-21 is "NATS Configuration Contexts" — the case-insensitive header semantics look like they're specified in a different ADR. Might be worth fixing the reference.

nats-core/tests/test_message.py
caspervonb: Done in 00772aa2 — pinned both with `match=` (`must be strings` and `must not be empty`) so each verifies its intended branch. Thanks!

PR #955, opened 2026-05-27
Don't log the CONNECT payload

caspervonb:
Operators enabling DEBUG to diagnose connection issues would otherwise leak `auth_token`, `password`, `jwt`, `nkey`, and `sig` values into logs. nats.go and async-nats never log the CONNECT payload at all; do the same here rather than redacting field-by-field, which leaves no field to forget when new credentials are added.


caspervonb:
PTAL @claude

aricart (approved):
LGTM - claude had some additional concerns

caspervonb:
@claude please review

PR #961, opened 2026-05-28
Add Subscription.unsubscribe_after for capped delivery

caspervonb:
Adds the missing user-facing API for `UNSUB <sid> <max_msgs>` so callers can ask "deliver exactly N more messages, then stop" without polling a counter and calling `unsubscribe()` themselves. Cap survives reconnect — the resub path re-issues `UNSUB <sid> remaining` (or drops the subscription if the cap is already met), matching nats.go's `resendSubscriptions`.


caspervonb:
@claude please re-review. Addressed both prior comments:
- Slow-consumer cap divergence: `_handle_msg`/`_handle_hmsg` now close the subscription when `_delivered + _dropped_messages >= _max_messages`, with a regression test.
- Reconnect resub: `remaining` is now computed once and reused.

Also renamed `max_msgs` → `max_messages` for naming consistency, and merged latest `main`.

caspervonb:
Non issue that queue shutdown isn't caught, because it isn't in the execution path. Tests would immediately flag this if it ever did.

caspervonb:
PTAL @claude

nats-core/src/nats/client/subscription.py
```diff
@@ -247,6 +294,44 @@ async def unsubscribe(self) -> None:
...
+        this multiple times replaces the previous cap (the server overwrites and
+        the local cap follows suit). If ``delivered`` already meets or exceeds
+        ``max_messages``, the subscription is closed right away.
+
+        On reconnect the cap is re-issued as ``UNSUB <sid> remaining``, where
+        ``remaining = max_messages - delivered``. If ``remaining <= 0`` the
+        subscription is not resubscribed.
+
+        Args:
+            max_messages: Number of messages to receive before auto-unsubscribing.
```
renerocksai, line 324: **Document slow-consumer-drop interaction with the cap**

The `_handle_msg` / `_handle_hmsg` fix to close when `delivered + dropped >= max_messages` is correct — without it, a slow-consuming sub with drops would stay stuck open forever once the server hit its cap. But the user-facing docs here still promise "Number of messages to receive before auto-unsubscribing", and the docstring body says "the subscription closes itself once :attr:`delivered` reaches it" — neither acknowledges that drops count against the cap.

A user who calls `unsubscribe_after(10)`, sees the sub auto-close, and then reads `subscription.delivered == 4` (because 6 were dropped) is likely to be confused. The behavior is intentional — `test_unsubscribe_after_closes_when_cap_met_by_dropped_messages` explicitly asserts `delivered == 2` with `max_messages=3` — so it's just a doc gap.

A one-line note would close the loop, e.g.:

> Note: messages dropped due to slow-consumer limits also count against the cap (the server counts every message it routes, not just successfully enqueued ones). So `delivered` may be less than `max_messages` when the subscription closes.

Same for the `delivered` property's docstring at line 150 — could mention this case alongside the existing "not those dropped due to slow consumer limits" phrasing.

renerocksai (approved):

LGTM — one small polish note on the `unsubscribe_after` docstring inline, take or leave.

PR #970, opened 2026-06-04
Install nats-server via go install in legacy test job

caspervonb:
The legacy `nats` test jobs fail with `OSError: [Errno 8] Exec format error` because `binaries.nats.dev` resolves `latest` to a main-branch commit that has no prebuilt binary — it returns an empty `200` response, so the installed `nats-server` is a 0-byte file that can't exec. This affects every recent run, including pushes to `main`.

Switch the legacy job to build a release with `go install`, the same way the `project` job already installs nats-server. Drops the dependency on `binaries.nats.dev` and the now-unused `install_nats.sh` step.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,17 +15,25 @@ jobs:
     nats:
         runs-on: ubuntu-latest
         timeout-minutes: 20
-        env:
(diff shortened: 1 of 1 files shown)


Jarema (approved):
LGTM

PR #972, opened 2026-06-04
Renew expired TLS test certificates

caspervonb:
The server and client leaf certificates in `nats-core/tests/certs` and `nats/tests/certs` expired on 2026-06-03, so every TLS and WebSocket test fails with `certificate verify failed: certificate has expired`.

Re-signed both leaf certs with the existing CA and private keys, preserving the subjects, SANs (`DNS:localhost`, `IP:127.0.0.1`), and key usage. Validity now runs to 2035 (within the CA's window) instead of one year, so this stops recurring annually.

Supersedes #971, which was accidentally merged into the #970 branch instead of `main`.


Jarema (approved):
LGTM

wallyqs (approved):

PR #973, opened 2026-06-04
Release nats 2.15.0

caspervonb:
Minor release of the `nats-py` client. Bumps version to 2.15.0 and adds a changelog covering everything since v2.14.0.

### DIFF
--- a/nats/pyproject.toml
+++ b/nats/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "nats-py"
-version = "2.14.0"
+version = "2.15.0"
 authors = [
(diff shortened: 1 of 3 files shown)


nats/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "uv_build"
 
 [project]
 name = "nats-py"
-version = "2.14.0"
+version = "2.15.0"
```
wallyqs, line 7: @claude can you check if there is anywhere else in the repo to bump version like this

wallyqs (approved):
LGTM

caspervonb:
Re-generated uv.lock

PR #982, opened 2026-06-22
Use get_running_loop in the client

caspervonb:
`asyncio.get_event_loop()` is soft-deprecated (and slower). Every `.time()` call site here runs inside a running loop — the client is only constructed from `async connect()`, the rest are in coroutines — so `get_running_loop()` is the correct modern form.

### DIFF
--- a/nats-core/src/nats/client/__init__.py
+++ b/nats-core/src/nats/client/__init__.py
@@ -449,12 +449,12 @@ class Client(AbstractAsyncContextManager["Client"]):
         self._max_pending_bytes = _DEFAULT_PENDING_BYTES_LIMIT
         self._max_pending_messages = _DEFAULT_PENDING_MESSAGES_LIMIT
         self._min_flush_interval = _DEFAULT_MIN_FLUSH_INTERVAL
-        self._last_flush = asyncio.get_event_loop().time() - self._min_flush_interval
+        self._last_flush = asyncio.get_running_loop().time() - self._min_flush_interval
         self._flush_waker = asyncio.Event()
         self._ping_interval = ping_interval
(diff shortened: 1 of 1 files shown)


nats-core/src/nats/client/__init__.py
```diff
@@ -449,12 +449,12 @@ def __init__(
         self._max_pending_bytes = _DEFAULT_PENDING_BYTES_LIMIT
         self._max_pending_messages = _DEFAULT_PENDING_MESSAGES_LIMIT
         self._min_flush_interval = _DEFAULT_MIN_FLUSH_INTERVAL
-        self._last_flush = asyncio.get_event_loop().time() - self._min_flush_interval
+        self._last_flush = asyncio.get_running_loop().time() - self._min_flush_interval
```
renerocksai: Minor polish (take or leave): `__init__` calls `asyncio.get_running_loop()` twice in the constructor — here and again at line 457. Consider binding it once to a local, which also matches the pattern already used at line 1095 (`loop = asyncio.get_running_loop()`):

```python
loop = asyncio.get_running_loop()
self._last_flush = loop.time() - self._min_flush_interval
...
self._last_pong_received = loop.time()
```

The duplicate lookup is negligible, so this is purely cosmetic.

renerocksai (approved):

One small polish nit — hedged inline, take or leave.

nats-core/src/nats/client/__init__.py
caspervonb: Done in 738b30de — bound `loop = asyncio.get_running_loop()` once and reused `loop.time()` for both, matching the line-1095 pattern. Thanks!

PR #993, opened 2026-07-14
Resolve WebSocket close task when connect never established

### LINKED ISSUE #969: `Client.close()` hangs forever after a failed WebSocket connect
### Observed behavior

`WebSocketTransport.wait_closed()` awaits `self._close_task`, but `WebSocketTransport.close()`
only assigns a real close task to `_close_task` **when `self._ws` is set**. If a wss connection
attempt fails before the WebSocket is established (TLS handshake reset, connection refused, DNS
failure, etc.), `_ws` is `None`, `close()` early-returns, and `_close_task` stays the unresolved
`asyncio.Future()` created in `__init__`. `await self._close_task` then never returns, so any
caller of `Client.close()` on such a client **hangs indefinitely**.

## Offending code

`nats/aio/transport.py` — `WebSocketTransport` (2.14.0 line numbers):

```python
def __init__(self, ...):
    self._ws: Optional[aiohttp.ClientWebSocketResponse] = None
    self._client: aiohttp.ClientSession = aiohttp.ClientSession()
    self._close_task = asyncio.Future()        # L194: unresolved Future

async def wait_closed(self):
    await self._close_task                      # L248-249: awaits it
    if self._client:
        await self._client.close()
    self._ws = self._client = None

def close(self):
    if not self._ws:
        return                                  # L254-256: _ws is None → returns, _close_task stays unresolved
    self._close_task = asyncio.create_task(self._ws.close())
```

`Client._close()` (`nats/aio/client.py`) unconditionally calls both on a transport that exists:

```python
if self._transport is not None:
    self._transport.close()
    try:
        await self._transport.wait_closed()     # hangs here forever
    except Exception as e:
        await self._error_cb(e)
```

`self._transport` is created in `_connect_to_server`/`_select_next_server` **before** the
`ws_connect` that fails, so it is non-`None` even though `_ws` was never set.

### Expected behavior

Regardless of whether the WebSocketTransport fails or succeeds, the `close()`call should just close the connection.

## Versions

- Affected: **2.13.1** and **2.14.0** (latest at time of writing). Code is identical in both.
- Python 3.13, `allow_reconnect=True`, transport `wss://` (WebSocketTransport, aiohttp).

### Host environment

_No response_

## Minimal example
```python
import asyncio
from nats.aio.transport import WebSocketTransport

async def main():
    t = WebSocketTransport()      # _ws is None; _close_task is an unresolved Future
    t.close()                     # early-returns (no _ws); _close_task left unresolved
    await asyncio.wait_for(t.wait_closed(), timeout=2)  # raises TimeoutError — hangs

asyncio.run(main())
```

Output: `asyncio.TimeoutError` — `wait_closed()` never returns.

## Larger example
```python
import asyncio
import nats

async def main():
    nc = nats.NATS()
    try:
        await nc.connect(
            "wss://127.0.0.1:1",          # nothing listening → ws_connect fails before _ws is set
            allow_reconnect=True,
            max_reconnect_attempts=1,
            reconnect_time_wait=0,
            connect_timeout=1,
        )
    except Exception as e:
        print("connect failed as expected:", type(e).__name__)

    await asyncio.wait_for(nc.close(), timeout=10)   # hangs → TimeoutError

asyncio.run(main())
```

caspervonb:
`WebSocketTransport.close()` early-returned when `_ws` was `None`, leaving the `_close_task` future unresolved, so a wss connect that failed before the socket was established hung `Client.close()` forever. Resolve the task in that case, mirroring the TcpTransport fix.

Fixes #969.

### DIFF
--- a/nats/src/nats/aio/transport.py
+++ b/nats/src/nats/aio/transport.py
@@ -264,6 +264,12 @@ class WebSocketTransport(Transport):
 
     def close(self):
         if not self._ws:
+            # A connect attempt that failed before the websocket was
+            # established leaves _ws unset. Resolve _close_task so a later
+            # wait_closed() does not block forever on the pending future
+            # created in __init__.
+            if not self._close_task.done():
+                self._close_task.set_result(None)
             return
         self._close_task = asyncio.create_task(self._ws.close())
 
--- a/nats/src/nats/aio/transport.py
+++ b/nats/src/nats/aio/transport.py
@@ -257,19 +257,14 @@ class WebSocketTransport(Transport):
             await self._ws.send_bytes(message)
 
     async def wait_closed(self):
-        await self._close_task
+        if self._ws is not None:
+            await self._close_task
         if self._client:
             await self._client.close()
         self._ws = self._client = None
 
     def close(self):
-        if not self._ws:
-            # A connect attempt that failed before the websocket was
-            # established leaves _ws unset. Resolve _close_task so a later
-            # wait_closed() does not block forever on the pending future
-            # created in __init__.
-            if not self._close_task.done():
-                self._close_task.set_result(None)
+        if self._ws is None:
             return
         self._close_task = asyncio.create_task(self._ws.close())
 
--- a/nats/tests/test_client_websocket.py
+++ b/nats/tests/test_client_websocket.py
@@ -37,6 +37,20 @@ class WebSocketTest(SingleWebSocketServerTestCase):
         await nc.close()
 
     @async_test
+    async def test_close_after_failed_connect_does_not_hang(self):
+        # Regression for #969: if a wss connect fails before the websocket is
+        # established, _ws stays None; close()/wait_closed() must not hang on
+        # the unresolved _close_task created in __init__.
+        if not aiohttp_installed:
+            pytest.skip("aiohttp not installed")
+
+        from nats.aio.transport import WebSocketTransport
+
+        transport = WebSocketTransport()
+        transport.close()
+        await asyncio.wait_for(transport.wait_closed(), timeout=2)
+
+    @async_test
     async def test_request_with_headers(self):
         if not aiohttp_installed:
             pytest.skip("aiohttp not installed")
--- a/nats/tests/test_client_websocket.py
+++ b/nats/tests/test_client_websocket.py
@@ -1,8 +1,11 @@
 import asyncio
 import unittest
+from unittest import mock
 
 import pytest
 from nats.aio.errors import *
+from nats.aio.transport import WebSocketTransport
+from nats.errors import NoServersError
 
 import nats
 from tests.utils import *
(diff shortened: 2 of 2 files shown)


nats/src/nats/aio/transport.py
```diff
@@ -264,6 +264,12 @@ async def wait_closed(self):
 
     def close(self):
         if not self._ws:
+            # A connect attempt that failed before the websocket was
+            # established leaves _ws unset. Resolve _close_task so a later
+            # wait_closed() does not block forever on the pending future
+            # created in __init__.
+            if not self._close_task.done():
```
renerocksai: Not blocking — the fix works. But I think this guard is doing something slightly different from what it reads as.

`_close_task` holds two structurally different objects: after `__init__` it's a bare `asyncio.Future` you complete by hand, and after a connected `close()` it's a `Task` wrapping `ws.close()`, which you must *not* call `set_result` on (`Task.set_result` raises `RuntimeError`). So `.done()` here looks like re-entrancy protection but is really disambiguating those two states. I couldn't find a reachable path where a *pending* Task lands in this branch — `_ws` is only cleared in `wait_closed()` after the await, and both call sites (`client.py:801-806` and `1595-1600`) do `close()` then `wait_closed()` back-to-back — so this isn't a live bug, just a trap a later edit could walk into.

Worth noting alongside it: on the connected path `close()` *replaces* `_close_task` on line 274, so the `__init__` Future is dropped while still pending. Harmless (a GC'd pending Future doesn't warn), but it means that sentinel only ever does anything on the failure path.

`TcpTransport` handles the same problem from the other side — its `wait_closed()` guards with `if self._io_writer is not None`. A closer mirror here would be a guard in `WebSocketTransport.wait_closed()` (skip the `await self._close_task` when nothing was ever established), which avoids manufacturing a result at all and keeps `_close_task` meaning exactly one thing. Entirely your call.

While in this branch: `if not self._ws:` on line 266 would read better as `if self._ws is None:`. `_ws` is `Optional[aiohttp.ClientWebSocketResponse]` and that class defines neither `__bool__` nor `__len__`, so the two are equivalent — but `is None` states the actual intent, and it's the form `TcpTransport` uses two methods up.

nats/tests/test_client_websocket.py
```diff
@@ -36,6 +36,20 @@ async def test_simple_headers(self):
...
+    async def test_close_after_failed_connect_does_not_hang(self):
+        # Regression for #969: if a wss connect fails before the websocket is
+        # established, _ws stays None; close()/wait_closed() must not hang on
+        # the unresolved _close_task created in __init__.
+        if not aiohttp_installed:
+            pytest.skip("aiohttp not installed")
+
+        from nats.aio.transport import WebSocketTransport
+
+        transport = WebSocketTransport()
```
renerocksai: A couple of small things, all hedged:

1. This pins the changed line but not the reported symptom. Constructing a bare `WebSocketTransport()` skips the failed connect entirely, so the wiring from #969 — `nats.connect("wss://…")` failing and then `Client.close()` hanging on `_transport.wait_closed()` at `client.py:801-806` — isn't exercised. `_connect_to_server` assigns `self._transport` *before* awaiting the connect, so the transport is live when the connect raises; driving it through `nats.connect(...)` inside a `pytest.raises` and then `await asyncio.wait_for(nc.close(), timeout=2)` would cover that path too. (The unit test isn't wrong, just narrow.)

2. The test lives in `WebSocketTest(SingleWebSocketServerTestCase)`, whose `setUp` starts a ws-enabled `nats-server` that this test never touches — it pays a server start/stop for a pure unit test.

3. Very minor: the function-local `from nats.aio.transport import WebSocketTransport` on line 47 isn't needed for import safety — `nats/aio/transport.py` already guards its `aiohttp` import (transport.py:9-14), so it's import-safe without aiohttp and could sit at the top with the other imports.

renerocksai (approved):

Nothing blocking came up. Two hedged suggestions inline — one on the shape of `_close_task`, one on how much of #969 the new test actually covers — both take-or-leave.

PR #994, opened 2026-07-14
Raise publish_async ack errors on the returned future

### LINKED ISSUE #985: publish_async ack errors are swallowed (logged as "nats: encountered error") instead of being raised on the returned future
### Observed behavior

When using `js.publish_async()`, an error returned by the server (e.g. a 503 `ServiceUnavailableError`) is **not** propagated to the future returned by `publish_async()`. Instead the exception leaks into the internal subscription task and is logged as `nats: encountered error`, and the future is **never resolved** — so `await`-ing it can neither return nor raise.

## Root cause

In `nats/js/client.py`, `_handle_async_reply` expects `APIError.from_error()` to **return** an error object so it can be attached via `set_exception`:

```python
# nats/js/client.py (~L168-169)
err = nats.js.errors.APIError.from_error(resp["error"])
future.set_exception(err)
```

But `APIError.from_error()` in `nats/js/errors.py` (~L79-91) **raises** in every branch instead of returning:

```python
@classmethod
def from_error(cls, err):
    code = err["code"]
    if code == 503:
        raise ServiceUnavailableError(**err)   # raises, never returns
    ...
```

So line 168 throws before `set_exception` (line 169) is reached. The exception propagates out of `_handle_async_reply` into `_wait_for_msgs`, where it is logged as `nats: encountered error`, and the caller's future is left pending forever.

**Expected:** `await ack` raises `ServiceUnavailableError` (caught by the `except`).  
**Actual:** error is logged as `nats: encountered error`; future never resolves; `except` never fires.

## Suggested fix

Either make `APIError.from_error()` **return** the constructed error instead of raising it (and keep `raise from_error(...)` at call sites that want to raise), or construct the exception directly in `_handle_async_reply`:

```python
if "error" in resp:
    try:
        APIError.from_error(resp["error"])
    except APIError as err:
        future.set_exception(err)
    return
```

Note that the synchronous `publish()` path is unaffected, because it calls `from_error` directly under `raise` within the caller's coroutine.

### Server and client version

nats-py 2.15.0

### Host environment

_No response_

## Minimal repro

```python
import asyncio
import nats
from nats.js.errors import ServiceUnavailableError

async def main():
    nc = await nats.connect("nats://127.0.0.1:4222")
    js = nc.jetstream()

    # A stream/subject configured to reject the publish, e.g.
    # max_msgs_per_subject=1 so the 2nd publish to the same subject
    # returns 503 err_code=10077 "maximum messages per subject exceeded".
    ack = await js.publish_async("packets.same.subject", b"payload")
    try:
        await ack  # expected: ServiceUnavailableError raised here
    except ServiceUnavailableError as e:
        print("caught as expected:", e)   # never reached
    # Instead: "nats: encountered error" is logged from the background task,
    # and `await ack` hangs because the future is never resolved.

    await nc.close()

asyncio.run(main())
```

caspervonb:
`APIError.from_error` raises rather than returns, so the error branch in `_handle_async_reply` threw before `set_exception`; the error escaped into the delivery task and the ack future was left pending forever. Capture and attach it to the future.

Fixes #985.

### DIFF
--- a/nats/src/nats/js/client.py
+++ b/nats/src/nats/js/client.py
@@ -165,19 +165,21 @@ class JetStreamContext(JetStreamManager):
         try:
             resp = json.loads(msg.data)
             if "error" in resp:
-                # APIError.from_error raises rather than returning, so capture
-                # the constructed error and attach it to the future instead of
-                # letting it escape and leave the future pending forever.
-                try:
-                    raise nats.js.errors.APIError.from_error(resp["error"])
-                except nats.js.errors.APIError as err:
-                    future.set_exception(err)
-                return
+                # Raises rather than returning; the handler below attaches the
+                # error to the future.
+                nats.js.errors.APIError.from_error(resp["error"])
 
             ack = api.PubAck.from_response(resp)
             future.set_result(ack)
         except (asyncio.CancelledError, asyncio.InvalidStateError):
             pass
+        except Exception as err:
+            # Anything escaping here would strand the future: its done callback
+            # is what releases the semaphore permit and clears the pending
+            # entry, so publish_async_completed() would block forever and the
+            # permit would never come back. Resolve the future instead.
+            if not future.done():
+                future.set_exception(err)
 
     async def publish(
         self,
--- a/nats/src/nats/js/client.py
+++ b/nats/src/nats/js/client.py
@@ -165,8 +165,13 @@ class JetStreamContext(JetStreamManager):
         try:
             resp = json.loads(msg.data)
             if "error" in resp:
-                err = nats.js.errors.APIError.from_error(resp["error"])
-                future.set_exception(err)
+                # APIError.from_error raises rather than returning, so capture
+                # the constructed error and attach it to the future instead of
+                # letting it escape and leave the future pending forever.
+                try:
+                    raise nats.js.errors.APIError.from_error(resp["error"])
+                except nats.js.errors.APIError as err:
+                    future.set_exception(err)
                 return
 
             ack = api.PubAck.from_response(resp)
--- a/nats/tests/test_js.py
+++ b/nats/tests/test_js.py
@@ -194,6 +194,23 @@ class PublishTest(SingleJetStreamServerTestCase):
         await nc.close()
 
     @async_test
+    async def test_publish_async_error_ack_is_raised(self):
+        # Regression for #985: a server error ack must be raised on the
+        # returned future rather than being swallowed (logged and left
+        # pending forever).
+        nc = NATS()
+        await nc.connect()
+        js = nc.jetstream()
+        await js.add_stream(name="AERR", subjects=["aerr"])
+
+        # Expected-stream mismatch forces the server to return an error ack.
+        future = await js.publish_async("aerr", b"data", stream="WRONGSTREAM")
+        with pytest.raises(nats.js.errors.APIError):
+            await asyncio.wait_for(future, timeout=2)
+
+        await nc.close()
+
+    @async_test
     async def test_publish_msg_ttl(self):
         """Test per-message TTL feature (requires NATS Server 2.11+)"""
         nc = NATS()
--- a/nats/tests/test_js.py
+++ b/nats/tests/test_js.py
@@ -206,8 +206,41 @@ class PublishTest(SingleJetStreamServerTestCase):
 
         # Expected-stream mismatch forces the server to return an error ack.
         future = await js.publish_async("aerr", b"data", stream="WRONGSTREAM")
-        with pytest.raises(nats.js.errors.APIError):
+        with pytest.raises(BadRequestError) as exc_info:
             await asyncio.wait_for(future, timeout=2)
+        self.assertEqual(exc_info.value.code, 400)
+        self.assertEqual(exc_info.value.err_code, 10060)
+
+        # Resolving the future is also what releases the pending slot, so a
+        # regression strands the publish as well as swallowing the error.
+        await asyncio.wait_for(js.publish_async_completed(), timeout=2)
+        self.assertEqual(js.publish_async_pending(), 0)
+
+        await nc.close()
+
+    @async_test
+    async def test_publish_async_unparsable_ack_does_not_strand_future(self):
+        # An ack the client cannot parse must be routed to the future too.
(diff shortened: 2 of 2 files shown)


nats/src/nats/js/client.py
```diff
@@ -165,8 +165,13 @@ async def _handle_async_reply(self, msg: Msg) -> None:
         try:
             resp = json.loads(msg.data)
```
renerocksai, line 166: I think this `try` block still leaves the non-error branch exposed to the very bug this PR fixes. The handler at the bottom catches only `asyncio.CancelledError` / `asyncio.InvalidStateError`, but two statements inside it can raise something else:

- `json.loads(msg.data)` → `json.JSONDecodeError` (a `ValueError`) on a payload that isn't JSON
- `api.PubAck.from_response(resp)` on line 177 → `TypeError`, since `Base.from_response` ends in `cls(**params)` and `PubAck.stream` / `PubAck.seq` have no defaults

Either one propagates out of the subscription callback into `Subscription._wait_for_msgs`, which hands it to `error_cb` and carries on — so the future is never resolved, `handle_done` never runs, the token stays in `_publish_async_futures`, the semaphore permit is never released, and `publish_async_completed()` blocks forever. Same failure mode as #985, just reached down a different branch.

Since it's the same defect class, it might be worth having the fallback route to the future as well:

```python
except (asyncio.CancelledError, asyncio.InvalidStateError):
    pass
except Exception as err:
    if not future.done():
        future.set_exception(err)
```

That keeps the current behaviour for the two existing cases and stops anything else from stranding the future.

nats/tests/test_js.py
```diff
@@ -193,6 +193,23 @@ async def test_publish_async(self):
...
+        # returned future rather than being swallowed (logged and left
+        # pending forever).
+        nc = NATS()
+        await nc.connect()
+        js = nc.jetstream()
+        await js.add_stream(name="AERR", subjects=["aerr"])
+
+        # Expected-stream mismatch forces the server to return an error ack.
+        future = await js.publish_async("aerr", b"data", stream="WRONGSTREAM")
+        with pytest.raises(nats.js.errors.APIError):
```
renerocksai: Two small tightenings, both hedged:

1. `pytest.raises(nats.js.errors.APIError)` pins only the base class that every JS API error inherits from. The point of the fix is that the *specific* error, with its fields intact, now reaches the caller — so asserting that would be stronger, e.g. `with pytest.raises(...) as exc_info:` plus a check on `exc_info.value.err_code` / `.code` (or a `match=` on the description). If an expected-stream mismatch comes back as a 400, `BadRequestError` would be the concrete type.

2. The other half of the regression isn't covered. Before the fix the future never completed, so `handle_done` never ran: `publish_async_pending()` stayed at 1 and `publish_async_completed()` would hang. `test_publish_async` just above already asserts exactly that, so mirroring it here would lock in the leak side too:

```python
await asyncio.wait_for(js.publish_async_completed(), timeout=2)
self.assertEqual(js.publish_async_pending(), 0)
```

(Wrapped in `wait_for` so a regression fails rather than hangs.)

Very minor: the file does `from nats.js.errors import *` and the neighbouring tests use the bare names (`NoStreamResponseError`, `TooManyStalledMsgsError`), so an unqualified `APIError` here would match the surrounding style.

renerocksai (commented):

1. I think the same "future left pending forever" hole is still open on the non-error branch of `_handle_async_reply` — the fallback handler only catches `CancelledError`/`InvalidStateError`, so a `JSONDecodeError` out of `json.loads` or a `TypeError` out of `PubAck.from_response` escapes exactly the way the `APIError` did. Details inline.

2. The root cause looks like `APIError.from_error` being an unannotated always-raising factory, sitting right under `from_msg(...) -> NoReturn`. No inline for this one since `nats/src/nats/js/errors.py` isn't in the diff: annotating `from_error` as `-> NoReturn` (errors.py:80) would turn `err = APIError.from_error(...)` into something a type checker can flag, rather than a runtime surprise that needs a comment to explain. The larger alternative — having `from_error` *return* the error and updating the three `raise APIError.from_error(...)` call sites — would drop the try/except round-trip and the comment entirely, but it changes public behaviour, so quite possibly out of scope for a fix PR.

3. The rest are small test tightenings — hedged inline, take or leave.

nats/src/nats/js/client.py
caspervonb: You are right, and it reproduces. Driving a malformed reply through `_handle_async_reply` on 2.14:

```
non-JSON payload           -> handler RAISED JSONDecodeError | pending=1 | HUNG (future stranded)
JSON missing stream/seq    -> handler RAISED TypeError       | pending=2 | HUNG (future stranded)
```

Note `pending` climbing: the permit is never returned, so with `publish_async_max_pending=2` the next `publish_async` deadlocks outright.

Fixed in ec65a71c, essentially your suggestion. Since the fallback now routes everything to the future, `from_error` can just raise into it and the inner `try`/`except` disappears — which also settles the bot's dead-`raise` nit above:

```python
        try:
            resp = json.loads(msg.data)
            if "error" in resp:
                nats.js.errors.APIError.from_error(resp["error"])
            ack = api.PubAck.from_response(resp)
            future.set_result(ack)
        except (asyncio.CancelledError, asyncio.InvalidStateError):
            pass
        except Exception as err:
            if not future.done():
                future.set_exception(err)
```

Same output after the fix: `pending=0`, `completed` for both payloads.

nats/tests/test_js.py
caspervonb: Both done in ec65a71c.

1. Confirmed the concrete error rather than guessing — an expected-stream mismatch comes back as `BadRequestError`, `code=400`, `err_code=10060`, `"expected stream does not match"`. The test now asserts `BadRequestError` plus `code`/`err_code`, unqualified to match the surrounding style.

2. Added the leak half (`publish_async_completed()` under `wait_for` + `publish_async_pending() == 0`), and a second test `test_publish_async_unparsable_ack_does_not_strand_future` covering the branch from your other comment. It patches `PubAck.from_response` to raise, then asserts the exception reaches the caller, the pending slot is released, and a subsequent publish still succeeds. Verified it fails without the fix.

PR #998, opened 2026-08-06
Organize workspace development dependency groups

caspervonb:
Split shared development tooling into composable workspace dependency groups while preserving member-local development dependencies.

- Add workspace `test` and `check` groups, with `dev` including both.
- Keep member-specific `dev` groups so `uv sync --dev` works from every workspace member under uv 0.12.
- Include the optional dependencies needed by the workspace type check.
- Run workspace checks explicitly on Python 3.13, matching the workspace Python constraint instead of silently selecting a newer interpreter than CI requested.
- Run the legacy client matrix in standalone environments so each job uses the Python version named by the matrix.
- Wait for intentional reconnects in the large ordered-consumer test before cleanup.

Validation:
- `uv 0.12.2` lock check and locked dev sync
- `uv 0.12.2` locked dev sync from all five workspace members
- format, lint, codespell, and ty checks
- legacy `nats` suite on Python 3.11: 290 passed, 21 skipped, 54 subtests passed
- affected callback and ordered-consumer tests: 3 passed
- ordered-consumer reconnect test stress run: 5/5 passed
- nats-server: 34 passed
- nats-core: 298 passed
- nats-jetstream: 304 passed, 4 pre-existing multi-value-header assertion failures in unchanged code
- nats-key-value: 63 passed

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,7 +82,7 @@ jobs:
               uses: astral-sh/setup-uv@v6
 
             - name: Install dependencies
-              run: uv sync --group test
+              run: uv sync --dev
               working-directory: ${{ matrix.project }}
 
             - name: Run tests for ${{ matrix.project }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,9 +82,9 @@ jobs:
               uses: astral-sh/setup-uv@v6
 
             - name: Install dependencies
-              run: uv sync --dev
+              run: uv sync --group test
               working-directory: ${{ matrix.project }}
 
             - name: Run tests for ${{ matrix.project }}
-              run: uv run pytest -v -n auto
+              run: uv run --no-sync pytest -v -n auto
               working-directory: ${{ matrix.project }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,6 +44,8 @@ jobs:
               uses: astral-sh/setup-uv@v6
 
             - name: Install dependencies and project
+              # Installed outside the workspace: uv.lock is pinned to >=3.13 for the
+              # modern packages, so it cannot resolve the 3.8-3.12 legs of this matrix.
               run: |
                   uv venv --no-project --python python
                   uv pip install --python .venv/bin/python --editable ./nats --group ./nats/pyproject.toml:dev
@@ -60,7 +62,7 @@ jobs:
                 python-version: ["3.13"]
                 os: ["ubuntu-latest", "macos-latest", "windows-latest"]
                 nats-server-version: ["latest"]
-                project: ["nats-server", "nats-core", "nats-key-value"]
+                project: ["nats-server", "nats-core", "nats-jetstream", "nats-key-value"]
         steps:
             - name: Checkout repository
               uses: actions/checkout@v5
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,7 +48,7 @@ jobs:
 
             - name: Run tests
               run: |
-                  uv run pytest -x -vv -s --continue-on-collection-errors ./nats/tests
+                  uv run --no-sync pytest -x -vv -s --continue-on-collection-errors ./nats/tests
 
     project:
         name: ${{ matrix.project }} (python-${{ matrix.python-version }}, nats-server-${{ matrix.nats-server-version }}, ${{ matrix.os }})
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,11 +44,13 @@ jobs:
               uses: astral-sh/setup-uv@v6
 
             - name: Install dependencies and project
-              run: uv sync --dev
+              run: |
+                  uv venv --no-project --python python
+                  uv pip install --python .venv/bin/python --editable ./nats --group ./nats/pyproject.toml:dev
 
             - name: Run tests
               run: |
-                  uv run --no-sync pytest -x -vv -s --continue-on-collection-errors ./nats/tests
+                  .venv/bin/python -m pytest -x -vv -s --continue-on-collection-errors ./nats/tests
 
     project:
         name: ${{ matrix.project }} (python-${{ matrix.python-version }}, nats-server-${{ matrix.nats-server-version }}, ${{ matrix.os }})
--- a/nats-core/pyproject.toml
+++ b/nats-core/pyproject.toml
@@ -39,15 +39,12 @@ Source = "https://github.com/nats-io/nats.py"
 module-name = "nats.client"
 
 [dependency-groups]
-package-dev = [
+dev = [
   "nats-server",
   "nkeys>=0.1.0",
   "pytest-benchmark>=5.2.1",
   "websockets>=16.0",
 ]
-
-[tool.uv]
-default-groups = ["package-dev"]
 
 [tool.uv.sources]
 nats-server = { workspace = true }
--- a/nats-core/pyproject.toml
+++ b/nats-core/pyproject.toml
@@ -42,7 +42,11 @@ module-name = "nats.client"
 dev = [
   "nats-server",
   "nkeys>=0.1.0",
+  "pytest>=9.0.3",
+  "pytest-asyncio>=0.21.0",
   "pytest-benchmark>=5.2.1",
+  "pytest-cov>=7.0.0",
+  "pytest-xdist>=3.0.0",
   "websockets>=16.0",
 ]
 
--- a/nats-core/pyproject.toml
+++ b/nats-core/pyproject.toml
@@ -39,16 +39,15 @@ Source = "https://github.com/nats-io/nats.py"
 module-name = "nats.client"
 
 [dependency-groups]
-dev = [
+package-dev = [
   "nats-server",
   "nkeys>=0.1.0",
-  "pytest>=7.0.0",
-  "pytest-asyncio>=0.21.0",
-  "pytest-cov>=7.0.0",
-  "pytest-xdist>=3.0.0",
   "pytest-benchmark>=5.2.1",
   "websockets>=16.0",
 ]
+
+[tool.uv]
+default-groups = ["package-dev"]
 
 [tool.uv.sources]
 nats-server = { workspace = true }
--- a/nats/tests/test_client.py
+++ b/nats/tests/test_client.py
@@ -921,8 +921,13 @@ class ClientTest(SingleServerTestCase):
         await nc.publish("tests.1", b"foo")
 
         received_message = None
+
+        async def first_message():
+            async for msg in sub.messages:
+                return msg
+
         try:
-            received_message = await asyncio.wait_for(sub.messages.__anext__(), timeout=0.5)
+            received_message = await asyncio.wait_for(first_message(), timeout=0.5)
         except asyncio.TimeoutError:
             pass
 
@@ -3114,11 +3119,6 @@ class ClientDrainTest(SingleServerTestCase):
 
     @async_test
     async def test_drain_cancelled_errors_raised(self):
-        try:
-            pass
-        except ImportError:
-            pytest.skip("skip since cannot use AsyncMock")
-
         nc = NATS()
         await nc.connect()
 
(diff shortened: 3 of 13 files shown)


renerocksai (commented):
Please bear in mind, I am fairly new to nats-py and might be wrong in my assessments. The review focused on uv dependency-group precedence and how it affects the test and check jobs; I haven't evaluated NATS-specific design choices.

I reproduced the shadowing against uv 0.12.2 to check the premise: on `main`, `uv sync --dev` inside `nats-key-value` installs neither `pytest` nor `pytest-xdist` (its own `dev` group shadows the root's rather than merging), which is what `test.yml` then needs for `uv run pytest -v -n auto`. After this PR all five members get the shared tooling.

Two things I think are worth a look:

1. **`test.yml`'s legacy Python matrix looks inoperative — I believe every entry runs on 3.13.** This is pre-existing and not something this PR introduces; I'm raising it here only because the PR fixes the identical problem in `check.yml` ("...instead of silently selecting a newer interpreter than CI requested") and the same pattern is still present in the job where it arguably matters more.

   The workspace root has no `[project]` table, so the effective `requires-python` is the intersection across members — `>=3.13`, from `nats-core` and `nats-key-value`. With 3.8 first on `PATH`, `uv sync --dev` at the root doesn't fail; it downgrades to a warning and builds a 3.13 environment anyway:

   ```
   warning: The requested interpreter resolved to Python 3.8.20, which is incompatible
            with the project's Python requirement: `>=3.13`

   $ .venv/bin/python --version
   Python 3.13.11
   ```

   So `python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]` followed by `uv run pytest ./nats/tests` would be exercising 3.13 six times rather than the declared matrix.

   I don't have a clean fix to offer: I tried `uv sync --package nats-py --python 3.8` both with and without `--dev` and both still fail on the workspace-wide intersection, and the new `test` group's `pytest>=9.0.3` carries its own floor well above 3.8. It may need the legacy job to step outside the workspace environment entirely. Very reasonable to split this into its own issue rather than grow this PR.

2. `default-groups = ["package-dev"]` makes `--dev` load-bearing for member-level syncs — plain `uv sync` in `nats-core` silently loses `pytest-xdist`. Details inline.

One smaller tightening on the `check` group markers — hedged inline, take or leave.

nats-core/pyproject.toml
```diff
@@ -39,17 +39,16 @@ Source = "https://github.com/nats-io/nats.py"
...
-  "pytest>=7.0.0",
-  "pytest-asyncio>=0.21.0",
-  "pytest-cov>=7.0.0",
-  "pytest-xdist>=3.0.0",
   "pytest-benchmark>=5.2.1",
   "websockets>=16.0",
 ]
 
+[tool.uv]
+default-groups = ["package-dev"]
```
renerocksai: `default-groups = ["package-dev"]` replaces uv's implicit `["dev"]` rather than adding to it, so a plain `uv sync` from inside a member no longer pulls the workspace `test` group. Clean syncs on uv 0.12.2:

| plain `uv sync` in… | main | this PR |
|---|---|---|
| `nats-core` | pytest ✓, xdist ✓ | pytest ✓, **xdist ✗** |
| `nats-key-value` | ✗, ✗ | ✗, ✗ (unchanged) |
| `nats-jetstream` | ✗, ✗ | ✗, ✗ (unchanged) |

The surviving `pytest` in `nats-core` looks incidental — `pytest-benchmark` depends on it — and `pytest-benchmark` doesn't pull `pytest-xdist`, so `uv run pytest -n auto` in that directory would fail with `unrecognized arguments: -n`.

CI isn't affected, since both jobs pass `--dev` explicitly; this only bites someone mirroring the CI step locally without it. I mention it mainly because it's a behaviour change from main specifically for `nats-core`.

I don't have a tidy fix either — the obvious `default-groups = ["package-dev", "dev"]` hard-errors, because the member has no `dev` group of its own:

```
error: Default group `dev` (from `tool.uv.default-groups`) is not defined in the project's `dependency-groups` table
```

so this may just be worth documenting as "`--dev` is required when syncing from a workspace member".

pyproject.toml
```diff
@@ -1,12 +1,20 @@
...
     "pytest>=9.0.3",
     "pytest-asyncio>=0.21.0",
     "pytest-cov>=7.0.0",
     "pytest-xdist>=3.0.0",
+]
+check = [
     "ty>=0.0.0a6",
     "ruff>=0.1.0",
     "codespell>=2.2.0",
+    "nkeys>=0.1.0; python_version >= '3.13'",
```
renerocksai: Small one: gating these two on `python_version >= '3.13'` means the type check degrades quietly rather than loudly. If the `type` job's interpreter ever drifts below 3.13 again, `nkeys` and `websockets` simply won't install and `ty check` goes back to passing without them — the same silent gap this PR is closing, just deferred to the next time someone edits the workflow.

Pinning the job to 3.13 here handles it for now. If you wanted it to stay handled, either asserting the interpreter in the job or putting the marker at the actual floor these two support (rather than at 3.13) would make a future drift fail instead of quietly weaken. Take or leave.

.github/workflows/test.yml
```diff
@@ -86,5 +86,5 @@ jobs:
               working-directory: ${{ matrix.project }}
 
             - name: Run tests for ${{ matrix.project }}
-              run: uv run pytest -v -n auto
+              run: uv run --no-sync pytest -v -n auto
```
renerocksai, line 93: Small consistency thing: this step and all four jobs in `check.yml` now pass `--no-sync`, but the legacy `nats` job above (line 51) still runs

```yaml
uv run pytest -x -vv -s --continue-on-collection-errors ./nats/tests
```

without it, after its own `uv sync --dev` on line 47.

Harmless as far as I can tell — the environment is already in the state `uv run` would sync it into — but it's the one run step that didn't get the treatment, and adding it there would make the "sync once, then run" pattern uniform across both workflows.

renerocksai (commented):
Please bear in mind, I am fairly new to nats-py and might be wrong in my assessments. The review focused on uv dependency-group precedence and how it affects the test and check jobs; I haven't evaluated NATS-specific design choices.

Re-checked both items from my last pass against uv 0.12.2 and they look resolved: plain `uv sync` now yields `pytest` and `pytest-xdist` from all four members (previously only `nats-core`), and `[tool.uv.dependency-groups] check = { requires-python = ">=3.13" }` makes an older interpreter fail loudly — `uv sync --only-group check --python 3.12` errors now rather than quietly dropping the deps. `uv lock --check` is clean.

One thing still open:

1. **The legacy `nats` matrix isn't testing the versions it names — and it's landing on 3.14, not 3.13.** I raised the interpreter-selection issue last time and you reasonably left it out of scope; I'm bringing it back only because this PR's own CI run gives much harder evidence than my local repro did.

   From the job logs on `d0d7e82`:

   | matrix entry | interpreter the venv actually got |
   |---|---|
   | `3.8` | **3.14.7** |
   | `3.9` | **3.14.7** |
   | `3.12` | **3.14.7** |
   | `3.13` | 3.13.14 ✓ |

   The `nats (3.8, latest)` job downloads and installs CPython 3.8.18 via `setup-python`, and then `uv sync --dev` builds the environment on 3.14.7 and the tests run there. Five of the six entries exercise an interpreter that isn't in the matrix at all, and all six report green. Unlike my local run there is no warning anywhere in the CI log — uv simply downloads a managed interpreter, so it's entirely silent.

   Still pre-existing, and still completely fine to split into its own issue rather than grow this PR. I'd just rather it not stay invisible, since as it stands the legacy client has no 3.8–3.12 coverage at all.

One small consistency nit — hedged inline, take or leave.

.github/workflows/test.yml
caspervonb: Addressed in f140985 — the legacy test run now uses `uv run --no-sync` after the explicit sync.

caspervonb:
Addressed the legacy matrix issue in fad0190. The `nats-py` member now has a Python-3.8-compatible `dev` group, and the matrix creates a standalone venv from the interpreter provided by `setup-python`, installs that member group directly, and invokes pytest through the venv Python. This avoids workspace-wide interpreter resolution and prevents uv from silently replacing the requested matrix version. Verified locally with Python 3.8.20: dependency installation succeeded and all 311 tests collected.

nats/tests/test_client.py
```diff
@@ -922,9 +922,7 @@ async def receive_message():
 
         received_message = None
         try:
-            async with asyncio.timeout(0.5):
-                async for received_message in sub.messages:
-                    break
+            received_message = await asyncio.wait_for(sub.messages.__anext__(), timeout=0.5)
```
renerocksai: Two small things in this file, now that the 3.8 lane is genuinely exercised.

1. **This line lets `StopAsyncIteration` escape.** `__anext__()` raises it when the iterator is exhausted, and it isn't caught by the `except asyncio.TimeoutError` below — so exhaustion would surface as a confusing error rather than the clean `assert received_message is not None` failure the old `async for` form produced. I don't think it's reachable as written, since the subscription is still live here, so purely a robustness nit. `anext()` isn't available on 3.8, so the dunder is somewhat forced, but a small local coroutine sidesteps both:

   ```python
   async def _first_message():
       async for msg in sub.messages:
           return msg

   received_message = await asyncio.wait_for(_first_message(), timeout=0.5)
   ```

2. **Dead guard at lines 3117-3120**, while you're in this file:

   ```python
   try:
       pass
   except ImportError:
       pytest.skip("skip since cannot use AsyncMock")
   ```

   `try: pass` can't raise, so the skip can never fire — it looks like the import it was guarding got removed at some point and the husk stayed behind. Pre-existing rather than anything you introduced, but it's the same AsyncMock-on-old-Python thread you've just been pulling on, so it may as well go.

renerocksai (commented):
The matrix fix does what it says — each entry now builds on the interpreter it names, where five of six previously landed on 3.14.7:

| entry | 3.8 | 3.9 | 3.10 | 3.11 | 3.12 | 3.13 |
|---|---|---|---|---|---|---|
| before | 3.14.7 | 3.14.7 | 3.14.7 | 3.14.7 | 3.14.7 | 3.13.14 |
| now | 3.8.18 | 3.9.25 | 3.10.20 | 3.11.15 | 3.12.13 | 3.13.14 |

I also checked the narrower `--group ./nats/pyproject.toml:dev` install didn't quietly shrink the suite: 311 tests collect on both 3.8 and 3.13, matching the 311 the old job collected. Worth confirming explicitly, since `--continue-on-collection-errors` would have swallowed an import failure.

One thing that falls out of it:

1. **`test_js.py::OrderedConsumerTest::test_ordered_consumer_larger_streams` is failing intermittently, and the real matrix makes it bite.**

   ```
   FAILED nats/tests/test_js.py::OrderedConsumerTest::test_ordered_consumer_larger_streams
       - ConnectionResetError: Connection lost
   ```

   It doesn't look version-determined — on `c0562af` it took down 3.11 in one workflow run and 3.9 in the other, with each version's counterpart passing, and it passes on re-run. That's 2 of the 12 legacy jobs for a single commit, so a six-entry matrix is going to be red more often than not.

   I can't tell from this sample whether it's a latent flake that the old all-3.14 matrix simply never exposed, or something genuinely worse on older interpreters — main's recent runs are green, but those were effectively six copies of 3.14, so they aren't much evidence either way. Either way it seems worth its own issue (quarantine, a retry, or a fix) rather than letting the matrix land intermittently red and train people to ignore it.

   Also worth noting `-x` means each job stops at the first failure, so nothing behind this one is visible yet on the older entries.

nats/tests/test_client.py
caspervonb: Addressed in 0feba1f7: the iterator now uses a small async-for helper so exhaustion preserves the previous behavior, and the dead ImportError guard has been removed. The ordered-consumer reconnect race reported in the review summary is also fixed by waiting for the reconnect callback before continuing and closing the producer connection.

caspervonb:
@renerocksai @claude PTAL — all addressed review threads are resolved, the reconnect race and callback-test cleanups are in 0feba1f7, and CI is green across Python 3.8–3.13.

caspervonb:
PTAL @claude

Jarema (approved):
LGTM!

PR #1001, opened 2026-08-14
Allow fractional connection durations

### LINKED ISSUE #1000: Annotate connect timeout and reconnect wait as float
Client.connect() annotates connect_timeout and reconnect_time_wait as int, but both are passed to asyncio APIs that accept fractional seconds (asyncio.wait_for and asyncio.sleep).

Please annotate these options as float. This remains compatible with integer callers while allowing clients to configure sub-second timeouts and reconnect delays without type-checker suppressions.

attilaolah:
Closes #1000.

Updates the public Client.connect() annotations for connect_timeout and reconnect_time_wait to float, matching their use as asyncio durations. Integer callers remain supported.

Validation:
- uv run ruff format --check nats/src/nats/aio/client.py
- uv run ruff check nats/src/nats/aio/client.py
- uv run ty check (fails with 167 existing diagnostics outside nats/)
- uv run pytest nats/tests not run because nats-server is not installed


attilaolah:
It would be even nicer to accept `timedelta | float` here, and take the `.total_seconds()` if a timedelta was passed. But that is already a runtime change.

This PR is trying to keep it minimal and only make a typing correction, no runtime change is intended.

PR #1008, opened 2026-08-20
fix: exception type

superlevure:
Tentative fix for https://github.com/nats-io/nats.py/pull/934/changes#r3823522044

### DIFF
--- a/nats/src/nats/js/client.py
+++ b/nats/src/nats/js/client.py
@@ -1231,7 +1231,7 @@ class JetStreamContext(JetStreamManager):
                 # Return any message that was already available in the internal queue.
                 if msgs:
                     return msgs
-                raise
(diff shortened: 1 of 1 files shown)


fallenmi (approved):
Reviewed with Codex at exact head `dcf8fcd`.

I independently exercised all three modified timeout exits against the exact base and head: 90 deterministic base executions raised `asyncio.TimeoutError`, while 90 exact-head executions raised `FetchTimeoutError`. The latter remains catchable through the documented `nats.errors.TimeoutError` base class, restoring the pre-2.15 public exception contract across the initial no-wait timeout and both exhausted-deadline branches.

The touched module also compiles cleanly, the patch passes `git diff --check`, and all 27 live checks are green. The head and zero-feedback state were unchanged immediately before submission.

PR #1009, opened 2026-08-21
fix: scheme when no port specified

joeriddles:
Fix the scheme `tls://` is dropped when a port is not specified in the server URL. Same for username and password.

For example: `tls://connect.ngs.global` (Synadia Cloud)


wallyqs (approved):
LGTM

PR #1011, opened 2026-08-26
Publish KV mutations through the JetStream API prefix for domain contexts

### LINKED ISSUE #1010: KV mutations ignore JetStream domain and publish to the local $KV subject
### Observed behavior

When a `JetStreamContext` targets a remote JetStream domain over a leafnode connection, the remote KV bucket can be discovered and read, but `KeyValue.put()`, `create()`, `update()`, and `delete()` publish to the unqualified local `$KV.<bucket>.<key>` subject.

For example:

```python
nc = await nats.connect("tls://hub.example.net:4222", ...)
js = nc.jetstream(domain="leaf")
kv = await js.key_value("example")
```

`key_value()` succeeds because the bucket lookup uses the domain-prefixed JetStream API:

```text
$JS.leaf.API.STREAM.INFO.KV_example
```

However:

```python
await kv.create("test", b"value")
```

fails with:

```text
NoStreamResponseError: nats: no response from stream
```

Tracing the client shows that the mutation is sent to:

```text
$KV.example.test
```

instead of:

```text
$JS.leaf.API.$KV.example.test
```

The equivalent NATS CLI operation succeeds over the same connection and permissions:

```bash
nats \
  --server tls://hub.example.net:4222 \
  --js-domain leaf \
  kv create example test value
```

Publishing the domain-qualified subject directly with `nats-py` also succeeds:

```python
await js.publish(
    "$JS.leaf.API.$KV.example.test",
    b"value",
    headers={"Nats-Expected-Last-Subject-Sequence": "0"},
)
```

CAS behavior also works through this explicit subject. A valid expected revision succeeds, while a stale revision returns JetStream error `10071` (`wrong last sequence`).

The relevant v2.15.0 implementation appears to explain the behavior:

- [`JetStreamContext.__init__`](https://github.com/nats-io/nats.py/blob/24626a43631a21af2e35261cf0a8ec615fd35c78/nats/src/nats/js/client.py#L105-L134) changes `_prefix` to `$JS.<domain>.API`.
- [`JetStreamContext.key_value`](https://github.com/nats-io/nats.py/blob/24626a43631a21af2e35261cf0a8ec615fd35c78/nats/src/nats/js/client.py#L1392-L1410) still initializes `KeyValue` with `KV_PRE_TEMPLATE`, which is defined as `$KV.{bucket}.` [at line 64](https://github.com/nats-io/nats.py/blob/24626a43631a21af2e35261cf0a8ec615fd35c78/nats/src/nats/js/client.py#L64).
- [`KeyValue._update`](https://github.com/nats-io/nats.py/blob/24626a43631a21af2e35261cf0a8ec615fd35c78/nats/src/nats/js/kv.py#L277-L304) publishes `f"{self._pre}{key}"` without applying the domain.
- [`JetStreamContext.publish`](https://github.com/nats-io/nats.py/blob/24626a43631a21af2e35261cf0a8ec615fd35c78/nats/src/nats/js/client.py#L177-L220) publishes the supplied subject verbatim.

This affects at least `put`, `create`, `update`, `delete`, and `purge`, since they all publish using `self._pre`.

### Expected behavior

A `KeyValue` obtained from `JetStreamContext(domain="leaf")` should perform all KV operations against that domain.

KV mutations should use the remote-domain data subject:

```text
$JS.leaf.API.$KV.<bucket>.<key>
```

This would make high-level `nats-py` KV operations consistent with:

- Other operations performed through the domain-scoped `JetStreamContext`.
- The behavior of `nats --js-domain leaf kv ...`.
- The ability to bind and read the same remote bucket through `nats-py`.
- The Go client behavior used by the NATS CLI.

Alternatively, if remote-domain KV mutations are intentionally unsupported, `key_value()` should fail early or the limitation should be documented rather than returning a partially functional `KeyValue`.

### Server and client version

```text
nats-server: 2.14.5
nats-py: 2.15.0
nats CLI: 0.2.3
Python: 3.14.3
```

The two NATS clusters use different JetStream domains and are connected through a leafnode. The same account is connected on both sides, and permissions allow the domain-prefixed KV API and data subjects.

### Host environment

```text
Client:
- macOS
- Apple Silicon / arm64
- CPython 3.14.3

Servers:
- nats-server 2.14.5
- Linux containers
- Kubernetes
- Two independent JetStream domains connected through a leafnode
```

The same behavior was reproduced against two separate remote JS domains.

### Steps to reproduce

1. Configure two NATS servers with distinct JetStream domains connected through a leafnode:

```text
hub:  domain = hub
leaf: domain = leaf
```

2. Create KV bucket `example` in the `leaf` domain.

3. Connect a Python client to `hub`.

4. Run:

```python
from __future__ import annotations

import asyncio
import ssl

import nats


async def main() -> None:
    tls = ssl.create_default_context()

    nc = await nats.connect(
        "tls://hub.example.net:4222",
        tls=tls,
        tls_handshake_first=True,
    )

    try:
        js = nc.jetstream(domain="leaf")

        # Succeeds: STREAM.INFO uses $JS.leaf.API.
        kv = await js.key_value("example")

        # Fails with NoStreamResponseError because this publishes
        # $KV.example.test instead of $JS.leaf.API.$KV.example.test.
        await kv.create("test", b"value")
    finally:
        await nc.close()


asyncio.run(main())
```

5. Confirm that the equivalent CLI command succeeds:

```bash
nats \
  --server tls://hub.example.net:4222 \
  --js-domain leaf \
  kv create example test value
```

6. Confirm that explicitly publishing the remote-domain subject from Python also succeeds:

```python
await js.publish(
    "$JS.leaf.API.$KV.example.test",
    b"value",
    headers={"Nats-Expected-Last-Subject-Sequence": "0"},
)
```

thejoeejoee:
Fixes #1010.

## Problem

When a `JetStreamContext` targets a JetStream domain (`nc.jetstream(domain="leaf")`), the API prefix becomes `$JS.<domain>.API`, but `KeyValue` mutations (`put`, `create`, `update`, `delete`, `purge`) still published to the unqualified local `$KV.<bucket>.<key>` subject. Over a leafnode this fails with `NoStreamResponseError`, while reads (bucket lookup, `get`, `watch`) already worked through the domain-scoped API.

## Fix

Match the Go client behavior (`nats.go` `useJSPfx`): when the context's API prefix is not the default `$JS.API`, prepend it to the KV data subject for mutations, yielding `$JS.<domain>.API.$KV.<bucket>.<key>`.

Reads (`get`, `history`, `keys`, `watch`, `purge_deletes`) intentionally keep the local `$KV.` prefix: stream subjects are not prefixed, and watchers filter on the stream's subjects. This mirrors the Go split between `pre` (read) and the JS-prefixed mutation subject.

## Tests

- New `SingleJetStreamServerDomainTestCase` (`nats/tests/utils.py`) starting a server with `jetstream { domain: "test-domain" }` (`nats/tests/conf/js-domain.conf`).
- New `KVDomainTest` (`nats/tests/test_js.py`):
  - `test_kv_mutations_use_domain_qualified_subject` — spies on the connection and asserts every mutation request goes to `$JS.test-domain.API.$KV.TEST_DOMAIN.<key>` (regression test for the reported bug).
  - `test_kv_mutations_work_over_domain` — full put/get/create/update/CAS/delete/purge round-trip through a domain-scoped context.
  - `test_kv_binding_via_key_value_uses_domain` — covers the `key_value()` binding path from the issue.

All `nats/tests` pass locally (237 passed, 21 skipped).

## Out of scope

`ObjectStore` mutations have the same unqualified-subject issue (`$O.<bucket>.C/M.`). Left for a follow-up to keep this change minimal.

### DIFF
--- a/nats/src/nats/js/kv.py
+++ b/nats/src/nats/js/kv.py
@@ -149,6 +149,14 @@ class KeyValue:
         self._js = js
         self._direct = direct
 
+        # KV mutations publish through the JetStream API prefix when the
+        # context targets a non-default domain (nats.go useJSPfx behavior):
+        # "$JS.<domain>.API.$KV.<bucket>.<key>". Reads and watchers keep the
+        # local data subject since stream subjects are not prefixed.
+        self._mutation_pre = pre
+        if js._prefix != api.DEFAULT_PREFIX:
+            self._mutation_pre = f"{js._prefix}.{pre}"
+
     async def get(self, key: str, revision: Optional[int] = None, validate_keys: bool = True) -> Entry:
         """
         get returns the latest value for the key.
@@ -219,7 +227,7 @@ class KeyValue:
         if validate_keys and not _is_key_valid(key):
             raise nats.js.errors.InvalidKeyError(key)
 
-        pa = await self._js.publish(f"{self._pre}{key}", value)
+        pa = await self._js.publish(f"{self._mutation_pre}{key}", value)
         return pa.seq
 
     async def create(self, key: str, value: bytes, validate_keys: bool = True, msg_ttl: Optional[float] = None) -> int:
@@ -292,7 +300,7 @@ class KeyValue:
 
         pa = None
         try:
-            pa = await self._js.publish(f"{self._pre}{key}", value, headers=hdrs, msg_ttl=msg_ttl)
+            pa = await self._js.publish(f"{self._mutation_pre}{key}", value, headers=hdrs, msg_ttl=msg_ttl)
         except nats.js.errors.APIError as err:
             # Check for a BadRequest::KeyWrongLastSequenceError error code.
             # 10071: JSStreamWrongLastSequenceErrF
@@ -336,7 +344,7 @@ class KeyValue:
         if last and last > 0:
             hdrs[api.Header.EXPECTED_LAST_SUBJECT_SEQUENCE] = str(last)
 
-        await self._js.publish(f"{self._pre}{key}", headers=hdrs)
+        await self._js.publish(f"{self._mutation_pre}{key}", headers=hdrs)
         return True
 
     async def purge(self, key: str, msg_ttl: Optional[float] = None) -> bool:
@@ -349,7 +357,7 @@ class KeyValue:
         hdrs = {}
         hdrs[KV_OP] = KV_PURGE
         hdrs[api.Header.ROLLUP] = MSG_ROLLUP_SUBJECT
-        await self._js.publish(f"{self._pre}{key}", headers=hdrs, msg_ttl=msg_ttl)
+        await self._js.publish(f"{self._mutation_pre}{key}", headers=hdrs, msg_ttl=msg_ttl)
         return True
 
     async def purge_deletes(self, olderthan: int = 30 * 60) -> bool:
--- a/nats/tests/conf/js-domain.conf
+++ b/nats/tests/conf/js-domain.conf
@@ -1,0 +1,6 @@
+port: 4222
+http: 8222
+
+jetstream: {
+  domain: "test-domain"
+}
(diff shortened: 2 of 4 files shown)


caspervonb (approved):
lgtm

PR #1015, opened 2026-08-29
Add jetstream consumer priority groups (ADR-42)

### LINKED ISSUE #784: Missing support for jetstream consumer groups
### Proposed change

Server 2.11 introduced jetstream consumer priority groups from [ADR-42](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-42.md) ([2.11 release notes](https://docs.nats.io/release-notes/whats_new/whats_new_212#consumers) for initial addition, [2.12 release notes](https://docs.nats.io/release-notes/whats_new/whats_new_212#consumers) for `prioritized` policy addition).

From looking at the original nats.go implementation in https://github.com/nats-io/nats.go/commit/93f7ce4c, I believe the translation of this would roughly be:
- 3 new js consumer config fields: `nats.js.api.ConsumerConfig.{priority_policy,priority_timeout,priority_groups}`
- 3 new options on `nats.js.client.JetStreamContext.{pull_subscribe(),pull_subscribe_bind()}` (or could just rely on sdk user manually creating multiple consumer configs for the same consumer group to configure these themselves): `pull_min_pending`, `pull_min_ack_pending`, `pull_priority_group`
- modification to the `nats.js.client.JetstreamContext.PullSubscription.{_fetch_one(),_fetch_n()}` methods to handle priority groups in the pull request: need to add `id`, `group`, `min_pending`, `min_ack_pending`, `priority`
- 1 new js api method: `nats.js.manager.JetstreamManager.unpin_consumer()`
- 1 new js error: `nats.js.errors.PinIdMismatchError` (error code 423)
- 1 new js consumer info field: `nats.js.api.ConsumerInfo.priority_groups`

### Use case

Would then allow the implementation of, among other things, partitioned consumer groups – which are super useful for doing things like event sourcing without having to set max ack pending to 1 and slowing a consumer to a crawl

### Contribution

~I'm happy to make a PR~ I have made a PR: #786

caspervonb:
Supersedes #786, which had drifted behind main. Carries @oliverlambson's original commit as-is, plus a follow-up commit addressing the review findings on that PR and a few more found while rebasing:

- `PriorityPolicy.NONE` sent `""`; the server only accepts `"none"`.
- `priority=0` (the highest priority) was dropped from pull requests by a truthiness check.
- `min_pending`/`min_ack_pending`/`priority` were only validated on the single-message fetch path.
- `ConsumerInfo` raised on any consumer with an unpinned priority group, since `pinned_client_id` was required but the server omits it until a client is pinned. `pinned_ts` is now parsed as well.
- Overflow/pinned/unpin tests were gated on 2.12 instead of 2.11.

Closes #784

### DIFF
--- a/nats/src/nats/js/api.py
+++ b/nats/src/nats/js/api.py
@@ -622,26 +622,27 @@ class ReplayPolicy(str, Enum):
 
 
 class PriorityPolicy(str, Enum):
-    """Priority policy for priority groups.
+    """Priority policy for pull consumer priority groups.
 
-    Enables flexible failover and priority management when multiple clients are
-    pulling from the same consumer
+    Enables flexible failover and priority management when multiple clients
+    are pulling from the same consumer.
 
-    Introduced in nats-server 2.12.0.
+    Introduced in nats-server 2.11.0 (``PRIORITIZED`` in 2.12.0).
 
     References:
         * `Consumers, Pull consumer priority groups <https://docs.nats.io/release-notes/whats_new/whats_new_211#consumers>`
         * `Consumers, Prioritized pull consumer policy <https://docs.nats.io/release-notes/whats_new/whats_new_212#consumers>`
     """  # noqa: E501
 
-    NONE = ""
-    "default"
+    NONE = "none"
+    """Default, no priority handling."""
     PINNED = "pinned_client"
-    "pins a consumer to a specific client"
+    """Pins the consumer to a single client per group; others take over when it goes away."""
     OVERFLOW = "overflow"
-    "allows for restricting when a consumer will receive messages based on the number of pending messages or acks"
+    """Only delivers to a client once ``min_pending`` or ``min_ack_pending`` thresholds are reached."""
     PRIORITIZED = "prioritized"
-    """allows for restricting when a consumer will receive messages based on a priority from 0-9 (0 is highest priority & default)
+    """Delivers to the client with the highest priority (0-9, 0 is highest and default).
+
     Introduced in nats-server 2.12.0.
     """
 
@@ -703,7 +704,7 @@ class ConsumerConfig(Base):
     priority_policy: Optional[PriorityPolicy] = None
 
     # The duration (seconds) after which the client will be unpinned if no new
-    # pull requests are sent.Used with PriorityPolicy.PINNED.
+    # pull requests are sent. Used with PriorityPolicy.PINNED.
     # Introduced in nats-server 2.11.0.
     priority_timeout: Optional[float] = None
 
@@ -729,7 +730,8 @@ class ConsumerConfig(Base):
         result["ack_wait"] = self._to_nanoseconds(self.ack_wait)
         result["idle_heartbeat"] = self._to_nanoseconds(self.idle_heartbeat)
         result["inactive_threshold"] = self._to_nanoseconds(self.inactive_threshold)
-        result["priority_timeout"] = self._to_nanoseconds(self.priority_timeout)
+        if self.priority_timeout is not None:
+            result["priority_timeout"] = self._to_nanoseconds(self.priority_timeout)
         if self.backoff:
             result["backoff"] = [self._to_nanoseconds(i) for i in self.backoff]
         return result
@@ -755,10 +757,28 @@ class SequenceInfo(Base):
 
 @dataclass
 class PriorityGroupState(Base):
+    """
+    State of a consumer priority group.
+
+    Introduced in nats-server 2.11.0.
+    """
+
     group: str
-    pinned_client_id: str
-    # FIXME: Do not handle dates for now.
-    # pinned_ts: datetime
+    # Generated ID of the pinned client. Only set when a client is pinned.
+    pinned_client_id: Optional[str] = None
+    # When the client was pinned. Only set when a client is pinned.
+    pinned_ts: Optional[datetime.datetime] = None
+
+    @classmethod
+    def from_response(cls, resp: Dict[str, Any]):
+        cls._convert_utc_iso(resp, "pinned_ts")
+        return super().from_response(resp)
+
+    def as_dict(self) -> Dict[str, object]:
+        result = super().as_dict()
+        if self.pinned_ts is not None:
+            result["pinned_ts"] = self._to_utc_iso(self.pinned_ts)
+        return result
 
 
 @dataclass
@@ -794,6 +814,7 @@ class ConsumerInfo(Base):
         cls._convert(resp, "ack_floor", SequenceInfo)
         cls._convert(resp, "config", ConsumerConfig)
         cls._convert(resp, "cluster", ClusterInfo)
+        cls._convert(resp, "priority_groups", PriorityGroupState)
         cls._convert_utc_iso(resp, "created")
         return super().from_response(resp)
 
--- a/nats/src/nats/js/api.py
+++ b/nats/src/nats/js/api.py
@@ -67,6 +67,7 @@ class StatusCode(str, Enum):
     NO_MESSAGES = "404"
     REQUEST_TIMEOUT = "408"
     CONFLICT = "409"
+    PIN_ID_MISMATCH = "423"
     CONTROL_MESSAGE = "100"
 
 
@@ -620,6 +621,31 @@ class ReplayPolicy(str, Enum):
     ORIGINAL = "original"
 
 
+class PriorityPolicy(str, Enum):
+    """Priority policy for priority groups.
+
+    Enables flexible failover and priority management when multiple clients are
+    pulling from the same consumer
+
+    Introduced in nats-server 2.12.0.
+
+    References:
+        * `Consumers, Pull consumer priority groups <https://docs.nats.io/release-notes/whats_new/whats_new_211#consumers>`
+        * `Consumers, Prioritized pull consumer policy <https://docs.nats.io/release-notes/whats_new/whats_new_212#consumers>`
+    """  # noqa: E501
+
+    NONE = ""
+    "default"
+    PINNED = "pinned_client"
+    "pins a consumer to a specific client"
+    OVERFLOW = "overflow"
+    "allows for restricting when a consumer will receive messages based on the number of pending messages or acks"
+    PRIORITIZED = "prioritized"
+    """allows for restricting when a consumer will receive messages based on a priority from 0-9 (0 is highest priority & default)
+    Introduced in nats-server 2.12.0.
+    """
+
+
 @dataclass
 class ConsumerConfig(Base):
     """Consumer configuration.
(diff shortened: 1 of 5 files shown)


nats/src/nats/js/client.py
```diff
@@ -1130,7 +1167,17 @@ async def _fetch_one(
                 next_req["expires"] = int(expires)
             if heartbeat:
                 next_req["idle_heartbeat"] = int(heartbeat * 1_000_000_000)  # to nanoseconds
-
+            if self._group:
+                next_req["group"] = self._group
+            pin_id = self.pin_id
+            if pin_id:
+                next_req["id"] = pin_id
+            if min_pending:
```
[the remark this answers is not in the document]
caspervonb: Fixed in 475e7655 — `min_pending` and `min_ack_pending` now use `is not None` in all three request blocks, matching `priority`.

nats/src/nats/js/client.py
```diff
@@ -1152,13 +1199,19 @@ async def _fetch_one(
...
+                            self.pin_id = None
+
                         # In case of a temporary error, treat it as a timeout to retry.
                         if JetStreamContext._is_temporary_error(status):
                             raise nats.errors.TimeoutError
                         else:
                             # Any other type of status message is an error.
                             raise nats.js.errors.APIError.from_msg(msg)
                     else:
+                        pin_id = msg.headers.get("Nats-Pin-Id") if msg.headers else None
```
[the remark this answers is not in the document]
caspervonb: Fixed in 475e7655 — added `PIN_ID = "Nats-Pin-Id"` to the `Header` enum and replaced all five literals with `api.Header.PIN_ID`.

nats/src/nats/js/client.py
```diff
@@ -1241,8 +1308,13 @@ async def _fetch_n(
                 # a possible i/o timeout error or due to a disconnection.
                 got_any_response = True
                 pass
+            elif JetStreamContext._is_pin_id_mismatch_error(status):
```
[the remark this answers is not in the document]
caspervonb: Good catch, and it turned out to be worse than a style difference — fixed in 475e7655.

Verified against nats-server 2.14: a 423 makes the server discard the pull request, so `_fetch_one` cleared the pin and then waited on a request that would never be served, burning the entire timeout (`fetch(1)` raised `TimeoutError` after doing nothing). `_fetch_n` only looked correct by accident — it falls through to its lingering request, which by then carries no stale id, so it recovered and returned messages.

`_fetch_one` now re-issues its request once without the stale id, so both paths recover in-call. That also matches ADR-42's "clear the id and continue pulling without it". Added `test_consumer_pin_id_mismatch_recovers` covering both `fetch(1)` and `fetch(2)`.

caspervonb:
PTAL @claude

renerocksai (commented):

Two things I think are worth a look:

1. `unpin_consumer` skips the name validation (`_validate_stream_name` / `_validate_consumer_name`) that its sibling manager methods perform. Details inline.
2. The 423 branches in `_fetch_n` don't set `got_any_response`, and a lingering-phase 423 can end as a silent empty return after burning the deadline. Details inline.

nats/src/nats/js/manager.py
```diff
@@ -523,6 +523,21 @@ async def get_last_msg(
...
+        """
+        unpin_consumer releases the client currently pinned to a priority group
+        of a consumer using ``PriorityPolicy.PINNED``, so that the next pull
+        request from any client in that group is pinned instead.
+
+        :param stream_name: Name of the stream the consumer belongs to.
+        :param consumer_name: Name of the consumer.
+        :param group: Priority group to unpin.
+        """
+        req_subject = f"{self._prefix}.CONSUMER.UNPIN.{stream_name}.{consumer_name}"
```
renerocksai, line 536: Every other manager method that interpolates names into an API subject validates them first (`consumer_info`, `delete_consumer`, `pause_consumer`, …). I think `unpin_consumer` should do the same — `_validate_stream_name(stream_name)` and `_validate_consumer_name(consumer_name)` — so a name containing `.`/`*`/`>`/space raises the crisp `ValueError` the siblings give instead of producing a malformed subject.

Tiny nit while here: `_ = await self._api_request(...)` on line 539 — the `_ =` can just be dropped; a bare `await` reads cleaner and matches how discarded results are handled elsewhere in the file.

nats/src/nats/js/client.py
```diff
@@ -1338,8 +1439,13 @@ async def _fetch_n(
                     if JetStreamContext._is_heartbeat(status):
                         got_any_response = True
                         continue
+                    if JetStreamContext._is_pin_id_mismatch_error(status):
```
renerocksai, line 1442: `_fetch_one` sets `got_any_response = True` when it sees a 423 (a 423 is a server response), but none of the four 423 branches in `_fetch_n` do (lines 1325, 1348, here, and 1472), even though the neighbouring heartbeat branches all do.

I think it matters most for this lingering-phase branch: per ADR-42 ("respond with a 423 … to any pulls, including waiting ones") the pin can be invalidated while the lingering request is waiting, and — per your own finding on the earlier review — the server then discards the request. As written the loop waits out the remaining deadline on a request that will never be served, and with zero collected messages it falls past the `len(msgs) == 0 and got_any_response` guard at line 1492 and returns `[]` without raising. Setting `got_any_response = True` in these branches would at least surface that as `FetchTimeoutError`; whether the lingering phase should also re-issue once without the stale id, like `_fetch_one` now does, I'll leave to you — I couldn't verify the mid-flight case statically (the new test covers the stale-id-at-request-time case, which recovers via the lingering request).

nats/src/nats/js/client.py
```diff
@@ -1383,6 +1494,14 @@ async def _fetch_n(
 
             return msgs
 
+        @property
```
renerocksai, line 1497: A getter/setter pair that just mirrors `_pin_id` is a plain attribute with extra steps — and the public setter invites callers to mutate pin state the fetch machinery owns. Consider making this a read-only property and pointing the internal writes at `self._pin_id` directly. Placement nit: the other properties (`pending_msgs`, `pending_bytes`, `delivered`) live at the top of the class; this one ended up below the private `_fetch_*` methods.

nats/src/nats/js/client.py
```diff
@@ -1218,6 +1288,17 @@ async def _fetch_n(
...
+                next_req["group"] = self._group
+            pin_id = self.pin_id
+            if pin_id:
+                next_req["id"] = pin_id
+            if min_pending is not None:
+                next_req["min_pending"] = min_pending
+            if min_ack_pending is not None:
+                next_req["min_ack_pending"] = min_ack_pending
+            if priority is not None:
+                next_req["priority"] = priority
```
renerocksai, line 1301: The `group`/`id`/`min_pending`/`min_ack_pending`/`priority` block now appears verbatim three times (`send_next_request` at 1171–1181, here, and the lingering request at 1393–1403). The pre-existing code already duplicated batch/expires/heartbeat, so this follows the local pattern, but the PR triples the duplicated surface — consider a small `_build_next_req(...)` helper so the three send sites can't drift.

nats/tests/test_js.py
```diff
@@ -6241,3 +6241,465 @@ async def error_handler(e):
...
+        # publish 100 messages
+        for i in range(100):
+            await js.publish("foo", f"{i}".encode())
+
+        # Start concurrent fetches:
+        # psub1 with priority=1 (lower priority) requesting 100 messages
+        # psub2 with priority=0 (higher priority) requesting 75 messages
+        # Expected: psub2 gets 75 first, psub1 gets remaining 25
+
+        fetch1_task = asyncio.create_task(psub1.fetch(100, timeout=2.0, priority=1))
```
renerocksai, line 6688: This may well be fine in practice — you've run it against a real server — but the exact 75/25 split looks timing-sensitive to me: the 100 messages are published before either pull request exists, so whichever no-wait probe the server processes first could plausibly be served before the other request registers. Publishing after both fetches are in flight (start the tasks, brief sleep, then publish) would make priority ordering the only thing under test. Take or leave.

nats/tests/test_js.py
```diff
@@ -6241,3 +6241,465 @@ async def error_handler(e):
...
+        # 1. Below threshold - no messages delivered
+        #   - publish 100 msgs
+        #   - fetch 10 msgs with min_pending 110
+        #   - should not get any msgs since 100<110
+        psub = await js.pull_subscribe_bind(
+            cinfo.name,
+            cinfo.stream_name,
+            priority_group="A",
+        )
+        for i in range(0, 100):
```
renerocksai, line 6349: Two tiny consistencies in the new tests, entirely cosmetic:

1. `range(0, 100)` here and at line 6364, vs `range(100)` in the sibling tests added by this PR.
2. The step-2 comment "should get 10 msgs since (200-10)>110" — the operative condition is just 200 pending ≥ 110; the `-10` had me re-reading.

nats/src/nats/js/api.py
```diff
@@ -620,6 +622,32 @@ class ReplayPolicy(str, Enum):
...
+class PriorityPolicy(str, Enum):
+    """Priority policy for pull consumer priority groups.
+
+    Enables flexible failover and priority management when multiple clients
+    are pulling from the same consumer.
+
+    Introduced in nats-server 2.11.0 (``PRIORITIZED`` in 2.12.0).
+
+    References:
+        * `Consumers, Pull consumer priority groups <https://docs.nats.io/release-notes/whats_new/whats_new_211#consumers>`
```
renerocksai, line 634: These two links (here and line 635) are the only `References:` entries in the file without the trailing underscore (`` `...`_ ``), so Sphinx will render them as interpreted text rather than hyperlinks like the others.

nats/src/nats/js/manager.py
caspervonb: Done in 13fe7f1c: unpin_consumer now validates both names first, and the `_ =` is gone.

nats/src/nats/js/client.py
caspervonb: Done in 13fe7f1c. All four 423 branches in _fetch_n now set got_any_response, and the lingering phase re-issues the request once without the stale id (same as _fetch_one) instead of waiting out the deadline on a discarded request.

nats/src/nats/js/client.py
caspervonb: Done in 13fe7f1c: pin_id is a read-only property next to the other properties, and internal writes go to _pin_id.

nats/src/nats/js/client.py
caspervonb: Done in 13fe7f1c: added _build_next_req and all three send sites use it.

nats/tests/test_js.py
caspervonb: Tried it, but publishing after the fetches are in flight breaks the test: the no-wait probe delivers whatever is available the instant the first message lands and then 404s the remainder, so the priority-0 fetch returns a single message. Keeping the original ordering, where both probes find all 100 already present.

nats/tests/test_js.py
caspervonb: Done in 13fe7f1c: range(100), and the comment now reads "200 pending >= 110".

nats/src/nats/js/api.py
caspervonb: Done in 13fe7f1c.

caspervonb:
PTAL @claude

caspervonb:
PTAL @renerocksai

wallyqs (approved):

PR #1020, opened 2026-09-12
Reject credentials without a JWT marker instead of looping at EOF

AayushDani:
The legacy client's JWT credential reader loops forever at EOF when a credential file or `RawCredentials` value has no `BEGIN NATS USER JWT` marker. Because this reader runs synchronously, the loop can block the asyncio thread; an asyncio timeout cannot interrupt it.

Raise the existing `InvalidUserCredentialsError` on EOF. Regression tests cover empty and seed-only synthetic input through string paths, `Path`, and `RawCredentials`, using a bounded subprocess so the old behavior cannot hang the test runner. Valid JWT blocks after a preamble remain covered. This change leaves reconnect/retry policy unchanged.

Validation on Python 3.12.14/macOS arm64 with NATS Server 2.14.6:
- Full legacy suite: 339 passed, 9 skipped, 54 subtests passed.
- Authentication suite: 16 passed.
- Repository Ruff, formatting, and spelling checks passed.
- A synthetic TCP handshake reproduction blocks on the base revision despite an asyncio timeout and raises the expected error with the patch.

The other Python/OS combinations and modern-client runtime tests remain for upstream CI.

AI assistance: OpenAI Codex implemented and tested this change.

### DIFF
--- a/nats/src/nats/aio/client.py
+++ b/nats/src/nats/aio/client.py
@@ -685,6 +685,8 @@ class Client:
             user_jwt = None
             while True:
                 line = bytearray(f.readline())
+                if not line:
+                    raise errors.InvalidUserCredentialsError
                 if b"BEGIN NATS USER JWT" in line:
                     user_jwt = bytearray(f.readline())
                     break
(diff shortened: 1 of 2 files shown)


caspervonb (approved):
lgtm, draft tho?

AayushDani:
> lgtm, draft tho?

Marked ready for review. @caspervonb can you please merge the PR?

PR #1021, opened 2026-09-14
Release nats v2.16.0

caspervonb:
Bump nats-py to 2.16.0 and add the changelog entry for everything since v2.15.0.


wallyqs (approved):
LGTM

